PR #1, opened 2018-06-21
Default value

zdenek-c:
Extracted parameter default value (initialization) from LLVM IR using regular expression.
Added new command-line option to print default values in the following format:
<<module>>
<<old_version>>: <<default_value>>
<<new_version>>: <<default_value>>

### DIFF
--- a/diffkemp/diffkemp.py
+++ b/diffkemp/diffkemp.py
@@ -24,6 +24,9 @@ def __make_argument_parser():
                     action="store_true")
     ap.add_argument("--report-stat", help="report statistics of the analysis",
                     action="store_true")
+    ap.add_argument("--default-value",
+            help="print parameter default value (in LLVM IR form)",
+                    action="store_true")
     return ap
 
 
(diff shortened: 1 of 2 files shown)


viktormalik:
I assume your solution will just return the line from the LLVM file where the parameter value is assigned. However, I would prefer to extract the actual value.
Example: in module sound/core/snd.bc, parameter cards_limit
What your function will return: 
"@cards_limit = internal global i32 1, align 4"
What I would like to get:
"1" (does not have to be an integer, just a string)

Implementing this will require parsing the LLVM file and extracting the information from CFG. To do this, I use the 'llvmcpy' python module, which creates bindings from python to C LLVM API.
LLVM C API: http://llvm.org/doxygen/group__LLVMC.html
llvmcpy: https://github.com/revng/llvmcpy

Check e.g. method LlvmKernelModule.find_param_var, there I use some of functions from C API to achieve a similar thing (I'm extracting value of another variable __param_#name).

viktormalik:
Travis CI checks have failed, please make sure that each PR you send passes the checks. In your case, the pyflake8 check fails. It is a tool that checks if you follow correct coding style. You can run the check locally by:
`flake8 diffkemp --exclude=llreve --ignore=F403,F405`
To see what fails, click on 'Details' of Travis build.

diffkemp/diffkemp.py
```diff
@@ -54,6 +57,33 @@ def run_from_cli():
...
+                                param_src.default_value)
+                        print "{}: {}".format(args.dest_version,
+                                param_dst.default_value)
+                    return 0
+
+                for param_name, param in module_src.params.iteritems():
+                    if module_dst.params.get(param_name):
+                        print "{}: {}".format(args.src_version, param.default_value)
+                        print "{}: {}".format(args.dest_version,
+                                module_dst.params[param_name].default_value)
```
viktormalik: This looks very similar to the code 7 lines above. Wouldn't it be better to merge it somehow?
Moreover, it is not necessary to create a whole new if branch for the new command line argument. I would prefer to put this functionality to the existing code (we can get the default values every time when we do the analysis).

PR #3, opened 2018-08-21
SimpLL: Assign global numbers to strings and APInts based on value instead of name.

lenticularis39:


diffkemp/simpll/DifferentialGlobalNumberState.cpp
```diff
@@ -20,24 +20,74 @@
...
-        if (otherValue)
-            GlobalNumbers.insert({otherValue, nextNumber});
+        auto GA = dyn_cast<GlobalVariable>(value);
+
+        // If the value is a GlobalVariable, has the attribute unnamed_addr and
+        // the initializer is either a ConstantDataSequential or a ConstantInt,
+        // use a map for comparison by value instead of name
+        if (GA && GA->hasGlobalUnnamedAddr() && GA->hasInitializer() &&
+            (dyn_cast<ConstantDataSequential>(GA->getInitializer()) ||
+             dyn_cast<ConstantInt>(GA->getInitializer()))) {
```
viktormalik: Use isa<>() instead of dyn_cast<>() if just checking for type.

diffkemp/simpll/DifferentialGlobalNumberState.cpp
```diff
@@ -20,24 +20,74 @@
...
+
+            if (ConstantDataSequential *CDS =
+                dyn_cast<ConstantDataSequential>(I)) {
+                // Value is a string - look it up in the string map
+                auto possibleResult = Strings.find(
+                    CDS->getAsString());
+
+                if(possibleResult != Strings.end()) {
+                    // If the string is in the map, set the result to it
+                    result = (*possibleResult).getValue();
```
viktormalik: Use -> instead of (*).
Also, shouldn't we insert {value, result} into the GlobalNumbers map in this case?

diffkemp/simpll/DifferentialGlobalNumberState.cpp
```diff
@@ -20,24 +20,74 @@
...
+                    result = nextNumber;
+                    nextNumber++;
+                }
+            } else if (ConstantInt *CI =
+                    dyn_cast<ConstantInt>(I)) {
+                auto possibleResult = Constants.find(CI->getValue());
+
+                if (possibleResult != Constants.end()) {
+                    // If the APInt is in the map, set the result to it
+                    result = (*possibleResult).second;
```
viktormalik: Same as above

diffkemp/simpll/DifferentialGlobalNumberState.cpp
```diff
@@ -20,24 +20,74 @@
...
+            // Value is a constant that can be compared by value
+            Constant *I = GA->getInitializer();
+
+            if (ConstantDataSequential *CDS =
+                dyn_cast<ConstantDataSequential>(I)) {
+                // Value is a string - look it up in the string map
+                auto possibleResult = Strings.find(
+                    CDS->getAsString());
+
+                if(possibleResult != Strings.end()) {
```
viktormalik: Space after "if".

diffkemp/simpll/DifferentialGlobalNumberState.h
```diff
@@ -28,6 +28,18 @@ class DifferentialGlobalNumberState : public GlobalNumberState {
     using ValueNumberMap = ValueMap<GlobalValue *, uint64_t>;
     ValueNumberMap GlobalNumbers;
 
+    // Mapping strings to numbers
+    StringMap<uint64_t> Strings;
+
+    // Mapping APInts to numbers
+    struct cmpConstants {
+        bool operator()(const APInt a, const APInt b) const {
```
viktormalik: Use references (APInt &) which is more effective (the object does not have to be copied).

PR #5, opened 2018-08-29
SimpLL: RemoveUnusedReturnValuesPass: Skip all functions without a body.

lenticularis39:
Checking whether the function has a body using `GlobalValue::LinkageTypes::ExternalLinkage` is unreliable; for example when run on simple programs it replaces the return type of the main function, making it incomparable by `DifferentialFunctionComparator`, which compares the deleted old function instead.

This bug didn't appear until the `callsTransitively` check was replaced by a function list, because the old method excluded the main function.

The proposed fix is to actually check the number of basic blocks in the function, and if it's not zero, then skip the function.


viktormalik:
But the changed condition (now using the list of called functions) is tested *after* the test for external linkage. Therefore I would suggest to change the condition to: if (&Fun == Main || CalledFuns.find(&Fun) == CalledFuns.end()) {...} to get the original behavior.
Another interesting thing is why the check for external linkage passes for the main function having a body. Anyway, this check should be done using Fun->isDeclaration() (which also does check for emptiness of the function).

PR #9, opened 2018-09-20
SimpLL: Updated DebugInfo to assign correct StructTypes as the keys of StructFieldNames in cases where their names differ.

lenticularis39:
Concurrent iteration through equivalent code in the other module is used to get the corresponding GEP instruction, from which the corresponding StructType is extracted. When the iteration fails at any level (for example by reaching the end prematurely), it is stopped and the old method of finding the corresponding StructType (searching for the type name in the other module) is used.

### DIFF
--- a/diffkemp/simpll/DebugInfo.cpp
+++ b/diffkemp/simpll/DebugInfo.cpp
@@ -79,23 +79,54 @@ void DebugInfo::calculateGEPIndexAlignments() {
         return;
 
     for (auto &Fun : ModFirst) {
+        auto OtherFun = ModSecond.getFunction(Fun.getName());
+
         if (CalledFirst.find(&Fun) == CalledFirst.end())
             continue;
 
+        auto OtherBB = OtherFun->begin();
         for (auto &BB : Fun) {
+            if(OtherBB == OtherFun->end())
+                // The end of the other function has been reached
+                break;
+
+            auto OtherInstr = OtherBB->begin();
             for (auto &Instr : BB) {
-                if (auto GEP = dyn_cast<GetElementPtrInst>(&Instr)) {
+                if(OtherInstr == OtherBB->end())
+                    // The end of the other basic block has been reached
+                    break;
+
+                auto GEP = dyn_cast<GetElementPtrInst>(&Instr);
+                auto OtherGEP = dyn_cast<GetElementPtrInst>(&*OtherInstr);
+
+                if (GEP && OtherGEP) {
                     std::vector<Value *> indices;
+
+                    std::vector<Value *> indicesOther;
+
+                    auto idx_other = OtherGEP->idx_begin();
                     // Iterate all indices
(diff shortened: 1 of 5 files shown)


diffkemp/simpll/DebugInfo.cpp
```diff
@@ -69,6 +69,135 @@ DICompositeType *DebugInfo::getStructTypeInfo(const StringRef name,
...
+
+                StringRef elementName =
+                        getElementNameAtIndex(*TypeDIFirst,
+                                                indexFirst);
+
+                // Find index of the element with the same name in
+                // the second module
+                auto TypeDISecond =
+                        getStructTypeInfo(typeName,
+                                            Program::Second);
```
viktormalik: Fix alignment

diffkemp/simpll/DebugInfo.cpp
```diff
@@ -69,6 +69,135 @@ DICompositeType *DebugInfo::getStructTypeInfo(const StringRef name,
...
+                // Get name of the element at the current index in
+                // the first module
+                auto TypeDIFirst =
+                        getStructTypeInfo(typeName, Program::First);
+                if (!TypeDIFirst)
+                    continue;
+
+                StringRef elementName =
+                        getElementNameAtIndex(*TypeDIFirst,
+                                                indexFirst);
```
viktormalik: Fix alignment.

diffkemp/simpll/DebugInfo.cpp
```diff
@@ -69,6 +69,135 @@ DICompositeType *DebugInfo::getStructTypeInfo(const StringRef name,
...
+                // the second module
+                auto TypeDISecond =
+                        getStructTypeInfo(typeName,
+                                            Program::Second);
+                if (!TypeDISecond)
+                    continue;
+
+                int indexSecond =
+                        getTypeMemberIndex(*TypeDISecond,
+                                            elementName);
```
viktormalik: Fix alignment.

diffkemp/simpll/DebugInfo.cpp
```diff
@@ -86,31 +86,32 @@ void DebugInfo::extractAlignmentFromInstructions(GetElementPtrInst *GEP,
...
-            if (indexedTypeOther && idx_other != OtherGEP->idx_end() &&
-                !indexedTypeOther->isStructTy()) {
-                indices.push_back(*idx_other);
-                continue;
+            if (indexedTypeOther && !indexedTypeOther->isStructTy()) {
+                // The pushing of idx_other is done at the end of the cycle, not
+                // in the for header, therefore it has to be pushed here
+                if (idx_other != OtherGEP->idx_end())
+                    indicesOther.push_back(*idx_other);
+
```
viktormalik: I think that we should either 'continue' here or do not insert *idx_other into indicesOther since this will be done in the end of the loop again.

PR #10, opened 2018-10-09
SimpLL: Fixed comparing kzalloc instructions - now comparing

lenticularis39:
structure types by names instead of comparing allocated memory size by
the raw value.
(Currently ignores kzalloc flags.)

Note: I don't have a test case yet, because all cases in the spreadsheet have another difference beside the allocation function inaccuracy.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -142,11 +142,9 @@ int DifferentialFunctionComparator::cmpAttrs(const AttributeList L,
 /// of the composite type is different
 int DifferentialFunctionComparator::cmpOperations(
     const Instruction *L, const Instruction *R, bool &needToCmpOperands) const {
-    int Result = FunctionComparator::cmpOperations(L, R, needToCmpOperands);
-
     // Check whether the instruction is a CallInst calling the function kzalloc.
     if (!isa<CallInst>(L) || !isa<CallInst>(R))
-        return Result;
+        return FunctionComparator::cmpOperations(L, R, needToCmpOperands);
 
     const CallInst *CL = dyn_cast<CallInst>(L);
     const CallInst *CR = dyn_cast<CallInst>(R);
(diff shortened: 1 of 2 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -137,3 +137,64 @@ int DifferentialFunctionComparator::cmpAttrs(const AttributeList L,
     }
     return FunctionComparator::cmpAttrs(LNew, RNew);
 }
+
+/// Handle comparing of memory allocation function in cases where the size
+/// of the composite type is different
+int DifferentialFunctionComparator::cmpOperations(
+    const Instruction *L, const Instruction *R, bool &needToCmpOperands) const {
+    int Result = FunctionComparator::cmpOperations(L, R, needToCmpOperands);
```
viktormalik: Wouldn't it make sense to run the original cmpOperations at the end of this function? Now, it is run every time, even when its result is not used.

diffkemp/simpll/DifferentialFunctionComparator.h
```diff
@@ -31,7 +31,10 @@ class DifferentialFunctionComparator : public FunctionComparator {
                                    const Function *F2,
                                    GlobalNumberState *GN,
                                    const DebugInfo *DI)
-            : FunctionComparator(F1, F2, GN), DI(DI) {};
+            : FunctionComparator(F1, F2, GN), DI(DI) {
+        LayoutL = new DataLayout(F1->getParent());
+        LayoutR = new DataLayout(F2->getParent());
```
viktormalik: Why allocating new DataLayout when you can get it directly from the module (F1->getParent()->getDataLayout())? This code would cause a memory leak because you do not free the memory anywhere.

diffkemp/simpll/DifferentialFunctionComparator.h
```diff
@@ -42,9 +45,15 @@ class DifferentialFunctionComparator : public FunctionComparator {
...
     int cmpAttrs(const AttributeList L, const AttributeList R) const override;
+    /// Handle comparing of memory allocation function in cases where the size
+    /// of the composite type is different
+    int cmpOperations(const Instruction *L, const Instruction *R,
+                    bool &needToCmpOperands) const override;
 
   private:
     const DebugInfo *DI;
+
+    DataLayout *LayoutL, *LayoutR;
```
viktormalik: I think it would be better to use a reference (DataLayout &) and to get the data layout from the module directly (see the note above).

diffkemp/simpll/DifferentialFunctionComparator.h
```diff
@@ -42,9 +45,15 @@ class DifferentialFunctionComparator : public FunctionComparator {
     /// Specific comparison of attribute lists.
     /// Attributes that do not affect the semantics of functions are removed.
     int cmpAttrs(const AttributeList L, const AttributeList R) const override;
+    /// Handle comparing of memory allocation function in cases where the size
+    /// of the composite type is different
+    int cmpOperations(const Instruction *L, const Instruction *R,
+                    bool &needToCmpOperands) const override;
```
viktormalik: Fix alignment.

PR #11, opened 2018-10-18
SimpLL: Assign the same number in DifferentialGlobalNumberState to di…

lenticularis39:
…fferent

print functions in the kernel.

Also modifies the behaviour of SimplifyKernelFunctionCallsPass to replace
arguments of printk with two null pointers instead of one so the calls of
printk and the other print functions can be compared as equal.

### DIFF
--- a/diffkemp/simpll/DifferentialGlobalNumberState.h
+++ b/diffkemp/simpll/DifferentialGlobalNumberState.h
@@ -17,6 +17,7 @@
 #define DIFFKEMP_SIMPLL_DIFFERENTIALGLOBALNUMBERSTATE_H
 
 #include <llvm/Transforms/Utils/FunctionComparator.h>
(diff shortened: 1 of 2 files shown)


diffkemp/simpll/DifferentialGlobalNumberState.h
```diff
@@ -42,6 +43,9 @@ class DifferentialGlobalNumberState : public GlobalNumberState {
     using IntegerMap = std::map<APInt, uint64_t, cmpConstants>;
     IntegerMap Constants{IntegerMap(cmpConstants())};
 
+    StringRef PrintFunctionList[5] = {"printk", "dev_warn", "dev_err",
+        "_dev_info", "sprintf"};
```
viktormalik: Use std::vector rather than a plain array.

PR #14, opened 2018-11-07
Download RHEL kernel sources from the CentOS Git when the internal

lenticularis39:
server with Brew is unavailable.

### DIFF
--- a/diffkemp/llvm_ir/build_llvm.py
+++ b/diffkemp/llvm_ir/build_llvm.py
@@ -13,6 +13,8 @@ import shutil
 from subprocess import CalledProcessError, call, check_call, check_output
 from subprocess import Popen, PIPE
 from urllib import urlretrieve
+from tempfile import mkdtemp
+from socket import gethostbyname, error as socket_error
 
 
 class BuildException(Exception):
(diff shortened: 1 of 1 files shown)


diffkemp/llvm_ir/build_llvm.py
```diff
@@ -124,6 +131,61 @@ def _get_kernel_tar_from_upstream(self):
...
+        os.chdir(self.kernel_base_path)
+        shutil.rmtree(tmpdir)
+
+        # Download the actual sources
+        for entry in filelist:
+            url = "https://git.centos.org/sources/kernel/c7/{}".format(entry[0])
+            filename = entry[1].split("/")[1]
+
+            urlretrieve(url, os.path.join(self.kernel_base_path, filename),
+                    show_progress)
```
viktormalik: Wouldn't it be better to clone the centos-common-git repo (https://git.centos.org/git/centos-git-common.git) and run the get_sources.sh script from there? It does the same as you do and if they change the way the files are downloaded, it will keep working.

diffkemp/llvm_ir/build_llvm.py
```diff
@@ -124,6 +131,61 @@ def _get_kernel_tar_from_upstream(self):
...
+        tarname = "linux-{}.el7.tar.xz".format(self.kernel_version)
+        self.kabi_tarname = "kernel-abi-whitelists-{}.tar.bz2".format(
+            self.kernel_version.split("-")[1])
+        for f in os.listdir("."):
+            if (os.path.isfile(f) and f != tarname and f != self.configfile and
+                    f != self.kabi_tarname):
+                os.remove(f)
+        self.configfile = os.path.abspath(self.configfile)
+        os.chdir(cwd)
+        return tarname
```
viktormalik: This code (from line 175) is essentially the same as in _get_kernel_tar_from_brew, it would be better to reuse it (put it in a separate function).

PR #15, opened 2018-11-15
Added regression tests for diffkabi.

lenticularis39:
Also moved building functions from diffkabi.py to build_llvm.py (as a function of LlvmKernelBuilder).


diffkemp/llvm_ir/build_llvm.py
```diff
@@ -690,6 +691,26 @@ def build_llvm_module(self, name, file_name, commands):
         mod = LlvmKernelModule(name, file_name, self.modules_path)
         return mod
 
+    def build_llvm_function(self, f):
```
viktormalik: Better call it build_file_for_function to be consistent with naming of other methods.

viktormalik: Comment the function with docstring.

tests/regression/diffkabi/diffkabi_test.py
```diff
@@ -0,0 +1,145 @@
+"""
+Regression testing using pytest.
+Individual tests are specified using YAML in the tests/regression/test_specs/
+directory. Each test describes a kernel module, two kernel versions between
+which the module is compared, and a list of compared module parameters. For
+each parameter, pairs of corresponding functions (using the parameter) along
+with the expected analysis results must be provided.
+This script parses the test specification and prepares testing scenarions for
+pytest.
+"""
```
viktormalik: Update the comment for diffkabi regression tests

tests/regression/diffkabi/diffkabi_test.py
```diff
@@ -0,0 +1,145 @@
...
+    Prepare testing task (scenario). Build old and new modules and copy
+    created files.
+    """
+    spec.built_functions = dict()
+
+    for function, desc in spec.functions.iteritems():
+        # Find the modules
+        first_builder = LlvmKernelBuilder(spec.old_kernel, None,
+                                          debug=spec.debug)
+        first_builder.source.build_cscope_database()
```
viktormalik: This is not needed, build_cscope_database is called by find_srcs_for_function.
Same below.

tests/regression/diffkabi/diffkabi_test.py
```diff
@@ -0,0 +1,147 @@
...
+        #
+        # Note: the files are copied to the task directory for reference only.
+        # The exact name of the module is not known before building the
+        # function, therefore building from module sources in kernel_modules or
+        # using already built LLVM IR files is not possible.
+        mod_path_old = mod_first.llvm
+        mod_path_new = mod_second.llvm
+        mod_old = os.path.basename(mod_first.name)
+        mod_new = os.path.basename(mod_second.name)
+        task_dir = os.path.join(tasks_path, mod_old)
```
viktormalik: Couldn't we name the directories by functions? It would be simpler I think.

tests/regression/diffkabi/diffkabi_test.py
```diff
@@ -0,0 +1,147 @@
...
+        old_llvm = os.path.join(task_dir, mod_old + "_old.ll")
+        new_llvm = os.path.join(task_dir, mod_new + "_new.ll")
+        old_source_src = os.path.splitext(mod_path_old)[0] + ".c"
+        new_source_src = os.path.splitext(mod_path_new)[0] + ".c"
+
+        shutil.copyfile(mod_path_old, old_llvm)
+        shutil.copyfile(old_source_src, old_src)
+
+        shutil.copyfile(mod_path_new, new_llvm)
+        shutil.copyfile(new_source_src, new_src)
```
viktormalik: Couldn't we use the prepare_module function from diffkabi/regression_test? It does basically the same copying, except that it builds a module instead of a single file for a function. I'd rather have this unified somehow.

tests/regression/diffkabi/diffkabi_test.py
```diff
@@ -0,0 +1,147 @@
...
+    prepare_task(spec)
+    return spec
+
+
+class TestClass(object):
+    """
+    Main testing class. One object of the class is created for each testing
+    task. Contains 2 tests for function couplings collection and for the
+    actual function analysis (semantic comparison).
+    """
```
viktormalik: Update comment.

tests/regression/diffkabi/diffkabi_test.py
```diff
@@ -0,0 +1,147 @@
...
+                                           debug=spec.debug)
+        second_builder.source.build_cscope_database()
+
+        # Build the modules
+        mod_first = first_builder.build_llvm_function(function)
+        mod_second = second_builder.build_llvm_function(function)
+
+        # Add the function to the dictionary
+        fun = (function, mod_first, mod_second)
+        spec.built_functions[fun] = Result[desc.upper()]
```
viktormalik: I think it would be better to use some named field of the TaskSpec class instead of a dictionary with a 3-tuple as a key.

tests/regression/diffkemp/clean_test.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
```
viktormalik: Isn't this file just moved? Shouldn't it be removed from the last location as well?

viktormalik:
Also remove the files from the kernel_modules/* directories that are now in .gitignore.

tests/regression/diffkemp/clean_test.sh
lenticularis39: Yes, the file was moved, and this was correctly detected by Git. It shows it here as being newly created because it was detected as being moved to tests/regression/diffkabi (which can be seen [here](https://github.com/viktormalik/diffkemp/pull/15/commits/bd9b04342eccbd6249be9470a8ebb997f3dc0a5d)), therefore the copy in tests/regression/diffkemp is seen as a new file. The same thing happened with regression_test.py.

viktormalik: Ah, I see, now we have two of these files.

diffkemp/test_tools/module_tools.py
```diff
@@ -0,0 +1,29 @@
+"""Tools for working with modules used by both diffkemp and diffkabi tests."""
```
viktormalik: I'd put this directly to the tests/regression directory.

tests/regression/diffkabi/diffkabi_test.py
```diff
@@ -1,19 +1,18 @@
 """
 Regression testing using pytest.
-Individual tests are specified using YAML in the tests/regression/test_specs/
-directory. Each test describes a kernel module, two kernel versions between
-which the module is compared, and a list of compared module parameters. For
-each parameter, pairs of corresponding functions (using the parameter) along
-with the expected analysis results must be provided.
+Individual tests are specified using YAML in the
+tests/regression/test_specs/diffkabi directory. Each test contains a list of
+KABI functions and two kernel versions between which the module is compared.
```
viktormalik: ...between which the functions are compared.

tests/regression/diffkabi/diffkabi_test.py
```diff
@@ -54,61 +58,53 @@ def __init__(self, spec):
...
+    #
+    # Note: the files are copied to the task directory for reference only.
+    # The exact name of the module is not known before building the
+    # function, therefore building from module sources in kernel_modules or
+    # using already built LLVM IR files is not possible.
+    mod_old = os.path.basename(old_module.name)
+    mod_new = os.path.basename(new_module.name)
+    spec.task_dir = os.path.join(tasks_path, spec.function)
+    if not os.path.isdir(tasks_path):
+        os.mkdir(tasks_path)
```
viktormalik: Is this necessary? We could suppose that this directory exists.

diffkemp/test_tools/module_tools.py
lenticularis39: I don't exactly know how to make Python find the file there. (I could turn the whole tests directory into a module, but then it could cause a conflict when including the diffkemp repository into another project.)

tests/regression/diffkabi/diffkabi_test.py
lenticularis39: I believe it is, because the directory isn't supposed to contain any files, therefore it isn't in the Git repository, so it has to be created.

viktormalik: I see, then let's do this somewhere else, so that we don't have to check its existence in each test spec.

diffkemp/test_tools/module_tools.py
viktormalik: I think it could be fine, because if someone includes the repo into another project, he would just use the 'diffkemp' package.

PR #20, opened 2018-12-13
diffsysctl tests

lenticularis39:
Added regression test support for diffsysctl and unit tests for the LlvmSysctlModule class.
Also fixed a bug in LlvmSysctlModule that caused processing GEPs without bitcats to lead to a type conversion error in LLVM (the function get_data attempted to get the const opcode of a non-ConstantExpr value).

### DIFF
--- a/diffkemp/llvm_ir/llvm_sysctl_module.py
+++ b/diffkemp/llvm_ir/llvm_sysctl_module.py
@@ -106,7 +106,8 @@ class LlvmSysctlModule:
                         all_constant = False
                 if all_constant:
                     data = data.get_operand(0)
-            if data.get_const_opcode() == Opcode['BitCast']:
+            if (data.get_kind() == ConstantExprValueKind and
+                data.get_const_opcode() == Opcode['BitCast']):
                 # Address is typed to "void *", we need to get rid of the
                 # bitcast operator.
                 data = data.get_operand(0)
--- a/diffkemp/llvm_ir/llvm_sysctl_module.py
+++ b/diffkemp/llvm_ir/llvm_sysctl_module.py
@@ -107,7 +107,7 @@ class LlvmSysctlModule:
                 if all_constant:
                     data = data.get_operand(0)
             if (data.get_kind() == ConstantExprValueKind and
-                data.get_const_opcode() == Opcode['BitCast']):
+                    data.get_const_opcode() == Opcode['BitCast']):
                 # Address is typed to "void *", we need to get rid of the
                 # bitcast operator.
                 data = data.get_operand(0)
--- a/diffkemp/llvm_ir/llvm_sysctl_module.py
+++ b/diffkemp/llvm_ir/llvm_sysctl_module.py
@@ -94,23 +94,23 @@ class LlvmSysctlModule:
         data = sysctl.get_operand(1)
         if data.is_null():
             return None
-        if data.get_kind() == ConstantExprValueKind:
-            if data.get_const_opcode() == Opcode['GetElementPtr']:
-                # Address is a GEP, we have to extract the actual variable.
-                all_constant = True
-                indices = list()
-                # Look whether are all indices constant.
-                for i in range(1, data.get_num_operands()):
-                    indices.append(data.get_operand(i).const_int_get_z_ext())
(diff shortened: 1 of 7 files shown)


diffkemp/llvm_ir/llvm_sysctl_module.py
```diff
@@ -106,7 +106,8 @@ def get_data(self, sysctl_name):
                         all_constant = False
                 if all_constant:
                     data = data.get_operand(0)
-            if data.get_const_opcode() == Opcode['BitCast']:
+            if (data.get_kind() == ConstantExprValueKind and
+                    data.get_const_opcode() == Opcode['BitCast']):
```
viktormalik: Use the same comparison pattern in the if condition above.

tests/regression/diffsysctl/diffsysctl_test.py
```diff
@@ -0,0 +1,220 @@
...
+    # Find the modules
+    first_builder = LlvmKernelBuilder(spec.old_kernel, None,
+                                      debug=spec.debug)
+    second_builder = LlvmKernelBuilder(spec.new_kernel, None,
+                                       debug=spec.debug)
+
+    sysctl_module = first_builder.build_sysctl_module(spec.sysctl)
+    spec.data_diffsysctl = sysctl_module.get_data(spec.sysctl)
+    spec.proc_handler_diffsysctl = \
+        sysctl_module.get_proc_fun(spec.sysctl)
```
viktormalik: I think we should call get_data and get_proc_fun in test_llvm_sysctl_module, not here.

tests/regression/diffsysctl/diffsysctl_test.py
```diff
@@ -0,0 +1,220 @@
...
+        self.old_kernel = spec["old_kernel"]
+        self.new_kernel = spec["new_kernel"]
+        self.sysctl = spec["sysctl"]
+        self.data = spec["data"]
+        self.data_functions = list()
+        for function, result in spec["data_functions"].iteritems():
+            self.data_functions.append(FunctionSpec(function, result))
+        self.proc_handler = list()
+        for function, result in spec["proc_handler"].iteritems():
+            self.proc_handler.append(FunctionSpec(function, result))
```
viktormalik: There is always a single proc_handler function, no need to use iteritems() and store it in a list.

tests/regression/diffsysctl/diffsysctl_test.py
lenticularis39: get_proc_fun is OK to move to test_llvm_sysctl_module, but the variable spec.data_diffsysctl has to be created in the prepare_task function, because functions_diff needs a KernelParam object with the indices inside it - this is not in the spec file, therefore it has to be extracted using LlvmSysctlModule from the sysctl itself.

tests/regression/diffsysctl/test_specs/rhel-75-76.yaml
```diff
@@ -0,0 +1,19 @@
...
+    proc_handler:
+      proc_do_rss_key: equal_syntax
+    data_variable:
+      name: netdev_rss_key
+      functions:
+        netdev_rss_key_fill: equal_syntax
+        proc_do_rss_key: equal_syntax
+  - sysctl: net.core.message_burst
+    proc_handler:
+      proc_dointvec: skipped
```
viktormalik: I'd suggest calling this "generic" instead of "skipped".

tests/regression/diffsysctl/diffsysctl_test.py
```diff
@@ -0,0 +1,228 @@
...
+        variable associated with the sysctl parameter and the proc_handler
+        function.
+        For each compared function, the module is first simplified using the
+        SimpLL tool and then the actual analysis is run. Compares the obtained
+        result with the expected one.
+        If timeout is expected, the analysis is not run to increase testing
+        speed.
+        """
+        if (task_spec.proc_handler.result != Result.TIMEOUT
+                and task_spec.proc_handler.result != "skipped"):
```
viktormalik: I'd suggest using "generic" instead of "skipped". Same below.

PR #29, opened 2019-01-29
SimpLL: Remove function sections when comparing the control flow only.

lenticularis39:
This is to fix the function usleep_range comparing as not equal. 

There is another problem with a return value when comparing the same function, which could be solved by extending the RemoveUnusedReturnValuesPass to all functions. (I didn't include that here because I'm not currently sure whether this causes any problems or not, that yet has to be tested.)

### DIFF
--- a/diffkemp/simpll/Transforms.cpp
+++ b/diffkemp/simpll/Transforms.cpp
@@ -21,6 +21,7 @@
 #include "passes/CalledFunctionsAnalysis.h"
 #include "passes/ControlFlowSlicer.h"
 #include "passes/FunctionAbstractionsGenerator.h"
+#include "passes/ReduceFunctionMetadataPass.h"
 #include "passes/RemoveLifetimeCallsPass.h"
 #include "passes/RemoveUnusedReturnValuesPass.h"
 #include "passes/SimplifyKernelFunctionCallsPass.h"
@@ -114,6 +115,8 @@ std::vector<FunPair> simplifyModulesDiff(Config &config) {
     PassManager<Module, AnalysisManager<Module, Function *>, Function *,
         Module *> mpm;
     mpm.addPass(RemoveUnusedReturnValuesPass {});
+    if (config.ControlFlowOnly)
+        mpm.addPass(ReduceFunctionMetadataPass {});
     mpm.run(*config.First, mam, config.FirstFun, config.Second.get());
     mpm.run(*config.Second, mam, config.SecondFun, config.First.get());
 
--- a/diffkemp/simpll/Transforms.cpp
+++ b/diffkemp/simpll/Transforms.cpp
@@ -73,6 +73,7 @@ void preprocessModule(Module &Mod,
     fpm.addPass(UnifyMemcpyPass {});
     fpm.addPass(DCEPass {});
     fpm.addPass(LowerExpectIntrinsicPass {});
+    fpm.addPass(ReduceFunctionMetadataPass {});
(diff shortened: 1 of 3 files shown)


diffkemp/simpll/Transforms.cpp
```diff
@@ -114,6 +115,8 @@ std::vector<FunPair> simplifyModulesDiff(Config &config) {
     PassManager<Module, AnalysisManager<Module, Function *>, Function *,
         Module *> mpm;
     mpm.addPass(RemoveUnusedReturnValuesPass {});
+    if (config.ControlFlowOnly)
+        mpm.addPass(ReduceFunctionMetadataPass {});
```
viktormalik: I wouldn't do this for control-flow-only, but always. Section attributes should not be relevant for the function semantics.

diffkemp/simpll/passes/ReduceFunctionMetadataPass.h
```diff
@@ -0,0 +1,30 @@
...
+
+using namespace llvm;
+
+/// A pass that transforms functions returning some value to void in case their
+/// return value is never used.
+class ReduceFunctionMetadataPass
+        : public PassInfoMixin<ReduceFunctionMetadataPass> {
+  public:
+    PreservedAnalyses run(Module &Mod, AnalysisManager<Module, Function *> &mam,
+                          Function *Main, Module *ModOther);
```
viktormalik: Why module pass? It should be a function pass.

diffkemp/simpll/Transforms.cpp
viktormalik: Also, this should be placed in preprocessModule.

diffkemp/simpll/passes/ReduceFunctionMetadataPass.cpp
```diff
@@ -0,0 +1,28 @@
...
+///
+/// \file
+/// This file contains the declaration of the ReduceFunctionMetadataPass pass.
+///
+//===----------------------------------------------------------------------===//
+
+#include "CalledFunctionsAnalysis.h"
+#include "ReduceFunctionMetadataPass.h"
+#include "Utils.h"
+#include <llvm/IR/Instructions.h>
```
viktormalik: I think that these includes are unused, except for the header of this pass.

diffkemp/simpll/passes/ReduceFunctionMetadataPass.cpp
```diff
@@ -0,0 +1,25 @@
+//===----  ReduceFunctionMetadataPass.h - Removes some function metadata --===//
```
viktormalik: ReduceFunctionMetadataPass.cpp

diffkemp/simpll/passes/ReduceFunctionMetadataPass.cpp
```diff
@@ -0,0 +1,25 @@
...
+//
+//       SimpLL - Program simplifier for analysis of semantic difference      //
+//
+// This file is published under Apache 2.0 license. See LICENSE for details.
+// Author: Tomas Glozar, tglozar@gmail.com
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file contains the declaration of the ReduceFunctionMetadataPass pass.
```
viktormalik: 'definition' instead of 'declaration' plus brief description of what the pass currently does (copy the description from below).

PR #35, opened 2019-02-20
Optimization of diffkemp tests

lenticularis39:
Removed some unnecessary tests for reasons described in the commit. Also reduced the test b2c2-flexcop, which is necessary, but running it on one function is enough (there was a lot of functions, so it took a long time to complete).


tests/regression/diffkemp/test_specs/b2c2-flexcop.yaml
```diff
@@ -8,19 +8,19 @@ debug: true
...
-      flexcop_pid_control: equal_syntax
-      flexcop_set_tone: equal_syntax
-      flexcop_dvb_exit: equal_syntax
-      flexcop_determine_revision: equal_syntax
-      flexcop_master_xfer: equal_syntax
-      flexcop_i2c_request: equal_syntax
-      flexcop_reset_block_300: equal_syntax
-      flexcop_rcv_data_ctrl: equal_syntax
-      flexcop_dump_reg: equal_syntax
+      flexcop_pid_control: none
```
viktormalik: Better call it "skip". Use here the same pattern as we do in the DiffSysctl tests with the "generic" value (setting the expected result to Result.Kind.NONE).

PR #37, opened 2019-03-05
SimpLL: Modified RemoveUnusedReturnValuesPass to be able to

### LINKED ISSUE #33: Change in function return type from void to non-void should not be a difference
This applies to cases when the only difference are returns from the new function. In the old version, no one could check the return value, so changing the return type cannot break anything.
This occurs for KABI whitelist symbol `__napi_complete`.

lenticularis39:
perform the same thing on the main function.

Also added code to remove kernel symbol table structures, because
they are not needed for the comparison and are interfering with
the aforementioned pass.

Fixes #33

### DIFF
--- a/diffkemp/simpll/Transforms.cpp
+++ b/diffkemp/simpll/Transforms.cpp
@@ -113,9 +113,6 @@ std::vector<FunPair> simplifyModulesDiff(Config &config) {
             mam.getResult<FunctionAbstractionsGenerator>(*config.Second,
                                                          config.SecondFun));
 
-    StringRef FirstFunName = config.FirstFun->getName();
-    StringRef SecondFunName = config.SecondFun->getName();
-
     // Module passes
     PassManager<Module, AnalysisManager<Module, Function *>, Function *,
         Module *> mpm;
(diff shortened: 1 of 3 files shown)


diffkemp/simpll/Transforms.cpp
```diff
@@ -113,13 +113,35 @@ std::vector<FunPair> simplifyModulesDiff(Config &config) {
...
+                config.First->getFunction(FirstFunName);
+    }
+
+    if (config.SecondFun !=
+        config.Second->getFunction(SecondFunName)) {
+        // Main function was replaced by a pass - fix the function pointer
+        NewSecondFun =
+                config.Second->getFunction(SecondFunName);
+    }
+
```
viktormalik: This seems a bit over-complicated. Wouldn't it be sufficient, after running the pass, to do `config.FirstFun = config.First.getFunction(FirstFunName)`?

lenticularis39: That was the original idea, but I think that the pointer to the original function is needed to get the source file path (somehow the `Fun->getSubprogram()` seems not to work with the new function - it's probably because it technically isn't a part of the original file), so I have to keep the pointer to it.

lenticularis39: Also (kind of unfortunately) while the getting of the source file needs the old function, the getting of its name needs the new function (because the old function doesn't have a name to avoid a conflict), that's why I keep both.

viktormalik: Could you copy the pointer to the subprogram from the old function to the new one (there is a method `Fun->setSubprogram()`? If this would suffice, the code would be much simpler.

lenticularis39: That would be good if it suffices, because then I could also exclude the replacing of the main function in `ComparedFuns`.

diffkemp/simpll/Transforms.cpp
```diff
@@ -149,7 +169,7 @@ std::vector<FunPair> simplifyModulesDiff(Config &config) {
             }
         }
         if (allEqual) {
-            // Functions are equal iff all functions that were compared by
+            // Functions are equal if all functions that were compared by
```
viktormalik: This wasn't a typo, "iff" is an abbreviation for "if and only if".

diffkemp/simpll/Transforms.cpp
```diff
@@ -113,13 +113,33 @@ std::vector<FunPair> simplifyModulesDiff(Config &config) {
...
+                config.First->getFunction(FirstFunName);
+    }
+
+    if (config.SecondFun !=
+        config.Second->getFunction(SecondFunName)) {
+        // Main function was replaced by a pass - fix the function pointer
+        config.SecondFun =
+                config.Second->getFunction(SecondFunName);
+    }
+
```
viktormalik: There is a method Config::refreshFunctions() available. It should do exactly what you need.

PR #48, opened 2019-05-09
SimpLL: Extending ignoring of casts when comparing control flow

lenticularis39:
only to constant expressions.

This fixes the function debugfs_create_u32 comparing as non-equal.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -401,6 +401,38 @@ int DifferentialFunctionComparator::cmpValues(const Value *L,
     return result;
 }
 
+/// Specific comparing of constants. If one of them (or both) is a cast
+/// constant expression, compare its interal value.
+int DifferentialFunctionComparator::cmpConstants(const Constant *L,
+        const Constant *R) const {
+    int Result = FunctionComparator::cmpConstants(L, R);
+
+    if (Result == 0)
+        return Result;
+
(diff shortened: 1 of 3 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -401,6 +401,38 @@ int DifferentialFunctionComparator::cmpValues(const Value *L,
     return result;
 }
 
+/// Specific comparing of constants. If one of them (or both) is a cast
+/// constant expression, compare its interal value.
```
viktormalik: Should be "integral value"?

diffkemp/simpll/DifferentialFunctionComparator.h
```diff
@@ -72,6 +72,9 @@ class DifferentialFunctionComparator : public FunctionComparator {
     /// Specific comparing of values. Handles values generated from macros
     /// whose value changed and values where at least one of them is a cast.
     int cmpValues(const Value *L, const Value *R) const override;
+    /// Specific comparing of constants. If one of them (or both) is a cast
+    /// constant expression, compare its interal value.
```
viktormalik: Same as above

diffkemp/simpll/DifferentialFunctionComparator.cpp
lenticularis39: It's actually a different typo - I meant "internal", nevertheless it's better to call it "operand".

PR #53, opened 2019-05-22
SimpLL: Extend the code for macro differences to cover differences

lenticularis39:
in inline assembly (to be able to show a diff when there is a
differering assembly generated from macros and the difference
cannot be found by following the macros).

This fixes an empty diff when comparing the function `slab_alloc_node` between kernel version 3.10.0-514.el7 and 3.10.0-693.el7.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -240,8 +240,8 @@ int DifferentialFunctionComparator::cmpOperations(
 
     if (Result) {
         auto macroDiffs = findMacroDifferences(L, R);
-        ModComparator->DifferingMacros.insert(
-            ModComparator->DifferingMacros.end(),
+        ModComparator->DifferingObjects.insert(
+            ModComparator->DifferingObjects.end(),
             macroDiffs.begin(), macroDiffs.end());
     }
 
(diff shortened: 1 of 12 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -350,6 +364,50 @@ int DifferentialFunctionComparator::cmpBasicBlocks(const BasicBlock *BBL,
...
+        const Value *L, const Value *R, const Function *ParentL,
+        const Function *ParentR) const {
+    auto FunL = dyn_cast<Function>(L);
+    auto FunR = dyn_cast<Function>(R);
+
+    if (!FunL || !FunR)
+        // Both values have to be functions
+        return {};
+
+    if (!FunL->getName().startswith("simpll__inlineasm") ||
```
viktormalik: Define a macro for the "simpll_inlineasm" string so that it can be changed at a single place. I'd define it in passes/FunctionAbstractionsGenerator.h.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -350,6 +364,50 @@ int DifferentialFunctionComparator::cmpBasicBlocks(const BasicBlock *BBL,
     return 0;
 }
 
+std::vector<SyntaxDifference> DifferentialFunctionComparator::findAsmDifference(
```
viktormalik: Add doxygen comment for the function.

diffkemp/simpll/MacroUtils.h
```diff
@@ -39,14 +39,14 @@ struct MacroElement {
     std::string sourceFile;
 };
 
-/// Difference between macro in left module and the same macro in the right
-/// module.
-struct MacroDifference {
-	// Name of the macro.
-	std::string macroName;
+/// Syntactic difference between objects that cannot be found in the original
+/// source files
```
viktormalik: I'd add here a note that these differences can be either in macros or in inline assemblies.

diffkemp/simpll/passes/FunctionAbstractionsGenerator.h
```diff
@@ -18,6 +18,9 @@
 #include <llvm/IR/PassManager.h>
 #include <set>
 
+const std::string inlineAsmPrefix = "simpll__inlineasm_";
+const std::string indirectFunctionPrefix = "simpll__indirect_";
```
viktormalik: Use better names to express that this is a SimpLL-specific constant:
SimpllInlineAsmPrefix and SimpllIndirectFunctionPrefix.

PR #54, opened 2019-05-22
Added option to filter diffs by regex to diffkabi.

lenticularis39:
Function whose diff does not contain the specified regular
expression is treated as equal.

### DIFF
--- a/diffkemp/diffkabi.py
+++ b/diffkemp/diffkabi.py
@@ -11,6 +11,7 @@ from diffkemp.semdiff.result import Result
 import os
 import shutil
 import sys
+import re
 
 
 def __make_argument_parser():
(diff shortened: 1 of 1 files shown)


diffkemp/diffkabi.py
```diff
@@ -112,6 +115,13 @@ def run_from_cli():
                         glob_first=globvar, glob_second=globvar, config=config)
 
                 if fun_result is not None:
+                    if args.regex_filter is not None:
+                        pattern = re.compile(args.regex_filter)
+                        for called_res in fun_result.inner.itervalues():
+                            if not pattern.search(called_res.diff):
+                                fun_result.diff = ""
+                                fun_result.kind = Result.Kind.EQUAL_SYNTAX
```
viktormalik: If I understand it correctly, you ignore the result if **one** of the called function diffs does not match the regex. I think it should be the other way, the result is only ignored if **none** of the called function diffs matches the regex.
I would use the `for ... else` construct for this.

viktormalik:
Rebase into a single commit please.

PR #61, opened 2019-07-04
Introduce new interface for DiffKemp.

viktormalik:
This allows to compare kernels incrementally.
The new interface separates generation of a "snapshot" (necessary LLVM files) and actual semantic comparison.
We introduce a lot of major changes:
- All tools (`diffkemp`, `diffsysctl`, `diffkabi`) are unified into a single one (`diffkemp`). Currently, only diff of a list of functions is supported, diff of sysctl and kernel module parameters will be added later.
- CLI interface is changed (separated `generate` and `compare` commands).
- `KernelSource` class is now the main one to work with. It contains and calls a `LlvmKernelBuilder` for building sources into LLVM IR. Config now also contains `KernelSources`.
- Methods for creating the snapshot are introduced:
    - Copying LLVM, C sources and CScope files.
    - New class FunctionList for representing the function list with YAML import/export.
- Some functionality is removed from LlvmKernelBuilder. It now only contains methods for building the sources. Methods for downloading and configuration of kernel were moved into a separate tool (see https://github.com/viktormalik/rhel-kernel-get).
- A number of unnecessary Python classes are removed:
    - `FunctionCollector`: functionality is moved to `LlvmKernelModule`.
    - `FunctionCouplings`: couplings are now computed directly in `function_diff` based on name equality.
    - `ModuleAnalyzer`: not used anymore.
- Unit tests are updated
- Regression tests are reworked
    - The three test directories are replaced by three testing scripts.
    - All scripts share common spec files - one spec for each pair of kernels.
- See commit messages for more info.


diffkemp/diffkemp.py
```diff
@@ -1,143 +1,259 @@
...
-                                           args.debug, args.rebuild)
-        if args.module:
-            second_mods = {
-                args.module: second_builder.build_module(args.module)
-            }
-        elif args.all_modules:
-            second_mods = second_builder.build_all_modules()
-        elif args.file:
-            second_mods = {args.file: second_builder.build_file(args.file)}
+    args.func(args)
```
lenticularis39: This does not work when diffkemp is run without any arguments (the func attribute is not created in that case). Setting the required attribute on the object returned by `add_subparsers` should fix this issue and notify the user in case that they have to specify the subparser.

diffkemp/simpll/Utils.cpp
```diff
@@ -86,7 +86,10 @@ std::string dropSuffix(std::string Name) {
 std::string getFileForFun(Function *Fun) {
     if (auto *SubProgram = Fun->getSubprogram()) {
         if (auto *File = SubProgram->getFile()) {
-            return File->getDirectory().str() + "/" + File->getFilename().str();
+            auto Directory = File->getDirectory();
```
lenticularis39: A similar change should be done to the `getSourceFilePath` function (it does essentially the same thing as `getFileForFun`, only for a scope instead of a function).

README.md
```diff
@@ -127,6 +104,7 @@ Additionally if you are not using the development container:
 
 ### Tests
 
-The project contains unit and regression testing using pytest that can be run by:
+The project contains unit and regression testing using pytest that can be run
+by:
 
     pytest tests
```
lenticularis39: Maybe it could be mentioned here that one needs to have the appropriate kernel sources downloaded and configured to run the tests and how to do that.

PR #63, opened 2019-07-18
SimpLL: Fixed function-macro difference reporting.

lenticularis39:
There were two problems - the first was the skipping of the whole block
of code including both function inlining and function-macro difference
reporting when one of the calls was a SimpLL abstraction (this should
apply only to the inlining), the second was the missing implementation
of the reporting when one of the instruction wasn't a call instruction.

(Both problems caused false empty diffs in some functions.)

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -210,63 +210,8 @@ int DifferentialFunctionComparator::cmpOperations(
                             !CalledR->getName().startswith("simpll"))
                         ModComparator->tryInline = {CL, CR};
 
-                    // When trying to inline, it is possible that in one of the
-                    // versions a function correspond to a macro in the other.
-                    // This is fixed by inlining when the code is actually
-                    // equal, but would lead to a difference being reported in a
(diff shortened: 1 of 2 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -278,6 +280,67 @@ int DifferentialFunctionComparator::cmpOperations(
                      dyn_cast<CallInst>(R)->getCalledValue())->getName().
                      startswith("simpll"))
                 ModComparator->tryInline = {nullptr, dyn_cast<CallInst>(R)};
+
+            // The equivalent of the case above (function changed to a macro)
+            // can also happen in a way that will result in only one function
+            // call (for example when the macro does not start with a call
+            // instruction).
```
viktormalik: Since this is an equivalent of the above, isn't it possible to make a function from this code? The two pieces are very similar.

lenticularis39: At first I thought the code is quite different, but after looking at it closely it's clear that the code for the case where both instructions are CallInsts is universal for all cases, therefore I moved it into a function and used it in the second two cases.

PR #71, opened 2019-07-25
Add inline assembly arguments to syntax diff.

lenticularis39:
This PR adds code that attempts to reconstruct the original arguments for inline assembly in a C-like syntax to be shown alongside the differing assembly code.

There are several cases where this produces slightly different code from the original source, since some of the information is lost when compiling from C to LLVM IR; however it is sufficient to enable the user to identify the actual arguments in the C source code.

The function supports both local and global variables (including function arguments), load instructions, casts, extensions and has an incomplete support for GEPs.

The code uses DebugInfo to get GEP index names - because of this StructFieldNames generating was extended to include cases when indices are equal. (The cases that do not work, i.e. show the index number, are constant expressions - there is no debug information related to them).


viktormalik:
It seems there are some merge conflicts.

PR #75, opened 2019-07-30
Support comparing sysctl options

viktormalik:
This mainly requires changing layout of the YAML file generated in the
snapshot:
 - functions can be split into groups (by sysctl option)
 - apart from LLVM file, each function has a global variable (in case it
   is a function using the sysctl data variable) and a tag that contains
   explanation why the function belongs to the given group
 - in case only a list of functions is compared, a single group with
   `None` key is created

Also, `generate` and `compare` functions are reworked to handle
functions split into groups.


lenticularis39:
The unit tests are OK (I already aproved the PR, so I cannot approve it again).

PR #80, opened 2019-08-07
Extend statistics reporting.

lenticularis39:
In addtional to statistics about compared symbols, print
statistics about unique differences, too.

### DIFF
--- a/diffkemp/semdiff/result.py
+++ b/diffkemp/semdiff/result.py
@@ -147,7 +147,7 @@ class Result:
         empty = len([r for r in unique_diffs if r.res.diff == ""])
 
         # Print statistics
-        print("Total differences:  {}".format(total))
+        print("Total differences:       {}".format(total))
         print("In functions:            {0} ({1:.0f}%)".format(functions,
               functions / total * 100))
         print("In macros:               {0} ({1:.0f}%)".format(macros,
--- a/diffkemp/semdiff/result.py
+++ b/diffkemp/semdiff/result.py
@@ -62,9 +62,9 @@ class Result:
         # a higher priority is chosen from the two).
         self.kind = Result.Kind(max(int(self.kind), int(result.kind)))
 
-    def report_stat(self, show_errors=False):
+    def report_symbol_stat(self, show_errors=False):
         """
-        Report statistics.
+        Report symbol statistics.
         Print numbers of equal, non-equal, unknown, and error results with
         percentage that each has from the total results.
         """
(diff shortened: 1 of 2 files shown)


diffkemp/semdiff/result.py
```diff
@@ -84,18 +84,17 @@ def report_stat(self, show_errors=False):
             lambda x: x.diff == "", r.inner.values())) and
             r.kind == Result.Kind.NOT_EQUAL])
         if total > 0:
-            print("Total params: {}".format(total))
-            print("Equal:        {0} ({1:.0f}%)".format(eq, eq / total * 100))
-            print(" same syntax: {0}".format(eq_syn))
-            print("Not equal:    {0} ({1:.0f}%)".format(
+            print("Total symbols: {}".format(total))
+            print("Equal:         {0} ({1:.0f}%)".format(eq, eq / total * 100))
+            print(" same syntax:  {0}".format(eq_syn))
```
viktormalik: Remove "same syntax" part, we're not distinguishing `equal` and `equal_syntax` at the moment.

diffkemp/semdiff/result.py
```diff
@@ -109,3 +108,60 @@ def report_stat(self, show_errors=False):
...
+                      r.res.first.name.startswith("assembly code"))])
+        empty = len([r for r in unique_diffs if r.res.diff == ""])
+
+        # Print statistics
+        print("Total differences:  {}".format(total))
+        print("In functions:       {0} ({1:.0f}%)".format(functions,
+              functions / total * 100))
+        print("In macros:          {0} ({1:.0f}%)".format(macros,
+              macros / total * 100))
+        print("In inline assembly: {0} ({1:.0f}%)".format(asm,
```
viktormalik: "In inline assemblies". Let's use plural everywhere.

diffkemp/semdiff/result.py
```diff
@@ -109,3 +108,60 @@ def report_stat(self, show_errors=False):
...
+              asm / total * 100))
+        print("Empty:              {0} ({1:.0f}%)".format(empty,
```
viktormalik: "Empty diffs". So that it is more clear that we're talking about `diff` here.

diffkemp/semdiff/result.py
lenticularis39: Actually neither "inline assembly" or "inline assemblies" is correct (the first should be "inline assembly code", because "assembly" refers to the **process**, not to the result; the second one is grammatically incorrect, because "assembly" in this sense is uncountable).

viktormalik: Agreed, inline assembly code is correct.

PR #83, opened 2019-08-08
Added debugging scripts.

lenticularis39:
Note: the script is supposed to be used like this:
```bash
gdb --batch --command=gdb/returns.gdb --args <simpll_command>
````
Additional functions may use `PrintReturnValueBreakpoint` in case they compare subclasses of `cmpValues`; otherwise a new class has to be written (possibly creating a parent class from which both `PrintReturnValueBreakpoint` and the new class will inherit) alongside a dump function for the type.

### DIFF
--- a/diffkemp/simpll/Utils.cpp
+++ b/diffkemp/simpll/Utils.cpp
@@ -547,3 +547,13 @@ std::string getIdentifierForValue(const Value *Val,
     } else
         return "<unknown>";
 }
+
+/// Converts value to its string representation.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: gdb/returns.gdb [hunk 1]
```diff
@@ -0,0 +1,73 @@
+# *** Script for debugging return values for methods in FunctionComparator ***
```
viktormalik: Add how to use this script into the comment.

PR #84, opened 2019-08-09
Bring back unknown functions.

lenticularis39:
Previously functions whose source code wasn't found by CScope
were not included into snapshots; this commit changes this in a
way that they are now included with their LLVM IR field set to
None, enabling adding them into statistics as unknown in the
compare phase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ The command compares functions from function lists stored inside the snapshots
 pairwise and prints syntactic diffs (thanks to the `--syntax-diff` option) of
 functions that are semantically different.
 
-Note: if `FUNCTION_LIST` contains any other symbols than functions (e.g. global
+Note: if `FUNCTION_LIST` contains any symbols other than functions (e.g. global
 variables), they will be ignored.
 
 ### Comparing sysctl options
--- a/README.md
+++ b/README.md
@@ -21,6 +21,9 @@ The command compares functions from function lists stored inside the snapshots
 pairwise and prints syntactic diffs (thanks to the `--syntax-diff` option) of
 functions that are semantically different.
 
+Note: if `FUNCTION_LIST` contains any other symbols than functions (e.g. global
+variables), they will be ignored.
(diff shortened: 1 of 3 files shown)


diffkemp/diffkemp.py
```diff
@@ -237,8 +238,17 @@ def compare(args):
...
+                continue
+
+            # Check if the module exists in both snapshots
+            if old_fun_desc.mod is None or new_fun_desc.mod is None:
+                result.add_inner(Result(Result.Kind.UNKNOWN, fun, fun))
+                print("{}: unknown".format(fun))
+                continue
+
+            # Check if the function exists in both modules
+            if not (old_fun_desc.mod.has_function(fun) and
```
viktormalik: I suggest we could move this check into the `generate` phase, i.e., do not include symbols that are not functions into `functions.yaml`. I believe that currently, we include global variables and as the LLVM file we give the one where the variable is defined, which is useless. Also, this would be more efficient (it is expected that `compare` is run more often than `generate`).

README.md
```diff
@@ -21,6 +21,9 @@ The command compares functions from function lists stored inside the snapshots
 pairwise and prints syntactic diffs (thanks to the `--syntax-diff` option) of
 functions that are semantically different.
 
+Note: if `FUNCTION_LIST` contains any other symbols than functions (e.g. global
```
viktormalik: I think the correct wording is: "...any symbols other than functions..."

diffkemp/diffkemp.py
lenticularis39: That's a good idea, the LLVM IR parsing also takes unnecessary time.

viktormalik: Well, `has_function` only searches for regex, but it still takes some time. 
Can you include this into this PR?

lenticularis39: I've included both this and the better wording.

PR #86, opened 2019-08-13
SimpLL: Extend anonymous union numbering to structures.

lenticularis39:
The original version of the pass didn't include this, because it
is not without problems (e.g. when there is a change of a
structure insinde another stucture, this wouldn't work), but it
turned out to be useful.


diffkemp/simpll/passes/StructHashGeneratorPass.h
```diff
@@ -7,25 +7,25 @@
...
-#ifndef DIFFKEMP_SIMPLL_UNIONHASHGENERATORPASS_H
-#define DIFFKEMP_SIMPLL_UNIONHASHGENERATORPASS_H
+#ifndef DIFFKEMP_SIMPLL_STRUCTHASHGENERATORPASS_H
+#define DIFFKEMP_SIMPLL_STRUCTHASHGENERATORPASS_H
 
 #include <llvm/IR/PassManager.h>
 
 using namespace llvm;
 
 /// This pass renames all anonymous union types to names with hashes
```
viktormalik: ...renames all anonymous struct and union types...

PR #91, opened 2019-08-21
SimpLL: Added pass to unify globals numbering.

### LINKED ISSUE #90: Static variable defined in a different scope gets different number
In `kernel/workqueue.c` (function `init_worker_pool`) the macro `mutex_init` is used two times in a row:
```c
mutex_init(&pool->manager_arb);
mutex_init(&pool->manager_mutex);
```
This expands to (with formatting):
```c
do {
    static struct lock_class_key __key;
    __mutex_init((&pool->manager_arb), "&pool->manager_arb", &__key); 
} while (0);
do {
    static struct lock_class_key __key;
    __mutex_init(&pool->manager_mutex), "&pool->manager_mutex", &__key); 
} while (0);
```
Here the first `__key` and the second one are obviously a different global variable (there are defined in different blocks). The problem is that Clang assigns different suffixes to the second one for some unknown reason (`@init_worker_pool.__key.55` in 3.10.0-957.el7, but `@init_worker_pool.__key.71` in 3.10.0-1062.el7), which causes SimpLL to report a difference in these values.

lenticularis39:
Currently this pass implements fixing the numbering of local
static variables, where Clang generates different number suffixes
in different modules in cases there is more than one such variable
with the same name in one function.

Fixes #90.

### DIFF
--- a/diffkemp/simpll/Transforms.cpp
+++ b/diffkemp/simpll/Transforms.cpp
@@ -28,6 +28,7 @@
 #include "passes/SimplifyKernelGlobalsPass.h"
 #include "passes/StructHashGeneratorPass.h"
 #include "passes/StructureSizeAnalysis.h"
+#include "passes/UnifyGlobalsNumberingPass.h"
 #include "passes/UnifyMemcpyPass.h"
 #include "passes/VarDependencySlicer.h"
 #include <llvm/IR/PassManager.h>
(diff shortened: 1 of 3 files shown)


viktormalik:
The issue this PR solves should be covered by #93. I believe that it can be closed.

lenticularis39:
Redundant, issue covered by #93.

PR #92, opened 2019-08-21
SimpLL: Indent info during comparison

### LINKED ISSUE #66: SimpLL: indent info about analysed functions
Add indentation to debugging prints in `ModuleComparator` so that it is easier to see how the functions call each other.
Every time an analysis of a new function starts, increase indentation by 2. When the analysis ends, decrease indentation.
Use the indentation for all prints during the function comparison (`DifferentialFunctionComparator`, `SourceCodeUtils`).

PSilling:
Adds indentation to debugging prints during function comparison. Each such print should be moved by two spaces if referenced from another compared function.

Closes #66.

### DIFF
--- a/diffkemp/simpll/ModuleComparator.cpp
+++ b/diffkemp/simpll/ModuleComparator.cpp
@@ -29,7 +29,7 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                     dbgs() << getDebugIndent() << "Comparing "
                            << FirstFun->getName() << " and "
                            << SecondFun->getName() << "\n";
-                    increaseIndentLevel());
+                    increaseDebugIndentLevel());
     ComparedFuns.emplace(std::make_pair(FirstFun, SecondFun), Result::UNKNOWN);
 
     // Comparing function declarations (function without bodies).
@@ -57,22 +57,38 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
             }
         }
 
-        DEBUG_WITH_TYPE(DEBUG_SIMPLL, decreaseIndentLevel());
+        DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+            decreaseDebugIndentLevel();
+            if (ComparedFuns.at({FirstFun, SecondFun}) == Result::EQUAL) {
+                dbgs() << getDebugIndent() << "Declarations with matching "
+                       << "names, assuming they are equal\n";
+            } else if (ComparedFuns.at({FirstFun, SecondFun})
+                          == Result::NOT_EQUAL) {
+                dbgs() << getDebugIndent() << "Declarations without matching "
+                       << "names, assuming they are not equal\n";
+        });
+
         return;
     }
 
     // Comparing functions with bodies using custom FunctionComparator.
     DifferentialFunctionComparator fComp(FirstFun, SecondFun, controlFlowOnly,
                                          showAsmDiffs, DI, this);
-    if (fComp.compare() == 0) {
+    int result = fComp.compare();
+
+    DEBUG_WITH_TYPE(DEBUG_SIMPLL, decreaseDebugIndentLevel());
+    if (result == 0) {
         DEBUG_WITH_TYPE(DEBUG_SIMPLL,
-                        dbgs() << getDebugIndent(-1) << "Function "
-                               << FirstFun->getName()
-                               << " is same in both modules\n");
+                        dbgs() << getDebugIndent() << "Functions are equal\n");
         ComparedFuns.at({FirstFun, SecondFun}) = Result::EQUAL;
     } else {
+        DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+                        dbgs() << getDebugIndent()
+                               << "Functions are not equal\n");
         ComparedFuns.at({FirstFun, SecondFun}) = Result::NOT_EQUAL;
         while (tryInline.first || tryInline.second) {
+            DEBUG_WITH_TYPE(DEBUG_SIMPLL, increaseDebugIndentLevel());
+
             // Try to inline the problematic function calls
             CallInst *inlineFirst = findCallInst(tryInline.first, FirstFun);
             CallInst *inlineSecond = findCallInst(tryInline.second, SecondFun);
@@ -88,7 +104,7 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                 DEBUG_WITH_TYPE(DEBUG_SIMPLL,
                                 dbgs() << getDebugIndent() << "Try to inline "
                                        << toInline->getName()
-                                       << " in first.\n");
+                                       << " in first\n");
                 if (toInline->isDeclaration()) {
                     DEBUG_WITH_TYPE(DEBUG_SIMPLL,
                                     dbgs() << getDebugIndent()
@@ -112,7 +128,7 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                 DEBUG_WITH_TYPE(DEBUG_SIMPLL,
                                 dbgs() << getDebugIndent() << "Try to inline "
                                        << toInline->getName()
-                                       << " in second.\n");
+                                       << " in second\n");
                 if (toInline->isDeclaration()) {
                     DEBUG_WITH_TYPE(DEBUG_SIMPLL,
                                     dbgs() << getDebugIndent()
@@ -137,8 +153,10 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
             }
             tryInline = {nullptr, nullptr};
             // If nothing was inlined, do not continue
-            if (!inlined)
+            if (!inlined) {
+                DEBUG_WITH_TYPE(DEBUG_SIMPLL, decreaseDebugIndentLevel());
                 break;
+            }
             // Reset the function diff result
             ComparedFuns.at({FirstFun, SecondFun}) = Result::UNKNOWN;
             // Re-run the comparison
@@ -146,13 +164,20 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                                                        controlFlowOnly,
                                                        showAsmDiffs,
                                                        DI, this);
-            if (fCompSecond.compare() == 0) {
+            result = fCompSecond.compare();
+
+            DEBUG_WITH_TYPE(DEBUG_SIMPLL, decreaseDebugIndentLevel());
+            if (result == 0) {
+                DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+                                dbgs() << getDebugIndent() << "After inlining, "
+                                       << "the functions are equal\n");
                 ComparedFuns.at({FirstFun, SecondFun}) = Result::EQUAL;
             } else {
+                DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+                                dbgs() << getDebugIndent() << "After inlining, "
+                                       << "the functions are not equal\n");
                 ComparedFuns.at({FirstFun, SecondFun}) = Result::NOT_EQUAL;
             }
         }
     }
-
-    DEBUG_WITH_TYPE(DEBUG_SIMPLL, decreaseIndentLevel());
 }
--- a/diffkemp/simpll/ModuleComparator.cpp
+++ b/diffkemp/simpll/ModuleComparator.cpp
@@ -26,8 +26,10 @@
 void ModuleComparator::compareFunctions(Function *FirstFun,
                                         Function *SecondFun) {
     DEBUG_WITH_TYPE(DEBUG_SIMPLL,
-                    dbgs() << "Comparing " << FirstFun->getName() << " and "
-                           << SecondFun->getName() << "\n");
+                    dbgs() << getDebugIndent() << "Comparing "
+                           << FirstFun->getName() << " and "
+                           << SecondFun->getName() << "\n";
+                    increaseIndentLevel());
     ComparedFuns.emplace(std::make_pair(FirstFun, SecondFun), Result::UNKNOWN);
 
     // Comparing function declarations (function without bodies).
@@ -54,6 +56,8 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                     this->MissingDefs.push_back({nullptr, SecondFun});
             }
         }
+
+        DEBUG_WITH_TYPE(DEBUG_SIMPLL, decreaseIndentLevel());
         return;
     }
 
@@ -62,7 +66,8 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                                          showAsmDiffs, DI, this);
     if (fComp.compare() == 0) {
         DEBUG_WITH_TYPE(DEBUG_SIMPLL,
-                        dbgs() << "Function " << FirstFun->getName()
+                        dbgs() << getDebugIndent(-1) << "Function "
+                               << FirstFun->getName()
                                << " is same in both modules\n");
         ComparedFuns.at({FirstFun, SecondFun}) = Result::EQUAL;
     } else {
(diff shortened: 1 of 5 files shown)


diffkemp/simpll/SourceCodeUtils.cpp
```diff
@@ -353,20 +357,20 @@ std::vector<SyntaxDifference> findMacroDifferences(
...
             DEBUG_WITH_TYPE(DEBUG_SIMPLL,
-                dbgs() << "Left stack:\n\t";
-                dbgs() << LValue->second.body << "\n";
+                dbgs() << getIndent() << "Left stack:\n\t";
+                dbgs() << getIndent() << LValue->second.body << "\n";
                 for (CallInfo &elem : StackL) {
-                    dbgs() << "\t\tfrom " << elem.fun <<
-                        " in file " <<
+                    dbgs() << getIndent() << "\t\tfrom " << elem.fun
+                        << " in file " <<
```
viktormalik: Make each line start with `<<`, aligned vertically.

diffkemp/simpll/Utils.cpp
```diff
@@ -32,6 +32,9 @@
 #include <llvm/Transforms/Scalar/SimplifyCFG.h>
 #include <llvm/Transforms/Utils/Cloning.h>
 
+/// Level of debug indentation. Each level corresponds to two characters.
+int indentationLevel = 0;
```
viktormalik: Make the variable `static` since it is local to this file.

diffkemp/simpll/SourceCodeUtils.cpp
```diff
@@ -353,20 +357,20 @@ std::vector<SyntaxDifference> findMacroDifferences(
...
             DEBUG_WITH_TYPE(DEBUG_SIMPLL,
-                dbgs() << "Right stack:\n\t";
-                dbgs() << RValue->second.body << "\n";
+                dbgs() << getIndent() << "Right stack:\n\t";
+                dbgs() << getIndent() << RValue->second.body << "\n";
                 for (CallInfo &elem : StackR) {
-                    dbgs() << "\t\tfrom " << elem.fun <<
-                        " in file " <<
+                    dbgs() << getIndent() << "\t\tfrom " << elem.fun
+                        << " in file " <<
```
viktormalik: Make each line start with `<<`, aligned vertically.

diffkemp/simpll/Utils.cpp
viktormalik: Name this variable `debugIndentLevel`.

diffkemp/simpll/Utils.h
```diff
@@ -129,4 +129,12 @@ std::string getIdentifierForValue(const Value *Val,
 /// methods.
 std::string valueToString(const Value *Val);
 
+/// Get a string matching the current indentation level.
+/// \param prefixChar Indentation prefix character, defaults to space.
+std::string getIndent(const char prefixChar = ' ');
```
viktormalik: Name this function `getDebugIndent`.

diffkemp/simpll/Utils.h
```diff
@@ -129,4 +129,12 @@ std::string getIdentifierForValue(const Value *Val,
...
+
+/// Change the indentation level based on offset.
+/// \param offset Indentation level offset.
+void offsetIndentLevel(const int offset);
```
viktormalik: Since this is always used with 1 or -1, I'd suggest replacing it with two functions:
`increaseDebugIndent` and `decreaseDebugIndent`.

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -62,7 +65,8 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                                          showAsmDiffs, &GS, DI, this);
     if (fComp.compare() == 0) {
         DEBUG_WITH_TYPE(DEBUG_SIMPLL,
-                        dbgs() << "Function " << FirstFun->getName()
+                        dbgs() << getIndent() << "Function "
```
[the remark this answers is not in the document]
lenticularis39: I think this should be indented with the same level as the "Comparing..." message (for example by setting the -1 indent before the print), not not indented at all.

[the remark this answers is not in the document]
viktormalik: Yes, that is what I wanted to say.

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -54,6 +55,8 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                     this->MissingDefs.push_back({nullptr, SecondFun});
             }
         }
+
+        DEBUG_WITH_TYPE(DEBUG_SIMPLL, offsetIndentLevel(-1));
```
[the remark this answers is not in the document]
lenticularis39: Also do not forget that there is a case in which `ComparedFuns.at({FirstFun, SecondFun})` is not set at all because of missing definitions.

diffkemp/simpll/ModuleComparator.cpp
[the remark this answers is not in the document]
PSilling: I have updated the `getDebugIndent` function to support indentation level offsetting for this use case.

PSilling:
Updated based on the review. Now also features new debugging prints for all possible function comparison results, not just full differential equality.

diffkemp/simpll/Utils.h
```diff
@@ -129,4 +129,15 @@ std::string getIdentifierForValue(const Value *Val,
 /// methods.
 std::string valueToString(const Value *Val);
 
+/// Get a string matching the current indentation level.
+/// \param offset Indentation level offset to use, defaults to zero.
+/// \param prefixChar Indentation prefix character, defaults to space.
+std::string getDebugIndent(const int offset = 0, const char prefixChar = ' ');
+
+/// Increase the level of debug indentation by one.
+void increaseIndentLevel();
```
viktormalik: Use rather `increaseDebugIndentLevel` or `increaseDebugIndent` to keep it shorter. It should be clear that these functions deal with debugging messages indent.

diffkemp/simpll/Utils.h
```diff
@@ -129,4 +129,15 @@ std::string getIdentifierForValue(const Value *Val,
...
+
+/// Decrease the level of debug indentation by one.
+void decreaseIndentLevel();
```
viktormalik: Same as above.

diffkemp/simpll/ModuleComparator.cpp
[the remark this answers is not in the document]
viktormalik: Is it necessary to change the `getDebugIndent` function? I think it would be better to decrease the indent level before.The same can be done for the message that the functions are not equal (actually, it can be decreased just once, right after running the `DifferentialFunctionComparator`). Then, when there is inlining done, the indent should be increased again.

viktormalik:
Also, please add debug message stating the result of the comparison after the inlining was done (lines 162 and 164 of `ModuleComparator`).

diffkemp/simpll/ModuleComparator.cpp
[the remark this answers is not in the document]
PSilling: This is definitely correct. However, I feel like this use case scenario would occur more times in the future and it's simpler to put the solution to it directly inside the `getDebugIndent` function. I also think that this approach makes the indentation maybe a little less readable though... Should I remove the offset functionality then, or keep it?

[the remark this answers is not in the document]
viktormalik: I agree with the fact that this approach makes the indentation less readable, which I think is the most important argument. Hence, I'd suggest to remove the offset functionality and require the user to explicitly increase or decrease the indentation level.
Looking for opinion from @lenticularis39 .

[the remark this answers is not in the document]
lenticularis39: It seems logical to me to explicitly decrease the indentation level at least in this case, because if I understand it correctly there is no additional decrease, it is only moved before the "Functions are equal" print.

diffkemp/simpll/Utils.h
PSilling: Here I will stick to the longer version, even though I skipped the `Debug` part in the function name to make it shorter. I believe it should be clearer that this function doesn't change the number of indentation characters, but the current indentation nesting level (currently two spaces per level by default).

diffkemp/simpll/ModuleComparator.cpp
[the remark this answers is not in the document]
PSilling: Alright, agreed. Will be removing the offset option.

diffkemp/simpll/Utils.h
lenticularis39: This function name length is actually not that bad considering there are longer function names like `getSubstringToMatchingBracket` or `convertInlineAsmToLLVMFormat` in SimpLL.

PSilling:
Amended the requested changes. Should be ready for review again.

lenticularis39:
Please check whether this command works on your branch:
```
bin/diffkemp -v compare snapshots/linux-3.10.0-693.el7/ snapshots/linux-3.10.0-862.el7/ --show-diff -f _cond_resched
```
For me it throws an error in the indentation code:
```
Functions are equal
terminate called after throwing an instance of 'std::length_error'
  what():  basic_string::_M_create

Program received signal SIGABRT, Aborted.
0x00007ffff383a428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff383a428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff383c02a in __GI_abort () at abort.c:89
#2  0x00007ffff3e80242 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff3e8be86 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff3e8bed1 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff3e8c105 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff3e82982 in std::__throw_length_error(char const*) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007ffff3f24d8c in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_create(unsigned long&, unsigned long) ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00007ffff3f24e1c in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct(unsigned long, char) ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#9  0x0000000000454bca in getDebugIndent[abi:cxx11](char) (prefixChar=32 ' ') at ../diffkemp/simpll/Utils.cpp:622
#10 0x0000000000476d1d in ModuleComparator::compareFunctions (this=0x7fffffffd8f0, FirstFun=0x8ad328, SecondFun=0xe65348) at ../diffkemp/simpll/ModuleComparator.cpp:78
#11 0x000000000047a648 in simplifyModulesDiff (config=..., nonequalFuns=std::vector of length 0, capacity 0, missingDefs=std::vector of length 0, capacity 0, 
    differingMacros=std::vector of length 0, capacity 0) at ../diffkemp/simpll/Transforms.cpp:163
#12 0x000000000040ca3d in main (argc=9, argv=0x7fffffffe658) at ../diffkemp/simpll/SimpLL.cpp:49
```

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -62,10 +76,16 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                                          showAsmDiffs, &GS, DI, this);
     if (fComp.compare() == 0) {
         DEBUG_WITH_TYPE(DEBUG_SIMPLL,
-                        dbgs() << "Function " << FirstFun->getName()
-                               << " is same in both modules\n");
+            decreaseDebugIndentLevel();
```
viktormalik: You are duplicating code here (calling `decreaseDebugIndentLevel` both here and in the `else` branch). I suggest moving the call before the condition (right after the invocation of the function comparator).

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -62,10 +76,16 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
...
+            dbgs() << getDebugIndent() << "Functions are equal\n";
+        );
         ComparedFuns.at({FirstFun, SecondFun}) = Result::EQUAL;
     } else {
+        DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+            decreaseDebugIndentLevel();
+            dbgs() << getDebugIndent() << "Functions are not equal\n";
+            increaseDebugIndentLevel()
```
viktormalik: I don't think that calling `increaseDebugIndentLevel` here is correct. In case the result is "not equal" but there are no functions to inline, the indent level will never be decreased. I suggest moving this call into the following loop.

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -140,8 +162,18 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                                                        showAsmDiffs,
                                                        &GS, DI, this);
             if (fCompSecond.compare() == 0) {
+                DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+                    decreaseDebugIndentLevel();
```
viktormalik: Code duplication - same as above.

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -140,8 +162,18 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
...
+                    dbgs() << getDebugIndent()
+                           << "Inlined functions are equal\n"
```
viktormalik: Better message: "After inlining, the functions are equal".
We're still comparing the same functions, just with some function calls within them inlined. We're not comparing the functions that were inlined.

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -140,8 +162,18 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
...
+                );
                 ComparedFuns.at({FirstFun, SecondFun}) = Result::EQUAL;
             } else {
+                DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+                    decreaseDebugIndentLevel();
+                    dbgs() << getDebugIndent()
+                           << "Inlined functions are not equal\n"
```
viktormalik: See better message above.

diffkemp/simpll/ModuleComparator.cpp
PSilling: Correct, but this was intentional. The problem is that the function comparator needs to be executed before the indentation decrease. So, at least as far as I understand it, this wouldn't be possible unless I also move `fComp.compare()` before the condition and save its result into a new variable, which is in my opinion worse than one extra debug line. The same goes for the other code duplication comments below.

diffkemp/simpll/ModuleComparator.cpp
PSilling: Yes, I've missed this one. I believe this is also the reason @lenticularis39 ended up having an error (`getDebugIndent` in the while loop was executed with a negative number). Maybe I should protect the `debugIndentLevel` variable from being negative and/or make it unsigned so that this cannot result in a crash in the future?

viktormalik: I agree, make it unsigned. However, its value should never underflow, therefore, add an assertion to `decreaseDebugIndentLevel`:

    assert(debugIndentLevel >= 0);

This will help us to find this kind of errors earlier.

diffkemp/simpll/ModuleComparator.cpp
viktormalik: Indeed, I did not realize this. Let's move `fComp.compare()` out of the condition and assign it into some variable `result`. Currently, it's kind of hidden there (I just overlooked it when writing the comment) and it decreases readability.

PSilling:
Should now be ready for review again.

PSilling:
Should now be ready for review again.

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -139,9 +162,18 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
                                                        controlFlowOnly,
                                                        showAsmDiffs,
                                                        DI, this);
-            if (fCompSecond.compare() == 0) {
+            result = fCompSecond.compare();
+
+            DEBUG_WITH_TYPE(DEBUG_SIMPLL, decreaseDebugIndentLevel());
```
viktormalik: There remains one problem, this loop can be terminated by the `break` statement on line 157. This would cause the indent not to be decreased correctly. I'd suggest putting one additional decrease before the break occurs.

PSilling: Right, missed this one after I moved the decrease into the loop. Should be corrected now.

PR #94, opened 2019-08-23
SimpLL: Cache macro map.

lenticularis39:
This saves the macro name to MacroElement map for each module to another
map when it is generated and reuses it the next time SimpLL looks for
macros in the same module.

### DIFF
--- a/diffkemp/simpll/SourceCodeUtils.cpp
+++ b/diffkemp/simpll/SourceCodeUtils.cpp
@@ -214,6 +214,10 @@ std::string extractLineFromLocation(DILocation *LineLoc, int offset) {
 /// map.
 std::unordered_map<std::string, MacroElement> getAllMacrosAtLocation(
     DILocation *LineLoc, const Module *Mod, int lineOffset) {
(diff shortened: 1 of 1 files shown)


diffkemp/simpll/SourceCodeUtils.cpp
```diff
@@ -214,6 +214,10 @@ std::string extractLineFromLocation(DILocation *LineLoc, int offset) {
 /// map.
 std::unordered_map<std::string, MacroElement> getAllMacrosAtLocation(
     DILocation *LineLoc, const Module *Mod, int lineOffset) {
+    // Store generated macro maps for modules to avoid having to regenerate them
+    // many times when comparing a module that has to be inlined a lot.
+    static std::map<const Module *, StringMap<MacroElement>> Cache;
```
viktormalik: Call the variable `MacroCache`.

viktormalik: Even better, `MacroMapCache`.

lenticularis39: Good idea.

PR #95, opened 2019-08-29
SimpLL: Generate field access abstractions. 

lenticularis39:
Add a pass that takes blocks starting with a GEP instruction and
containing GEP and cast instructions (in case they meet certain
conditions, see the comments in the code) and creates a separate
function into which the block is moved. This makes comparing the
block as a whole simple, while also retaining the possibility to
compare it instruction by instruction by inlining.

The pass also adds one of this whole-unit field access comparisons -
the comparison of such blocks by the total constant offset, which
solves a common false positive case where a field is moved into an
anonymous structure, union or both.


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -693,6 +699,46 @@ int DifferentialFunctionComparator::cmpGlobalValues(GlobalValue *L,
...
+/// Specific comparing of structure field access.
+int DifferentialFunctionComparator::cmpFieldAccess(const Function *L,
+                                                   const Function *R) const {
+    // First compute the complete offset of all GEPs in both functions.
+    // It is possible that although the process contains unpacking several
+    // anonymous unions and structs, the offset stays the same, which means that
+    // without a doubt this is semantically equal (and probably would, in fact,
+    // be also equal in machine code).
+    uint64_t OffsetL = 0, OffsetR = 0;
+    for (auto &Inst : *(L->begin())) {
```
viktormalik: This loop is exactly the same as the following one. Can we make it a function (something like `accumulateAllOffsets`)?

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+    return PreservedAnalyses();
+}
+
+/// First checks whether the instructions in the stack can be moved away from
+/// the original function without causing an error, then creates the abstraction
+/// function, moves the instructions into it, modify the input and output of the
+/// block to match the argument and returns value of the function and inserts
+/// the call to the newly create abstraction instead of the instruction block to
+/// the original function.
+void FieldAccessFunctionGenerator::processStack(
```
viktormalik: I'd suggest a slightly different implementation:
- make `Stack` constant: this function should not modify it except for clearing it at the end - this can be done in the caller function in a better place (see other comment above)
- since you don't have to clear it here, you can get rid of `ValidUserSet`, `allUsersValid`, and `allOperandsValid` variables (more details in the comments below).

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+    // If not, do not generate an abstraction, since this would break the code.
+    std::set<Value *> ValidUserSet;
+    for (Instruction *Inst : Stack)
+        ValidUserSet.insert(Inst);
+    bool allUsersValid = true;
+    for (Instruction *Inst : Stack) {
+        if (Inst == Stack.back())
+            continue;
+        for (User *U : Inst->users()) {
+            if (ValidUserSet.find(U) == ValidUserSet.end()) {
```
viktormalik: Search in `Stack` (after making it constant) and directly return if user is not there.

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+void FieldAccessFunctionGenerator::processStack(
+        std::vector<Instruction *> &Stack, Module &Mod) {
+    if (Stack.empty())
+        // Do not attempt to process empty stacks, that would cause an error.
+        return;
+
+    // Check if all the instruction (besides the last one, of course) are used
+    // only by other instructions in the stack.
+    // If not, do not generate an abstraction, since this would break the code.
+    std::set<Value *> ValidUserSet;
```
viktormalik: Remove.

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+        if (Inst == Stack.back())
+            continue;
+        for (User *U : Inst->users()) {
+            if (ValidUserSet.find(U) == ValidUserSet.end()) {
+                allUsersValid = false;
+                break;
+            }
+        }
+    }
+    if (!allUsersValid) {
```
viktormalik: Remove.

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+        if (Inst == Stack.front()) {
+            // The GEP is guaranteed to have all indices constant and the first
+            // value is extracted.
+            continue;
+        }
+        for (Value *Op : Inst->operands()) {
+            if (isa<ConstantData>(Op))
+                // Constants are fine, they can be moved without problem.
+                continue;
+            if (ValidUserSet.find(Op) != ValidUserSet.end())
```
viktormalik: Search in `Stack` (after making it constant) and directly return if the operand is not found.

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+                // The value was found in the stack set.
+                continue;
+            allOperandsValid = false;
+            break;
+        }
+    }
+    if (!allOperandsValid) {
```
viktormalik: Remove.

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+        for (auto &BB : Fun) {
+            for (auto &Inst : BB) {
+                if (isa<GetElementPtrInst>(Inst) && InstStack.empty()) {
+                    // Possible start of field access block.
+                    // Check if GEP has constant indices (this pass could
+                    // technically support all GEPs, but this is not implemented
+                    // yet)
+                    if (dyn_cast<GetElementPtrInst>(&Inst)->
+                            hasAllConstantIndices()) {
+                        DIL = Inst.getDebugLoc().get();
```
viktormalik: Since this is the only possible start of a field access block, you can clear `Stack` here.

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+            // be replaced with the function call in the original function.
+            Instruction *Clone = Inst->clone();
+            Clone->setDebugLoc(DebugLoc());
+            auto CallInst = CallInst::Create(FT, Abstraction,
+                    {Stack.front()->getOperand(0)});
+            CallInst->insertAfter(Inst);
+            Inst->replaceAllUsesWith(CallInst);
+            Inst->removeFromParent();
+            BB->getInstList().push_back(Clone);
+            Stack[Stack.size() - 1] = Clone;
```
viktormalik: Instead of modifying the last instruction in `Stack`, modify the last instruction in `BB`.

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+        } else {
+            Inst->removeFromParent();
+            Inst->setDebugLoc(DebugLoc());
+            BB->getInstList().push_back(Inst);
+        }
+    }
+    Stack.front()->setOperand(0, Arg);
```
viktormalik: Instead of modifying the first instruction in `Stack`, modify the first instruction in `BB`.

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
lenticularis39: I don't think I can do that, because the algorithm uses the emptiness of the stack as an indication whether it is inside or outside a block.

viktormalik: Right, then you can clear it after calling `processStack`. Or you can create a wrapper function that would call `processStack` and then it would clean the stack.

viktormalik: Maybe, the best would be to name the wrapper function `processStack` and rename the current `processStack` to something like `createFunctionFromStack`.

lenticularis39: I think that creating a wrapper function just for that is not necessary, there are not that many places where the stack is processed, clearing it explicitly after each call should be fine.

viktormalik: Ok, however, try to think if it is possible that we will do something more with the stack in future that could extend this wrapper function. If no, then stick with clearing the stack after each call. If it is possible that we do something more with the stack or we call `processStack` from other locations, then create a wrapper function.

diffkemp/simpll/passes/FieldAccessFunctionGenerator.cpp
```diff
@@ -0,0 +1,175 @@
...
+        if (Inst == Stack.front()) {
+            // The GEP is guaranteed to have all indices constant and the first
+            // value is extracted.
+            continue;
+        }
+        for (Value *Op : Inst->operands()) {
+            if (isa<ConstantData>(Op))
+                // Constants are fine, they can be moved without problem.
+                continue;
+            if (std::find(Stack.begin(), Stack.end(), Op) != std::end(Stack))
```
viktormalik: Use the same style as in the above loop:

    if (std::find(...) == std::end(...))
        return;

It is confusing when you use different styles for the same functionality.

PR #97, opened 2019-08-30
SimpLL: Remove function abstraction hashes.

lenticularis39:
The comparison of the functions is handled by cmpBasicBlocks and
cmpOperations (the comparing of the function type and the argument
types) and by cmpGlobalValues in the case of inline assembly, which
is compared by inline assembly code strings from the map.

Note: The comparison by hashes was created when the comparison of global values was still handled by DifferentialGlobalNumberState and has the problem that the hash changes when the type name changes, even in cases when the result of cmpTypes is.


lenticularis39:
It seems like this causes some problems as of now.

PR #99, opened 2019-09-02
SimpLL: Remove function abstraction hashes.

lenticularis39:
The comparison of the abstractions is handled by cmpBasicBlocks and
cmpOperations (the comparing of the function type and the argument
types) and by cmpGlobalValues in the case of inline assembly, which
is compared by inline assembly code strings from the map.

Note: Function abstraction unification is removed, since it is no longer necessary. This enables better abstraction comparison, for example when comparing the control flow only, integer types are compared as the same, but that wouldn't work with the hashes, because the type change would change the hash. Also abstraction suffixes are changed to use the dot notation to ignore the numbers in cmpGlobalValues.


lenticularis39:
I think in that case the functions would have different names, so they wouldn't be compared as equal.

viktormalik:
Ok, that sounds correct.

PR #105, opened 2019-09-05
Improve comparing attributes and integer types

viktormalik:
- ignore differences in attributes (always compare attribute lists as equal)
- distinguish boolean from other integer types even when comparing the control flow only

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -134,32 +134,10 @@ int DifferentialFunctionComparator::cmpGEPs(
     return 0;
 }
 
-
-/// Remove chosen attributes from the attribute set at the given index of
-/// the given attribute list. Since attribute lists are immutable, they must be
-/// copied all over.
-AttributeList cleanAttributes(AttributeList AS, unsigned Idx, LLVMContext &C) {
-    AttributeList result = AS;
-    result = result.removeAttribute(C, Idx, Attribute::AttrKind::AlwaysInline);
-    result = result.removeAttribute(C, Idx, Attribute::AttrKind::InlineHint);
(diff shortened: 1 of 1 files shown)


lenticularis39:
Is there any particular reason why boolean is treated differently than the other integer types? It seems to me like the impact that the treating of it as the other integer types has on the comparison is at least the same, if not smaller than with the other integer types (usually there would only be logical operations and comparison with the booleans, which work exactly the same as with other integers; in case there are other operations, they will be only in the version with the integer type, therefore there'll be a difference anyway).

viktormalik:
Yes, the thing is that treating all integers as the same is accompanied by ignoring bitcast operations, which is essential for successful comparison (because one version may contain additional bitcasts due to the changed type). However, an integer can be cast to bool by `icmp ne` instruction, which should be ignored as well in such case. However, this instruction is not only used for casting and therefore cannot be ignored.

lenticularis39:
Ah, I see. Is there a case where this causes a false positive/negative? I think that in case there isn't, it may be better to leave the condition as it is and file an enhancement issue on the `icmp ne` 
"cast" - a `ne` comparision against `0` alone shouldn't cause a semantic difference for the reasons described in the previous comment (and this could be a nice first issue if that it true).

viktormalik:
Yes, it is actually in the regression tests (`dev_trans_start` in `rhel-74-75`). It calls a function `is_vlan_dev` whose return type changed from int to bool. This is solved by inlining the function, however, with `--control-flow-only`, it was inlined by "accident" due to different function attributes. When ignoring the attribute differences, the calls are treated as equal, the function is not inlined, and the comparison fails at an extra `icmp ne` instruction at the end of the function. 
This could be solved by ignoring `icmp ne` with a `0` as an operand, however, I think that it is not a sound approach. This is because `icmp ne` can be generated from an `if` condition and in such case, we don't want to ignore it (with `--control-flow-only`) since it has a direct impact on the control flow.

PR #110, opened 2019-09-18
SimpLL: Fix inlining causing empty diffs.

lenticularis39:
The problem was that there were some cases when the call instructions
differed in a way that didn't show up in the C source code (e.g. the
return type changed without affecting the call), leading to inlining,
but the original difference was inside the inlined function (and after
the inlining it was reported in the parent one).

The fix uses a similar technique to syntax difference handling - in case
two functions with the same name are being inlined, they are compared
using ModuleComparator and in case both the original and the inlined
functions are compared as non-equal, the difference between the original
ones is marked as covered (in case the original functions are equal after
the inlining, the difference of the inlined function is removed.

Note: the variables `missingDefs`, `nonequalFuns` and `differingSynDiffs`, which are present at several locations in DiffKemp (in SimpLL.cpp, arguments of the functions `simplifyModulesDiff` and `reportOutput`) and share the fact that they are all main and additonal information generated by ModuleComparator (in which they have the same type, only slightly different names), are merged into a structure type called `ComparisonResult`, including the newly introduced set `coveredFuns`.
Also following this change `coveredBySyntaxDiff` was renamed to `covered` at all places, since it now represents two different types of diff covering.

### DIFF
--- a/diffkemp/simpll/ModuleComparator.cpp
+++ b/diffkemp/simpll/ModuleComparator.cpp
@@ -95,6 +95,24 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
 
             ConstFunPair missingDefs;
             bool inlined = false;
+            // If inlining two functions with the same name, it is possible that
+            // there is a difference in their bodies and only a formal
+            // difference that won't show up in the syntax diff.
+            // For this reason, the functions are compared using
+            // ModuleComparator and the original functions are marked as
+            // covered, leading to the diff not being show when empty.
+            Function *InlinedFunFirst = !inlineFirst ? nullptr :
+                    getCalledFunction(inlineFirst->getCalledValue());
+            Function *InlinedFunSecond = !inlineSecond ? nullptr :
+                    getCalledFunction(inlineSecond->getCalledValue());
+            if (InlinedFunFirst && InlinedFunSecond &&
+                !InlinedFunFirst->getName().startswith("simpll__") &&
+                (InlinedFunFirst->getName() == InlinedFunSecond->getName())) {
+                compareFunctions(InlinedFunFirst, InlinedFunSecond);
+                if (ComparedFuns.at({InlinedFunFirst, InlinedFunSecond}) ==
+                        Result::NOT_EQUAL)
+                    CoveredFuns.insert(FirstFun->getName().str());
+            }
             // If the called function is a declaration, add it to missingDefs.
             // Otherwise, inline the call and simplify the function.
             // The above is done for the first and the second call to inline.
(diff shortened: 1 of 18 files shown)


diffkemp/simpll/Utils.cpp
```diff
@@ -49,6 +49,18 @@ const Function *getCalledFunction(const Value *CalledValue) {
     return fun;
 }
 
+Function *getCalledFunction(Value *CalledValue) {
```
viktormalik: Add commentary for the function.

diffkemp/simpll/Utils.h
```diff
@@ -46,6 +46,7 @@ typedef std::vector<CallInfo> CallStack;
 
 /// Extract called function from a called value
 const Function *getCalledFunction(const Value *CalledValue);
+Function *getCalledFunction(Value *CalledValue);
```
viktormalik: Add separate comment for the function.

diffkemp/simpll/Utils.cpp
viktormalik: This can be implemented by (no code duplication):

    return const_cast<Function *>(getCalledFunction(const_cast<const Value *>(CalledValue)))

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -95,6 +95,24 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
...
+            // difference that won't show up in the syntax diff.
+            // For this reason, the functions are compared using
+            // ModuleComparator and the original functions are marked as
+            // covered, leading to the diff not being show when empty.
+            Function *InlinedFunFirst = !inlineFirst ? nullptr :
+                    getCalledFunction(inlineFirst->getCalledValue());
+            Function *InlinedFunSecond = !inlineSecond ? nullptr :
+                    getCalledFunction(inlineSecond->getCalledValue());
+            if (InlinedFunFirst && InlinedFunSecond &&
+                !InlinedFunFirst->getName().startswith("simpll__") &&
```
viktormalik: This can replaced by `isSimpllAbstractionDeclaration`.

lenticularis39: That wouldn't work, because I also have to exclude field access abstractions - `isSimpllAbstractionDeclaration` checks only for abstraction declarations (i.e. those generated by `FunctionAbstractionGenerator`).

viktormalik: Then let's write a function for checking field access abstractions?

PR #113, opened 2019-10-10
Update README with info about RPM

viktormalik:
Also move the installation section to the top.


README.md
```diff
@@ -3,7 +3,44 @@
...
+* Python packages from `requirements.txt` (run `pip install -r requirements.txt`)
+* CScope
+
+Additionally, to build manually, you need to install the following tools:
+* CMake
+* Ninja build system
+
+Build can be done by running:
+
+	mkdir build
```
lenticularis39: This looks like mixed spaces and tabs.

README.md
```diff
@@ -3,7 +3,44 @@
...
 
+## Installation
+
+There are two options to install the project, either build from source or use a
+prepared RPM package for Fedora.
+
+### Install from source
+
+Currently, DiffKemp runs on Linux and needs the following software installed:
+* LLVM (supported versions are 5, 6, 7 and 8)
```
lenticularis39: Maybe it would be helpful for the user to mention that DiffKemp also required Clang, since Clang is usually not included in LLVM packages, although it is build together with LLVM.

PR #114, opened 2019-10-12
Travis CI: Add and apply clang-format rules to C++ files

### LINKED ISSUE #69: Automatic code-style checking for C++ code
Add code-style checking per LLVM style guide into Travis CI.

PSilling:
**Overview**
Adds `clang-format` rules to enable automatic C++ code formatting and style checking. The rules are also applied to all existing SimpLL files which are now also automatically style checked during CI by the included bash script. For development, `clang-format-8` and new symbolic links have been added to the `diffkemp-devel` docker container for easier style checking.

Closes #69 

**Notes**
- `.clang-format` has been placed to the root directory as `clang-format` does not allow specifying a direct path for the settings file. It can only search for a `.clang-format` file in parent directories (non-recursively, however).
- Even though both `clang` and `llvm` have varying versions included for CI, as `clang-format` is for development purposes only, only version 8 of `clang-format` is included. Another reason for this is that the `.clang-format` settings file is not fully compatible between the versions.
- A normal symbolic link has been used for `check-clang-format.sh` in the Dockerfile as the Diffkemp directory structure is not visible during its setup.
- I was not sure how much information should `check-clang-format.sh` print to `stdout`, so don't forget to look inside the Travis CI build logs (if a difference in style is found, the output there will also feature a `diff` and a list of all incorrectly formatted files at the end). The other question here is whether to leave this as a script inside the `tools` directory or move it to a new repository. I believe it should stay in `tools`, but it is something to consider as the script is not necessarily tied to Diffkemp.
- The rules inside `.clang-format` are mostly based on the LLVM standard, however, the LLVM standard is pretty liberal when it comes to changes, so feel free to suggest anything. On the other hand, `clang-format` is not very customizable, so the proposed changes may be impossible to implement or may have other consequences.


PSilling:
Yes, that should work. May change the version after #115 gets merged I suppose.

lenticularis39:
I think the version of clang-format does not matter that much here, since it doesn't have to match the version of LLVM used for DiffKemp. It may actually be better to keep the version at 8, because for someone who uses DiffKemp outside a container 9 may not be available (for example on Fedora 30).

diffkemp/simpll/DebugInfo.cpp
```diff
@@ -412,29 +410,31 @@ void DebugInfo::collectMacrosWithValue(const Constant *Val) {
...
+                auto ValMD =
+                        dyn_cast<MetadataAsValue>(WrappedVal)->getMetadata();
                 auto Val = dyn_cast<ValueAsMetadata>(ValMD);
                 if (!Val)
                     continue;
 #if LLVM_VERSION_MAJOR < 6
                 MetadataAsValue *DIVal;
                 if (getCalledFunction(CInst->getCalledValue())->getName() ==
-                        "llvm.dbg.declare")
+                    "llvm.dbg.declare")
```
viktormalik: I don't like the fact that this is aligned to the parenthesis since the last line of the condition has the same indent as the first command of the block and it is not clear where the split is. However, as far as I remember, this cannot be configured using `clang-format`, right?

diffkemp/simpll/DebugInfo.h
```diff
@@ -65,7 +67,7 @@ class DebugInfo {
 
     /// Maps local variable names to their values.
     std::unordered_map<std::string, const Value *> LocalVariableMapL,
-                                                   LocalVariableMapR;
+            LocalVariableMapR;
```
viktormalik: This is not very nice since this alignment may lead to missing the fact that there are 2 variables defined here. It would suffice to put them at the same line, but I guess that `clang-format` won't allow that either without breaking other stuff.

viktormalik:
I agree with @lenticularis39, let's keep version 8 as default (we're doing the same in the container).

viktormalik:
I merged the LLVM 9 PR, there are now merge conflicts, please rebase.

viktormalik:
> * Even though both `clang` and `llvm` have varying versions included for CI, as `clang-format` is for development purposes only, only version 8 of `clang-format` is included. Another reason for this is that the `.clang-format` settings file is not fully compatible between the versions.

Is it at least compatible among some versions? Which ones?

diffkemp/simpll/DebugInfo.h
PSilling: I agree, however, `clang-format` only supports indentation after an open `(` bracket and for semicolon-separated declarations and assignments. The reason for the lack of support here is, at least based on information I have found, to discourage the use of comma-separated variable declarations altogether. Therefore, I will change the declaration to semicolon-separated one in such cases for better readability.

diffkemp/simpll/DebugInfo.cpp
PSilling: Technically, this is possible, but I do not recommend such an approach. It requires the disabling of this alignment after all `(` braces. Because of that even functions with a lot of parameters would end up looking like this (taken directly from `clang-format-8` documentation):
```
someLongFunction(argument1,
    argument2);
```
Instead, I suggest to break before the binary operands, so the break is more clearly visible as the line will begin with `&&`, `||` or `==`. I will implement this approach including what I have commented below so that the changes can be reviewed.

PSilling:
> > * Even though both `clang` and `llvm` have varying versions included for CI, as `clang-format` is for development purposes only, only version 8 of `clang-format` is included. Another reason for this is that the `.clang-format` settings file is not fully compatible between the versions.
> 
> Is it at least compatible among some versions? Which ones?

Yes, some versions are able to process a `.clang-format` style file generated by a different version. To be specific, if you use a version 5.0 generated `.clang-format`, you can process it using versions 5.0, 6.0, 7 or even 8. You can also process a version 7 generated `.clang-format`, but only with versions 7 or 8. In other scenarios, however, there is no compatibility. I can't even guarantee that the limited backward compatibility listed above will always generate the same result, although it at least seems to do so.

> I merged the LLVM 9 PR, there are now merge conflicts, please rebase.
I pushed from a different local branch to preview the proposed changes (and keep the old ons locally). Unfortunately, I did not check the commit history which I somehow squashed during a rebase I suppose. Please do not merge this, I will push again from the previous branch after I recieve feedback on the changes.

diffkemp/simpll/DebugInfo.cpp
viktormalik: This is a good solution, I like the way it looks in the code.

diffkemp/simpll/DebugInfo.h
viktormalik: Good, this should be the recommended approach from now in case of multiline variable definitions.

PR #116, opened 2019-10-23
Better cast handling

lenticularis39:
Casts that do not cause any semantic difference are now always skipped, not only when comparing the control flow only.

Furthermore cmpBasicBlocks was refactored in a way so it doesn't include additional comparisons of call instructions using C source code (they were moved to a separate function). This reduces the length of cmpBasicBlocks and makes the code more comprehensible.


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -349,9 +349,58 @@ int DifferentialFunctionComparator::cmpAllocs(const CallInst *CL,
...
+                // Integer-to-integer casts are safe to ignore only if two
+                // conditions are met:
+                // 1. The destination type is larger than the source type.
+                // 2. No arithmetic is performed on the number.
+                // Note: store instructions aren't a problem here, since the
+                // types of the arithmetic instructions would be different if
+                // there is arithmetic performed on the value after being
+                // loaded back.
+                auto IntTyL = dyn_cast<IntegerType>(Cast->getSrcTy());
+                auto IntTyR = dyn_cast<IntegerType>(Cast->getDestTy());
```
viktormalik: I'd rather call these variables `IntTySrc` and `IntTyDest` because we use `L` and `R` suffices when comparing things between modules, which is not the case here.

PR #139, opened 2020-01-17
SimpLL: fix bug in DifferentialFunctionComparator

viktormalik:
Due to previous commit, it is possible that a call instruction type is NULL while the called function can be found. Use type of the called function instead of the instruction type when needed.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -198,6 +198,8 @@ int DifferentialFunctionComparator::cmpOperations(
             if (CalledL && CalledR) {
                 if (isSimpllFieldAccessAbstraction(CalledL)
                     && isSimpllFieldAccessAbstraction(CalledR)
+                    && isa<PointerType>(CL->getType())
+                    && isa<PointerType>(CR->getType())
                     && (dyn_cast<PointerType>(CL->getType())->getElementType()
                         != dyn_cast<PointerType>(CR->getType())
                                    ->getElementType())) {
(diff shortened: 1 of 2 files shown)


viktormalik:
@lenticularis39 please check also the previous commit, I already merged it into the master but obviously it can cause problems that I overlooked (as the one solved by this PR).

lenticularis39:
Do you have an example where the type of the call instruction is null? Such a case seems unusual to me. 

Also the type returned by `CL->getType()` and `CalledL->getType()` is different, see for example this call:
```
(gdb) p valueToString(L).c_str()
$2 = 0xecad50 "  %31 = call i64 @simpll__fieldaccess.19(%struct.ahash_request_priv* %17), !dbg !21768"
(gdb) p typeToString(CL->getType()).c_str()
$9 = 0x7fffffff88b0 "i64"
(gdb) p typeToString(CalledL->getType()).c_str()
$10 = 0x905030 "i64 (%struct.ahash_request_priv*)*"
```
The first returns the return value type, the second the type of a pointer to the function type; pointers to function types of two different functions will probably always be different (unlike pointers to return value types), which doesn't lead to the desired behavior (the if should detect field access abstractions returning values of different types, but this way the comparison is unnecessarily done for all of them).

Note: 
I know of a different bug causing a null reference at the same location, but it is caused by the return type of the field access abstraction not being a pointer type, the null coming from the dynamic cast. This bug is also seemingly fixed by this PR; the reason why is described above.

viktormalik:
Indeed, you are right, the bug that I encountered is the one that you describe - field access abstraction returning a non-pointer type. It should be fixed now and I added a test for it.

PR #141, opened 2020-02-03
Tests for correctness of syntax diffs

viktormalik:
Add support for regression tests for checking correctness of syntax
diffs. This is implemented as an extension to the existing regression
test infrastructure.
The individual tests can be specified using the "syntax_diff" key in the
appropriate YAML spec file (see existing tests for inspiration).
Currently supports only checking of differences in macros and inline
assembly code.
Includes also few initial tests, tips for additional tests are welcome.

### DIFF
--- a/tests/regression/syntax_diff_test.py
+++ b/tests/regression/syntax_diff_test.py
@@ -1,0 +1,100 @@
+"""
+Regression test using pytest.
+Tests are specified using YAML files in the tests/test_specs directory.
+This module runs tests for syntax differences using the "syntax_diffs" key in
+the YAML spec file.
+"""
+from diffkemp.semdiff.function_diff import functions_diff
+from .task_spec import SyntaxDiffSpec, specs_path, tasks_path
+import glob
+import os
+import pytest
+import re
+import yaml
+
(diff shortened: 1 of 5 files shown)


viktormalik:
> `__alloc_pages_nodemask` between RHEL 7.3 and 7.4 would be a good candidate for regression testing (that's one of the cases with the macro uses composed with `##` which didn't work for a long time and because of which assembly code diffs were implemented).

Is this the asm difference in `percpu_from_op`? Or the macro diff in `this_cpu_read_X`?

lenticularis39:
All macro differences that have `__pcpu_size_call_return` and `this_cpu_add_X` (or a similar macro ending with a number) in the macro stack, including `percpu_from_op` and `this_cpu_read_X`, are cases when the `##` joining happens (the joining itself is implemented in `__pcpu_size_call_return`).

(Actual assembly differences, i.e. those enabled by `--print-asm-diffs` and generated in `findAsmDifference`, are unrelated to this.)

viktormalik:
Ok, I'll choose a diff of macros that have a short body.

PR #142, opened 2020-02-05
Reuse SimpLL results

### LINKED ISSUE #136: Assuming the equality of function declarations leads to false negatives
Currently when two function definitions with the same name are compared, they are assumed to be equal. This assumption is wrong in many cases, leading to a significant number of false negatives.

For example between versions `3.10.0-862.el7` and `3.10.0-957.el7`, comparison without result caching leads to this result:
```
Statistics
----------
Total symbols: 739
Equal:         601 (81%)
Not equal:     130 (18%)
(empty diff):  3 (0%)
Unknown:       8 (1%)
Errors:        0 (0%)
```
while comparison with result caching (that enables the use of the comparison result of the functions where the body is missing from a previous comparison) is this:
```
Statistics
----------
Total symbols: 739
Equal:         427 (58%)
Not equal:     305 (41%)
(empty diff):  4 (1%)
Unknown:       7 (1%)
Errors:        0 (0%)
```
Clearly many (non-equal) sub-results were assigned to otherwise equal function pairs from the cache (and these diffs are actually true, but even here we don't have all of such diffs).

Another example (this time of a concrete function and KABI symbol on which this happens) is `fc_remove_host` between kernel versions `4.18.0-80.el8` and `4.18.0-147.el8`: when compared separately the symbols are considered to be semantically equal by DiffKemp, but with a cache generated from the comparison of other KABI symbols this macro difference (which appears very often in diffs between these kernel versions) appears:
```
fc_remove_host:
  percpu_from_op differs:
  {{{
    Callstack (snapshots/linux-4.18.0-80.el8/):
    fc_queue_work at drivers/scsi/scsi_transport_fc.c:2452
    dump_stack at drivers/scsi/scsi_transport_fc.c:2338
    __dump_stack at lib/dump_stack.c:113
    dump_stack_print_info at lib/dump_stack.c:76
    raw_smp_processor_id (macro) at lib/dump_stack.c:48
    this_cpu_read (macro) at ./arch/x86/include/asm/smp.h:159
    __pcpu_size_call_return (macro) at ./include/linux/percpu-defs.h:509
    this_cpu_read_8 (macro) at ./include/linux/percpu-defs.h:318
    percpu_from_op (macro) at ./arch/x86/include/asm/percpu.h:477

    Callstack (snapshots/linux-4.18.0-147.el8/):
    fc_queue_work at drivers/scsi/scsi_transport_fc.c:2455
    dump_stack at drivers/scsi/scsi_transport_fc.c:2341
    __dump_stack at lib/dump_stack.c:113
    dump_stack_print_info at lib/dump_stack.c:76
    raw_smp_processor_id (macro) at lib/dump_stack.c:48
    this_cpu_read (macro) at ./arch/x86/include/asm/smp.h:165
    __pcpu_size_call_return (macro) at ./include/linux/percpu-defs.h:509
    this_cpu_read_8 (macro) at ./include/linux/percpu-defs.h:318
    percpu_from_op (macro) at ./arch/x86/include/asm/percpu.h:477

    Diff:
      ({ typeof(var) pfo_ret__; switch (sizeof(var)) { case 1: asm(op "b "__percpu_arg(1)",%0" : "=q" (pfo_ret__) : "m" (var)); break; case 2: asm(op "w "__percpu_arg(1)",%0" : "=r" (pfo_ret__) : "m" (var)); break; case 4: asm(op "l "__percpu_arg(1)",%0" : "=r" (pfo_ret__) : "m" (var)); break; case 8: asm(op "q "__percpu_arg(1)",%0" : "=r" (pfo_ret__) : "m" (var)); break; default: __bad_percpu_size(); } pfo_ret__; })
    
      ({ typeof(var) pfo_ret__; switch (sizeof(var)) { case 1: asm volatile(op "b "__percpu_arg(1)",%0" : "=q" (pfo_ret__) : "m" (var)); break; case 2: asm volatile(op "w "__percpu_arg(1)",%0" : "=r" (pfo_ret__) : "m" (var)); break; case 4: asm volatile(op "l "__percpu_arg(1)",%0" : "=r" (pfo_ret__) : "m" (var)); break; case 8: asm volatile(op "q "__percpu_arg(1)",%0" : "=r" (pfo_ret__) : "m" (var)); break; default: __bad_percpu_size(); } pfo_ret__; })
  }}}
```
(from `dump_stack` down the call stack the functions are in a different module - they got into the cache from one of the previous comparisons when this specific module was involved)

lenticularis39:
Instead of returning a list of non-equal functions with call stacks and a syntax difference list in SimpLL, return a list of all compared functions together with all functions called by them (with attached information about the call site) and attached syntax differences.

This makes it possible to build a graph based on the SimpLL output with vertices being function comparison results and edges representing the calls. Call stacks are then generated by breadth-first search and comparison results by depth-first search.

The representation of SimpLL results by a graph enables keeping one copy of the graph for the entire DiffKemp run, merging the subsequent SimpLL results into it. A list of functions already present in the graph (excluding non-definitive results) is built sequentially and written in a file that is passed to SimpLL, avoiding the comparison of already compared functions, therefore acting like a cache.

Note 1: Call stack depth boundaries are implemented in the depth-first search (method `ComparisonGraph.reachable_from`); functions defined in a C source file different than the one in which the main function resides are excluded from the result. This aligns with the equal assumption for function definitions with the same name in SimpLL. Thus this closes #136.

Note 2: A graph transformation of the original one (representing the SimpLL output) called normalization is implemented to properly handle void-returning variants of functions; this mechanism includes the notion of a "weak edge" - see the comments in `diffkemp/semdiff/caching.py`.

Note 3: the PR also contains small improvements at other places in the code.

Also maybe some squashing of commits would clean up the PR.


diffkemp/semdiff/function_diff.py
```diff
@@ -124,9 +126,117 @@ def functions_semdiff(first, second, fun_first, fun_second, config):
...
+                else fun_second
+            if fun == vertex.name[side]:
+                # There is no callstack from the base function.
+                calls = None
+            else:
+                # Transform the Edge objects returned by
+                # get_shortest_path to a readable callstack.
+                try:
+                    edges = graph.get_callstack(
+                        side, fun, vertex.name[side])
```
viktormalik: I don't like the fact that we first compute the set of all reachable vertices using DFS and then we run another graph BFS traversal for each node in the computed set. I believe that `graph.reachable_from` effectively produces a tree, with the source function at its root. Therefore, if we stored pointers to "parent" or say "caller" functions, it should be easy to reconstruct the callstack in a much shorter time.

diffkemp/semdiff/function_diff.py
```diff
@@ -124,9 +126,117 @@ def functions_semdiff(first, second, fun_first, fun_second, config):
...
+    vertices_to_compare = graph.reachable_from(
+        ComparisonGraph.Side.LEFT, fun_first)
+    # Use methods from ComparisonGraph (on the graph variable) and
+    # vertices_to_compare to generate objects_to_compare.
+    objects_to_compare = []
+    syndiff_bodies_left = dict()
+    syndiff_bodies_right = dict()
+    for vertex in vertices_to_compare:
+        if vertex.result in [Result.Kind.EQUAL,
+                             Result.Kind.ASSUMED_EQUAL]:
```
viktormalik: Maybe replace by `vertex.has_equal_result()` (and get rid of the comment).

diffkemp/semdiff/function_diff.py
```diff
@@ -124,9 +126,117 @@ def functions_semdiff(first, second, fun_first, fun_second, config):
...
+            # Do not include equal functions into the result.
+            continue
+        # Generate and add the function difference.
+        fun_pair = []
+        for side in [ComparisonGraph.Side.LEFT,
+                     ComparisonGraph.Side.RIGHT]:
```
viktormalik: This appears multiple times, could be replaced by `for side in ComparisonGraph.sides()`

diffkemp/semdiff/function_diff.py
```diff
@@ -124,9 +126,117 @@ def functions_semdiff(first, second, fun_first, fun_second, config):
...
+                # get_shortest_path to a readable callstack.
+                try:
+                    edges = graph.get_callstack(
+                        side, fun, vertex.name[side])
+                except ValueError:
+                    raise SimpLLException()
+                calls = _edge_callstack_to_string(edges)
+            # Note: a function diff is covered (i.e. hidden when empty
+            # if and only if there is a non-function difference
+            # referencing it).
```
viktormalik: I think that the closing parenthesis in this comment should be at the end of its first line.

diffkemp/semdiff/function_diff.py
```diff
@@ -124,9 +126,117 @@ def functions_semdiff(first, second, fun_first, fun_second, config):
...
+            fun_pair.append(Result.Entity(
+                vertex.name[side],
+                vertex.file[side],
+                vertex.line[side],
+                calls,
+                "function",
+                len(vertex.nonfun_diffs) != 0
```
viktormalik: Add explicit name of the last parameter (`is_covered`) to make relation to the above comment.

diffkemp/semdiff/function_diff.py
```diff
@@ -124,9 +126,117 @@ def functions_semdiff(first, second, fun_first, fun_second, config):
...
+            ))
+        fun_pair.append(vertex.result)
+        objects_to_compare.append(tuple(fun_pair))
+
+        # Process non-function differences.
+        for nonfun_diff in vertex.nonfun_diffs:
+            nonfun_pair = []
+            for side in [ComparisonGraph.Side.LEFT,
+                         ComparisonGraph.Side.RIGHT]:
```
viktormalik: See above.

diffkemp/semdiff/function_diff.py
```diff
@@ -124,9 +126,117 @@ def functions_semdiff(first, second, fun_first, fun_second, config):
...
+                syndiff_bodies = (syndiff_bodies_left
+                                  if side == ComparisonGraph.Side.LEFT
+                                  else syndiff_bodies_right)
+                # Convert the YAML callstack format to string.
+                calls = ["{} at {}:{}".format(call["function"],
+                                              call["file"],
+                                              call["line"])
+                         for call in nonfun_diff.callstack[side]]
+                # Join the elements in the list to get a string.
+                calls = "\n".join(calls)
```
viktormalik: This reminds body of the `_edge_callstack_to_string` function. Could it be somehow merged?

diffkemp/semdiff/caching.py
```diff
@@ -0,0 +1,312 @@
...
+    class Vertex:
+        """
+        Vertex in the comparison graph.
+
+        Represents a single function difference and contains both function
+        names, the comparison result, a list of attached non-function
+        differences and a list of its edges (i.e. the direct callees).
+
+        Note: name, file and line are tuples containing the values for both
+        modules.
```
viktormalik: I guess in such case `names`, `files`, and `lines` would be more appropriate names.

diffkemp/simpll/simpll.py
```diff
@@ -19,10 +23,14 @@ def add_suffix(file, suffix):
 
 
 def simplify_modules_diff(first, second, fun_first, fun_second, var,
-                          suffix=None, control_flow_only=False,
-                          print_asm_diffs=False, verbose=False):
+                          suffix=None, equal_funs=None,
+                          control_flow_only=False, print_asm_diffs=False,
+                          verbose=False):
```
viktormalik: I think that this is a good time to rename this function, since it no longer has much to do with simplifying, it performs semantic comparison running SimpLL. Perhaps the best name would be `run_simpll`.

diffkemp/simpll/simpll.py
```diff
@@ -75,41 +86,86 @@ def simplify_modules_diff(first, second, fun_first, fun_second, var,
...
+                            (res_left["function"], res_right["function"]),
+                            Result.Kind.from_string(fun_result["result"]),
+                            (res_left["file"] if "file" in res_left else None,
+                             res_right["file"] if "file" in res_right
+                             else None),
+                            (res_left["line"] if "line" in res_left
+                             else None,
+                             res_right["line"] if "line" in res_right
+                             else None)
+                        )
```
viktormalik: The call is quite long, add explicit parameter names. Or, even better, make the `Vertex` constructor take `fun_result` as a parameter and parse it inside the constructor.

diffkemp/simpll/simpll.py
```diff
@@ -75,41 +86,86 @@ def simplify_modules_diff(first, second, fun_first, fun_second, var,
...
+                                if "body-first" in nonfun_diff:
+                                    # Syntax difference
+                                    diff_obj = ComparisonGraph.SyntaxDiff(
+                                        nonfun_diff["name"],
+                                        nonfun_diff["function"],
+                                        (nonfun_diff["stack-first"],
+                                         nonfun_diff["stack-second"]),
+                                        (nonfun_diff["body-first"],
+                                         nonfun_diff["body-second"])
+                                    )
```
viktormalik: Same as above for `Vertex` - the constructor could take `nonfun_diff` as a single parameter, which would simplify this function heavily.

diffkemp/simpll/simpll.py
```diff
@@ -75,41 +86,86 @@ def simplify_modules_diff(first, second, fun_first, fun_second, var,
...
+                                    diff_obj = ComparisonGraph.TypeDiff(
+                                        nonfun_diff["name"],
+                                        nonfun_diff["function"],
+                                        (nonfun_diff["stack-first"],
+                                         nonfun_diff["stack-second"]),
+                                        (nonfun_diff["file-first"],
+                                         nonfun_diff["file-second"]),
+                                        (nonfun_diff["line-first"],
+                                         nonfun_diff["line-second"]),
+                                    )
```
viktormalik: Same as above.

diffkemp/semdiff/function_diff.py
lenticularis39: Actually just `for side in ComparisonGraph.Side` works (and is already used in `normalize`).

diffkemp/simpll/simpll.py
lenticularis39: The idea was to keep the YAML parsing in `simpll.py`, keeping only code related to the graph logic in `caching.py` (this also applies to other calls in the function).

diffkemp/semdiff/function_diff.py
lenticularis39: I'd rather keep it explicit that "equal" means "equal or assumed equal" here; wrapping it into a method called `has_equal_results` seems not to convey the desired information.

diffkemp/semdiff/function_diff.py
lenticularis39: It could, but that would require having a function take two different types of values as its parameter, since a dictionary is used in one case and a class in the other. (And the use of a trick like renaming the YAML attributes and passsing `call.__dict__` in the other case seems ugly to me.)

viktormalik: Right, in that case, at least enclose this to a function, too. I.e., we'll have two functions with same functionality for two different types.

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -51,18 +59,22 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
...
+                ComparedFuns.at({FirstFun, SecondFun}).kind =
+                        Result::ASSUMED_EQUAL;
             else if (FirstFunName != SecondFunName)
-                ComparedFuns.at({FirstFun, SecondFun}) = Result::NOT_EQUAL;
+                ComparedFuns.at({FirstFun, SecondFun}).kind = Result::NOT_EQUAL;
             else {
                 // One function has a body, the second one does not; add the
                 // missing definition
+                ComparedFuns.at({FirstFun, SecondFun}).kind =
+                        Result::ASSUMED_EQUAL;
```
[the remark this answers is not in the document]
lenticularis39: I discovered (after reading my original commit message) that this is actually just a wrong implementation - the idea was to set assumed equal only in cases when the function names are the same, but I apparently didn't read the if condition correctly.

lenticularis39:
Note: this is not ready for merging yet due to the merge conflict and also the caching of header functions hasn't been resolved.

diffkemp/semdiff/function_diff.py
```diff
@@ -159,7 +159,8 @@ def functions_semdiff(first, second, fun_first, fun_second, config):
 
 def functions_diff(mod_first, mod_second,
                    fun_first, fun_second,
-                   glob_var, config):
+                   glob_var, config,
+                   cache=None, simpll_cache=None):
```
viktormalik: The naming here is a bit confusing - what is the difference between `cache` and `simpll_cache`? And how does `graph` come into it? We should probably distinguish between a comparison graph (a structure that holds all results obtained so far) and a cache (a set of files generated from the graph).

diffkemp/semdiff/caching.py
```diff
@@ -0,0 +1,636 @@
...
+        return []
+    edges = [backtracking_map[end_vertex]]
+    while edges[-1].parent_vertex is not start_vertex:
+        edges.append(backtracking_map[edges[-1].parent_vertex])
+    return list(reversed(edges))
+
+
+class SimpLLCache:
+    """
+    A counterpart to ComparisonGraph responsible for providing the required
```
viktormalik: This is not really a "counterpart", it's more like a _class responsible for providing the results stored in the comparison graph to SimpLL in the form of a cache so that SimpLL does not have to re-compare functions that have been compared before_
or _a class responsible for generating a cache from the comparison result that SimpLL can use to prevent repeated comparison of functions that have been compared before_

diffkemp/simpll/Config.h
```diff
@@ -19,6 +19,7 @@
 #include <llvm/IR/Module.h>
 #include <llvm/IRReader/IRReader.h>
 #include <llvm/Support/SourceMgr.h>
+#include <set>
```
viktormalik: Seems as an unused import.

diffkemp/simpll/ModuleComparator.h
```diff
@@ -116,17 +48,22 @@ class ModuleComparator {
     /// DebugInfo class storing results from analysing debug information
     const DebugInfo *DI;
 
+    /// Cache used for dynamic lookup of already compared functions using
+    /// data passed from DiffKemp.
+    Cache *C;
```
viktormalik: `C` is definitely not a good name. Use `FunctionCache` or `ResultsCache`. Or, maybe better, call the class `ResultsCache` and the variable `Cache`.

diffkemp/simpll/Transforms.cpp
```diff
@@ -156,13 +157,15 @@ void simplifyModulesDiff(Config &config, ComparisonResult &Result) {
                                                         config.FirstFun),
                  mam.getResult<CalledFunctionsAnalysis>(*config.Second,
                                                         config.SecondFun));
+    Cache C(config.CacheDir);
```
viktormalik: Same as above, `C` is not a good variable name.

diffkemp/semdiff/function_diff.py
lenticularis39: `cache` is the graph from the previous comparison, `graph` is the current graph (which represents the result of SimpLL before merging and all results up to this KABI symbol after the merging), `simpll_cache` is passed *to* SimpLL (while `graph` is passed *from* SimpLL) - the naming is indeed quite confusing, I should definitely improve it.

diffkemp/simpll/ModuleComparator.h
lenticularis39: The renaming of the class seems like a good idea - it will also make it clear that it caches results and not something else (like macros etc.).

diffkemp/diffkemp.py
```diff
@@ -234,6 +236,8 @@ def compare(args):
         else:
             group_dir = None
 
+        graph = None
```
viktormalik: `graph` is an extremely generic name - could we change it to `results_graph`?

diffkemp/semdiff/function_diff.py
```diff
@@ -173,8 +174,14 @@ def functions_diff(mod_first, mod_second,
...
     :param glob_var: Global variable whose effect on the functions to compare
     :param config: Configuration
+    :param previous_graph: Graph generated by the previous comparison (used to
+    pass already known results to be used in this comparison).
+    :param function_cache: Cache for SimpLL containing all functions
+    present in the current graph (passed to this function to be updated with
+    the results of the comparison).
     """
     result = Result(Result.Kind.NONE, fun_first, fun_second)
+    current_graph = None
```
viktormalik: Same as above, `current_graph` -> `current_result_graph`.

diffkemp/semdiff/function_diff.py
```diff
@@ -159,7 +159,8 @@ def functions_semdiff(first, second, fun_first, fun_second, config):
 
 def functions_diff(mod_first, mod_second,
                    fun_first, fun_second,
-                   glob_var, config):
+                   glob_var, config,
+                   previous_graph=None, function_cache=None):
```
viktormalik: Same as elsewhere, I'd suggest replacing `previous_graph` by `previous_result_graph` (or some shorter variant)

diffkemp/semdiff/result.py
```diff
@@ -46,6 +61,7 @@ def __init__(self, kind, first_name, second_name):
         self.second = Result.Entity(second_name)
         self.diff = None
         self.macro_diff = None
+        self.graph = None
```
viktormalik: Here it makes sense to use `graph` since the class is called `Result`.

diffkemp/simpll/simpll.py
```diff
@@ -75,41 +81,36 @@ def simplify_modules_diff(first, second, fun_first, fun_second, var,
         first_out = LlvmKernelModule(first_out_name)
         second_out = LlvmKernelModule(second_out_name)
 
-        objects_to_compare = []
         missing_defs = None
-        syndiff_defs = None
         try:
+            graph = ComparisonGraph()
```
viktormalik: Same as elsewhere, I'd suggest using `results_graph` or `comparison_graph` rather than just `graph`.

PR #143, opened 2020-02-05
Linking externally defined global constants 

### LINKED ISSUE #43: Find and link sources containing definitions of global constants
If there is a syntactic difference in names of constant global variables that are declared as external, try to find sources that define these constants, link them to the current module, and rerun the analysis.

A case when this is needed can be reproduced by running:

    bin/diffkemp compare snapshots/linux-3.10.0-693.el7 snapshots/linux-3.10.0-862.el7 -f sprintf [--show-diff]

There is a difference between constants `uuid_be_index` and `uuid_index` that are defined in other source (`lib/uuid.c`) but their value is the same. The source can be discovered by CScope. The same applies for a pair of variables `uuid_le_index` and `guid_index`.

This could be implemented in a similar manner as the inlining of missing function definitions.

tmalecova:
If there is a syntactic difference in names of constant global variables
that are declared as external, sources that define these constants are
found, linked to the current module, and the analysis rerun. 
This closes #43.

Note: Implementation of MissingDefs is changed from const *Function pair
to const *GlobalValue pair.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -965,7 +965,16 @@ int DifferentialFunctionComparator::cmpGlobalValues(GlobalValue *L,
                 }
             }
             return 0;
-        } else
+
+        } else if (NameL != NameR && GVarL && GVarR && GVarL->isConstant()
+                   && GVarR->isConstant() && !GVarL->hasInitializer()
+                   && !GVarR->hasInitializer()) {
+            // Externally defined constants (those without initializer
+            // and with different names) need to have their definitions linked.
+            ModComparator->MissingDefs.push_back({GVarL, GVarR});
+        }
+
+        else
             return 1;
     } else
         return L != R;
--- a/diffkemp/simpll/ModuleComparator.h
+++ b/diffkemp/simpll/ModuleComparator.h
@@ -111,7 +111,7 @@ class ModuleComparator {
     // Counter of assembly diffs
     int asmDifferenceCounter = 0;
 
-    std::vector<ConstFunPair> MissingDefs;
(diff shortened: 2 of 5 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -965,7 +965,12 @@ int DifferentialFunctionComparator::cmpGlobalValues(GlobalValue *L,
                 }
             }
             return 0;
-        } else
+        } else if (NameL != NameR && GVarL && GVarR && !GVarL->hasInitializer()
+                   && !GVarR->hasInitializer()) {
```
viktormalik: I think that we should also add `&& GVarL->isConstant() && GVarR->isConstant`. Preferably add this before checking that the variables have initializers.

viktormalik: Also add a comment here explaining what happens. Something like "Externally defined constants (those without initializer) need to have their definitions linked."

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -965,7 +965,15 @@ int DifferentialFunctionComparator::cmpGlobalValues(GlobalValue *L,
                 }
             }
             return 0;
-        } else
+            // Externally defined constants (those without initializer
+            // and with different names) need to have their definitions linked.
```
viktormalik: The comment should go below the condition (so that it is inside the block that it belongs to).

viktormalik:
There is a merge conflict, so you need to rebase your branch onto the current master. The problem is that the `rhel-74-75.yaml` file has changed (there are some new parts added to the file).
Run `git rebase master` from your branch and then resolve the conflict using `git mergetool`.

PR #144, opened 2020-02-07
Regression tests: separate control-flow-only tests

viktormalik:
There are some bugs in control-flow-only slicing that cause some tests that do not need it to fail. Sometimes, this happens non-deterministically and on some LLVM versions only. Therefore, we put
all regression tests that require the control-flow-only flag into separate spec files.

Also, fix indentation of all YAML spec files to 2 spaces.

### DIFF
--- a/diffkemp/simpll/Utils.cpp
+++ b/diffkemp/simpll/Utils.cpp
@@ -292,7 +292,9 @@ void simplifyFunction(Function *Fun) {
     FunctionPassManager fpm(false);
     FunctionAnalysisManager fam(false);
     pb.registerFunctionAnalyses(fam);
+#if LLVM_VERSION_MAJOR > 5
     fpm.addPass(SimplifyCFGPass{});
+#endif
     fpm.addPass(DCEPass{});
     fpm.addPass(NewGVNPass{});
     fpm.run(*Fun, fam);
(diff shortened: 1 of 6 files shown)


lenticularis39:
It's strange that the PR tests fail (on a test without CFO slicing) - they are supposed to be the same as the branch ones, because the branch is derived directly from the current state of master (i.e. after merging the branch stays the same).

viktormalik:
Indeed, there is a very strange bug that I originally thought was caused by CFO slicing, but it's not. I'm debugging it right now.

viktormalik:
There seems to be a bug in `NewGVNPass` in LLVM 5.0. It removes some code that it should not remove which causes comparison of some function to fail. Moreover, this happens in a non-deterministic way.
This problem could use more investigation, however, I'm afraid that we may be force to end support for LLVM 5.0.

viktormalik:
If you want to investigate this, the failing test is the one for `sprintf` in `rhel-74-75.yaml`, the failing function is `ip6_addr_string_sa`, it is incorrectly simplified after inlining, when `NewGVNPass` is removed, tests always pass.

viktormalik:
After an investigation, there seems to be a bug in LLVM 5.0 that appears when using `NewGVNPass` in combination with `SimplifyCFGPass`. It removes instructions that should not be removed, which causes a false positive in DiffKemp. The new commit in this PR suppresses such combination by not running `SimplifyCFGPass` for LLVM 5.0.

PR #148, opened 2020-03-03
SimpLL optimisations

viktormalik:
This PR features two major SimpLL optimisations:
- finding macro differences: macro definitions extracted from debug info and macro usages for each program location are cached so that every macro and every line needs to be parsed only once
- field access abstractions: abstractions are created only if all GEPs have constant indices and comparison of two abstractions having a single instruction is done using the original `cmpGEPs` method

Also, the diffs found within a KABI symbol are sorted so that comparison of large logs is easier.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -249,7 +249,7 @@ int DifferentialFunctionComparator::cmpOperations(
                         }
                     }
 
-                    if (Result && controlFlowOnly
+                    if (Result && config.ControlFlowOnly
                         && abs(CL->getNumOperands() - CR->getNumOperands())
                                    == 1) {
                         needToCmpOperands = false;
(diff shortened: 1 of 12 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -453,8 +454,10 @@ void DifferentialFunctionComparator::findMacroFunctionDifference(
     // First look whether this is the case described above.
     auto LineL = extractLineFromLocation(L->getDebugLoc());
     auto LineR = extractLineFromLocation(R->getDebugLoc());
-    auto MacrosL = getAllMacrosAtLocation(L->getDebugLoc(), L->getModule());
-    auto MacrosR = getAllMacrosAtLocation(R->getDebugLoc(), R->getModule());
+    auto MacrosL = ModComparator->MacroDiffs.getAllMacroUsesAtLocation(
```
lenticularis39: Here a reference could also be used.

diffkemp/simpll/SourceCodeUtils.cpp
```diff
@@ -520,15 +519,15 @@ std::vector<std::string> findInlineAssemblySourceArguments(
     std::string line = extractLineFromLocation(LineLoc);
     if (line == "")
         return {};
-    auto MacroMap = getAllMacrosAtLocation(LineLoc, Mod);
+    auto MacroMap = MacroDiffs->getAllMacroUsesAtLocation(LineLoc, 0);
```
lenticularis39: See above.

PR #153, opened 2020-03-17
Add unit tests for SimpLL.

lenticularis39:
Currently tests cover only DifferentialFunctionComparator (without macro and type differences) and need LLVM 8 or greater because of differences in LLVM classes (especially in debug information classes).

### DIFF
--- a/tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
+++ b/tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
@@ -1,0 +1,1191 @@
+//===--------- DifferentialFunctionComparatorTest.cpp - Unit tests ---------==//
+//
+//       SimpLL - Program simplifier for analysis of semantic difference      //
+//
+// This file is published under Apache 2.0 license. See LICENSE for details.
+// Author: Tomas Glozar, tglozar@gmail.com
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file contains unit tests for the DifferentialFunctionComparator class,
+/// along with necessary classes and fixtures used by them.
+///
+//===----------------------------------------------------------------------===//
+
+#include <Config.h>
+#include <DebugInfo.h>
+#include <DifferentialFunctionComparator.h>
+#include <ModuleComparator.h>
+#include <ResultsCache.h>
+#include <gtest/gtest.h>
+#include <passes/FieldAccessFunctionGenerator.h>
+#include <passes/StructureDebugInfoAnalysis.h>
+#include <passes/StructureSizeAnalysis.h>
+
+#if LLVM_VERSION_MAJOR > 7
+/// This class is used to expose protected functions in
+/// DifferentialFunctionComparator.
+class TestComparator : public DifferentialFunctionComparator {
+  public:
+    TestComparator(const Function *F1,
+                   const Function *F2,
+                   const Config &config,
+                   const DebugInfo *DI,
+                   ModuleComparator *MC)
+            : DifferentialFunctionComparator(F1, F2, config, DI, MC) {}
+    int testCompareSignature(bool keepSN = false) {
+        if (!keepSN)
+            beginCompare();
+        return compareSignature();
+    }
+
+    int testCmpAttrs(const AttributeList L,
+                     const AttributeList R,
+                     bool keepSN = false) {
+        if (!keepSN)
+            beginCompare();
+        return cmpAttrs(L, R);
+    }
+
+    int testCmpAllocs(const CallInst *CL,
+                      const CallInst *CR,
+                      bool keepSN = false) {
+        if (!keepSN)
+            beginCompare();
+        return cmpAllocs(CL, CR);
+    }
+
+    int testCmpConstants(const Constant *CL,
+                         const Constant *CR,
+                         bool keepSN = false) {
+        if (!keepSN)
+            beginCompare();
+        return cmpConstants(CL, CR);
+    }
+
+    int testCmpMemset(const CallInst *CL,
+                      const CallInst *CR,
+                      bool keepSN = false) {
+        if (!keepSN)
+            beginCompare();
+        return cmpMemset(CL, CR);
+    }
+
+    int testCmpCallsWithExtraArg(const CallInst *CL,
+                                 const CallInst *CR,
+                                 bool keepSN = false) {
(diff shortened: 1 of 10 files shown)


tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -0,0 +1,1159 @@
...
+    new AllocaInst(Type::getInt8Ty(CtxR), 0, "var2", BBR);
+    ReturnInst::Create(CtxL, BBL);
+    ReturnInst::Create(CtxR, BBR);
+
+    ASSERT_EQ(DiffComp->testCmpBasicBlocks(BBL, BBR), 0);
+    ASSERT_EQ(DiffComp->testCmpBasicBlocks(BBR, BBL), 0);
+}
+
+/// Tests the comparison of global values.
+TEST_F(DifferentialFunctionComparatorTest, CmpGlobalValues) {
```
viktormalik: This test could be split into several smaller tests. I suggest doing so, since it will improve the developer experience - if a test fails, the developer needs to read and understand a short test only.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -0,0 +1,1159 @@
...
+    GVR1->setName("missing2");
+    ASSERT_EQ(DiffComp->testCmpGlobalValues(GVL1, GVR1), 1);
+    ASSERT_EQ(ModComp->MissingDefs.size(), 1);
+    ASSERT_EQ(ModComp->MissingDefs[0].first, GVL1);
+    ASSERT_EQ(ModComp->MissingDefs[0].second, GVR1);
+}
+
+/// Tests comparison of values that are behind a cast (in some cases the cast
+/// is supposed to be ignored, in some cases not).
+TEST_F(DifferentialFunctionComparatorTest, CmpValuesCastIgnore) {
```
viktormalik: Same as above - this test can be split into several smaller test.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -0,0 +1,1159 @@
...
+    /// Creates an enviroment for cmpGlobalValues to be able to compare
+    /// functions and then calls it via cmpBasicBlocks.
+    int testFunctionComparison(Function *FunL, Function *FunR) {
+        const std::string auxFunName = "AuxFunComp";
+
+        // Testing function comparison is a little bit tricky, because for the
+        // callee generation the call location must be set at the time the
+        // comparison is done.
+        // To ensure this a pair of auxilliary functions containing a call to
+        // the functions is added, along with their locations.
```
viktormalik: This comment is quite hard to understand, I did not get it until after many readings. It should be at least improved so that it explains better why function comparison is done in such a complicated way. The comment should say that the functions `FunL` and `FunR` cannot be compared directly since `DifferentialFunctionComparator` expects a single pair of functions to compare (and such have already been provided in the constructor - `FL` and `FR`) and should another pair need to be compared, a new comparator must be created.
Anyway, why not do exactly that instead of doing this complicated stuff with function calls? Why not create a new (temporary) `TestComparator` object that will compare `FunL` and `FunR`?

lenticularis39: The point of `testFunctionComparison` isn't to test the comparison of `FunL` and `FunR` using `DifferentialFunctionComparator`'s `compare` method (this is unnecessary because the original method from `FunctionComparator` is used, which has already been tested in LLVM unit tests). 

The function is used for testing the part of `cmpGlobalValues` that compares functions (i.e. it checks whether it returns the correct return value and whether it calls `ModComparator->compareFunctions` properly). This part uses the private variables `CurrentLocL` and `CurrentLocR` that are set in `cmpBasicBlocks` based on the debug location of the instructions (call instructions in this case), hence the creation of the calls, their debug locations and using `cmpBasicBlocks` to get the result.

tests/unit_tests/simpll/SimpLLTest.cpp
```diff
@@ -1,8 +1,6 @@
 #include <gtest/gtest.h>
 
-TEST(EmptyTest, EmptyVariant) {
-    ASSERT_EQ(true, true);
-}
+TEST(EmptyTest, EmptyVariant) { ASSERT_EQ(true, true); }
```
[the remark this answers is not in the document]
lenticularis39: No, it's not, I'll remove it.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -0,0 +1,1159 @@
...
+#include <DifferentialFunctionComparator.h>
+#include <ModuleComparator.h>
+#include <ResultsCache.h>
+#include <gtest/gtest.h>
+#include <passes/FieldAccessFunctionGenerator.h>
+#include <passes/StructureDebugInfoAnalysis.h>
+#include <passes/StructureSizeAnalysis.h>
+
+#if LLVM_VERSION_MAJOR > 7
+class TestComparator : public DifferentialFunctionComparator {
```
viktormalik: This class should have a comment explaining its purpose.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -0,0 +1,1159 @@
...
+            beginCompare();
+        return cmpTypes(TyL, TyR);
+    }
+
+    void setLeftSerialNumber(const Value *Val, int i) { sn_mapL[Val] = i; }
+
+    void setRightSerialNumber(const Value *Val, int i) { sn_mapR[Val] = i; }
+};
+
+class DifferentialFunctionComparatorTest : public ::testing::Test {
```
viktormalik: Same as above - the class should have a comment explaining its purpose.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
viktormalik: I see, ok, then the comment should say this. Also, it is a bit strange that this is implemented as a class method while all other tests are implemented as standard GTest tests. The reason that I see here is that this function is used several times, but the function comment should probably state that.

PR #155, opened 2020-03-25
Support calling SimpLL through Python CFFI.

lenticularis39:
C interface was added to SimpLL (in files FFI.cpp and FFI.h) and support
for calling SimpLL through it using CFFI was added to DiffKemp.

The compare option --enable-simpll-ffi enables this feature. Currently it
doesn't work well with llvmcpy, so it is disabled in tests and not
enabled by default, but can be used with KABI comparisons to improve
performance.

The CFFI module was added to setup.py so it is built together with DiffKemp.
This means that SimpLL has now to be already built in the "build" directory
before installing DiffKemp since the module links the auxilliary static
library simpll-lib (libsimpll-lib.a).

In addition an option to CMake to (re)build the CFFI module while building
SimpLL (with an dependency on simpll-lib) called SIMPLL_REBUILD_BINDINGS
was added. It is not enabled by default because it needs CFFI to be
installed, which is typically done in setup.py and setup.py needs SimpLL
to be already built.


diffkemp/simpll/FFI.cpp
```diff
@@ -0,0 +1,61 @@
...
+///
+//===----------------------------------------------------------------------===//
+
+#include "FFI.h"
+#include "Config.h"
+#include "Output.h"
+#include <cstring>
+
+extern "C" {
+void runSimpLL(const char *ModL,
```
viktormalik: A large part of this function is common with `main` in `SimpLL.cpp`. I suggest to put what is common into a new function. It could prove useful if we change the analysis workflow in future (e.g. if post-processing is re-introduced). 

I'd put the function into `Transforms`. Actually, while looking into it now, the `Transforms` file should probably be renamed to something like `Analysis` or `MainAnalysis` (other suggestions?) since it would better express its purpose.

viktormalik:
I have a question, just to make sure. The CFFI module is built automatically when the DiffKemp Python module is installed and the purpose of the `SIMPLL_REBUILD_BINDINGS` option in CMake is to enable re-building of the CFFI module in case when SimpLL is changed (we don't need to re-install the Python module in such case)?

lenticularis39:
> I have a question, just to make sure. The CFFI module is built automatically when the DiffKemp Python module is installed and the purpose of the `SIMPLL_REBUILD_BINDINGS` option in CMake is to enable re-building of the CFFI module in case when SimpLL is changed (we don't need to re-install the Python module in such case)?
Yes - the rebuilding of the CFFI module takes much less time than re-installing DiffKemp.

diffkemp/simpll/FFI.cpp
lenticularis39: Maybe `ModuleAnalysis` would be a good name - it makes it clear that it does more than `ModuleComparator` and that it operates on modules.

diffkemp/simpll/ModuleAnalysis.h
```diff
@@ -1,4 +1,4 @@
-//===------------- Transforms.h - Simplifications of modules --------------===//
+//===----------- ModuleAnalysis.h - Simplifications of modules ------------===//
```
viktormalik: Could you also change the header comment for this file (and the .cpp file) so that it explains its current purpose better?

lenticularis39: That's a good idea - the current header omits the comparison part, which has been a part of the file for a long time.

viktormalik: I meant the entire file commentary. It now says that the file contains declarations of functions that do simplification, but it should be something like:
"This file contains the declaration of functions that do the actual semantic comparison of functions (modules)."

Analogously for ModuleAnalysis.cpp

PR #156, opened 2020-03-27
Support LLVM 10.

lenticularis39:
FunctionComparator from LLVM 10 was added to llvm-lib (it is actually
the same as in LLVM 9) and a now depracated call to
LLVMBuilder::CreateMemCpy was replaced with a call to the new function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ jobs:
     - env: LLVM_VERSION=7
     - env: LLVM_VERSION=8
     - env: LLVM_VERSION=9
+    - env: LLVM_VERSION=10
 
 addons:
   apt:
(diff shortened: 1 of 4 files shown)


lenticularis39:
I think it's a good idea considering the older LLVM versions don't have a specific use (except for using LLReve). Also now when comparing the entire KABI of two kernels takes a few minutes it would be also possible to do that regularly, too, maybe less frequently than the all LLVM versions check.

viktormalik:
Good, it should suffice to add this to `.travis.yml`:

    - env: LLVM_VERSION=5.0
      if: type = cron

lenticularis39:
Looks like it works.

PR #157, opened 2020-04-02
Several improvements and fixes

viktormalik:
The proposed fixes are:
- disable `asm inline` constructs not supported by older LLVM versions
- always pass relative path to the object file to `make`
- if an LLVM module includes a file outside of the kernel source tree, do not copy it into the snapshot
- when missing definitions are found, and the linking is successful, do not merge the comparison graphs since the new one contains invalid results that will be re-computed

### DIFF
--- a/diffkemp/llvm_ir/build_llvm.py
+++ b/diffkemp/llvm_ir/build_llvm.py
@@ -305,7 +305,7 @@ class LlvmKernelBuilder:
             cwd = os.getcwd()
             os.chdir(self.kernel_dir)
 
-            name = source_file[:-2]
+            name = os.path.relpath(source_file, self.kernel_dir)[:-2]
             try:
                 # Get GCC command for building the .o file
                 command = self.kbuild_object_command("{}.o".format(name))
--- a/diffkemp/llvm_ir/build_llvm.py
+++ b/diffkemp/llvm_ir/build_llvm.py
@@ -28,37 +28,47 @@ class LlvmKernelBuilder:
         """
         Prepare kernel so that it can be compiled into LLVM IR.
         """
-        self._disable_asm_goto()
+        self._disable_asm_features()
 
     def finalize(self):
         """Restore the kernel state."""
-        self._enable_asm_goto()
+        self._enable_asm_features()
 
-    def _disable_asm_goto(self):
+    def _disable_asm_features(self):
         """
-        Transform 'asm goto(x)' command into 'asm("goto(x)")'.
-        This is because LLVM does not support asm goto yet.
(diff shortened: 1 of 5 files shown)


diffkemp/llvm_ir/build_llvm.py
```diff
@@ -28,37 +28,45 @@ def initialize(self):
...
         """
         for header in self.compiler_headers:
-            command = ["sed", "-i", "s/asm goto(x)/asm (\"goto(\" #x \")\")/g",
-                       header]
+            commands = [
+                ["sed", "-i", "s/asm goto(x)/asm (\"goto(\" #x \")\")/g",
+                 header],
+                ["sed", "-i",
+                 "/#ifdef CONFIG_CC_HAS_ASM_INLINE/i "
+                 "#undef CONFIG_CC_HAS_ASM_INLINE",
```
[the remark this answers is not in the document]
viktormalik: That should never happen since `CONFIG_CC_HAS_ASM_INLINE` is set automatically. Still, this is a good idea, I'll add a marking commentary.

diffkemp/semdiff/function_diff.py
```diff
@@ -227,30 +242,14 @@ def functions_diff(mod_first, mod_second,
                                            [Result.Kind.UNKNOWN,
                                             Result.Kind.ASSUMED_EQUAL]])
 
-                if prev_result_graph:
+                if prev_result_graph and not simplify:
```
lenticularis39, line 245: This works, but I think there might be an even better solution. When `simplify` is `True`, it means that SimpLL will be re-run anyway, therefore it is unnecessary to do the function cache update and `graph_to_fun_pair_list`. This would also allow us to get rid of the function cache rollback.

viktormalik: I moved the cache update into this condition. May I remove the cache rollback completely?

lenticularis39: Yes, now the cache is not updated when `simplify` is `True`, therefore there is nothing to roll back.

Also the call to `graph_to_fun_pair_list` can be moved outside the `while simplify` loop.

viktormalik: Done, please review.

lenticularis39:
Looks good, I'm currently doing some tests on whole snapshots to ensure nothing is broken after the changes in `function_diff` which should be finished soon, enabling me to approve the PR.

viktormalik:
> Looks good, I'm currently doing some tests on whole snapshots to ensure nothing is broken after the changes in `function_diff` which should be finished soon, enabling me to approve the PR.

Great, thanks!

PR #161, opened 2020-04-03
SimpLL: Add SeparateCallsToBitcastPass pass

### LINKED ISSUE #42: llvm::InlineFunction cannot inline a call to a bitcast operator
This can be resolved by running the InstCombine pass, but this pass runs a lot of other transformations which cause the compared functions to diverge and to be compared by SimpLL as non-equal. Hence, it would be needed just to run the transformation of calls (see `InstCombiner::transformConstExprCastCall` in `lib/Transforms/InstCombine/InstCombineCalls.cpp`). The problem is that this function is not public in LLVM (nor is the `InstCombiner` class).

The problem can be reproduced by running:

    bin/diffkemp compare snapshots/linux-3.10.0-693.el7 snapshots/linux-3.10.0-862.el7 -f kill_block_super

PSilling:
The SeparateCallsToBitcastPass pass separates bitcasts from calls to
bitcast operators by inserting new bitcast instructions that transform
function arguments and the return value separately.

This change makes these calls inlinable during the comparison phase. Fixes #42.

Example:
`call void bitcast (void (%struct.super_block.30*)* @fsnotify_unmount_inodes to void (%struct.super_block*)*)(%struct.super_block* %0), !dbg !25663`

will be changed to:
`%12 = bitcast %struct.super_block* %0 to %struct.super_block.30*, !dbg !25663`
`call void @fsnotify_unmount_inodes(%struct.super_block.30* %12), !dbg !25663`

### DIFF
--- a/diffkemp/simpll/Utils.h
+++ b/diffkemp/simpll/Utils.h
@@ -17,6 +17,7 @@
 #include <llvm/IR/Constants.h>
 #include <llvm/IR/DebugInfoMetadata.h>
 #include <llvm/IR/Function.h>
+#include <llvm/IR/Instructions.h>
 #include <unordered_map>
 
 using namespace llvm;
@@ -27,6 +28,9 @@ enum Program { First, Second };
 typedef std::pair<Function *, Function *> FunPair;
 typedef std::pair<const Function *, const Function *> ConstFunPair;
 typedef std::pair<const GlobalValue *, const GlobalValue *> GlobalValuePair;
+
+// Invalid attributes for void functions and calls.
+extern std::vector<Attribute::AttrKind> badVoidAttributes;
 
 /// Extract called function from a called value. Handles situation when the
 /// called value is a bitcast.
@@ -130,6 +134,12 @@ std::string getIdentifierForValue(
         const std::map<std::pair<StructType *, uint64_t>, StringRef>
                 &StructFieldNames,
         const Function *Parent = nullptr);
+
+/// Copies properties from one call instruction to another.
+void copyCallInstProperties(CallInst *srcCall, CallInst *destCall);
+
+/// Copies properties from one function to another.
+void copyFunctionProperties(Function *srcFun, Function *destFun);
 
 /// Converts value to its string representation.
 /// Note: Currently the only place that calls this is returns.gdb, which lacks
--- a/diffkemp/simpll/Utils.h
+++ b/diffkemp/simpll/Utils.h
@@ -29,9 +29,6 @@ typedef std::pair<Function *, Function *> FunPair;
 typedef std::pair<const Function *, const Function *> ConstFunPair;
 typedef std::pair<const GlobalValue *, const GlobalValue *> GlobalValuePair;
 
-// Invalid attributes for void functions and calls.
-extern std::vector<Attribute::AttrKind> badVoidAttributes;
-
 /// Extract called function from a called value. Handles situation when the
 /// called value is a bitcast.
 const Function *getCalledFunction(const Value *CalledValue);
--- a/diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
+++ b/diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
@@ -1,0 +1,141 @@
+//===----  SeparateCallsToBitcastPass.cpp - Separate calls to bitcasts ----===//
+//
+//       SimpLL - Program simplifier for analysis of semantic difference      //
+//
+// This file is published under Apache 2.0 license. See LICENSE for details.
+// Author: Petr Silling, psilling@redhat.com
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file contains the definition of the SeparateCallsToBitcastPass pass.
+/// The pass separates bitcasts from calls to bitcast operators by inserting
+/// new bitcast instructions that transform function arguments and the return
+/// value separately.
+///
+//===----------------------------------------------------------------------===//
+
+#include "SeparateCallsToBitcastPass.h"
+
+#include "Utils.h"
+#include <Config.h>
+#include <llvm/IR/Instructions.h>
+#include <llvm/IR/Operator.h>
+
+/// Separate bitcasts from calls to bitcast operators to make the calls
+/// inlinable.
+PreservedAnalyses
+        SeparateCallsToBitcastPass::run(Function &Fun,
+                                        FunctionAnalysisManager &fam) {
+    std::vector<Instruction *> toRemove;
+
+    for (auto &BB : Fun) {
+        for (auto &Inst : BB) {
+            if (auto Call = dyn_cast<CallInst>(&Inst)) {
+                // Find call instructions calling a bitcast
+                // operator that directly corresponds to a function.
+                auto BitCast =
+                        dyn_cast<BitCastOperator>(Call->getCalledValue());
+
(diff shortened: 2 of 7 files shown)


diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
```diff
@@ -0,0 +1,138 @@
...
+                    // arguments is lower than the number of parameters.
+                    if (Call->getNumArgOperands()
+                        < srcFun->getFunctionType()->getNumParams()) {
+                        continue;
+                    }
+
+                    // Ignore the instruction if the source function returns
+                    // void and the return value is used.
+                    if (srcFun->getReturnType()->isVoidTy()
+                        && Call->getType()->isVoidTy() && !Call->use_empty()) {
```
viktormalik, line 54: Shouldn't this be `!Call->getType()->isVoidTy()`? Call can't have any uses if its type is `void`, therefore the second test at this line does not make sense.

diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
```diff
@@ -0,0 +1,138 @@
...
+                    // parameters. If the number of arguments is higher, use the
+                    // remaining arguments without bitcasting as a vargars is
+                    // present in the call instruction.
+                    std::vector<Value *> newArgs;
+                    auto paramType = srcFun->getFunctionType()->param_begin();
+                    auto endParamType = srcFun->getFunctionType()->param_end();
+                    auto arg = Call->arg_begin();
+                    auto endArg = Call->arg_end();
+
+                    while (paramType != endParamType) {
```
viktormalik, line 68: Why not use a range-based loop?

    for (auto paramType: srcFun->getFunctionType()->params())

diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
```diff
@@ -0,0 +1,138 @@
...
+                            newArg->setDebugLoc(Call->getDebugLoc());
+                            newArgs.push_back(newArg);
+                        }
+
+                        ++paramType;
+                        ++arg;
+                    }
+
+                    // Add the remaining arguments if there are any.
+                    while (arg != endArg) {
```
viktormalik, line 91: It doesn't make sense to have `endArg` as a separate variable if it is used only once. Use `Call->arg_end()` instead.

diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
PSilling: Right, I did not realise that a `void` Call cannot have any uses.

diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
PSilling: Agreed (and same for the one below). I was originally using these iterators a bit differently and did not later realise that I can use a range-based loop here, too.

diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
```diff
@@ -0,0 +1,138 @@
...
+                        newArgs.push_back(*arg);
+                        ++arg;
+                    }
+
+                    // Create a new call instruction using the
+                    // source function and bitcasted arguments.
+                    auto newCall = CallInst::Create(srcFun, newArgs, "", Call);
+                    newCall->setDebugLoc(Call->getDebugLoc());
```
lenticularis39, line 99: There are a few more properties that can be copied from the original call like the calling convention, tail call mark or attributes (not all of them are currently compared in `DifferentialFunctionComparator`, but this could change in the future).

diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
```diff
@@ -0,0 +1,138 @@
...
+                }
+            }
+        }
+    }
+
+    // Remove replaced call instructions.
+    for (auto Call : toRemove) {
+        Call->eraseFromParent();
+    }
+    toRemove.clear();
```
lenticularis39, line 135: I think the clearing of the vector can be omitted, since it is done automatically when the variables on the stack are destroyed on leaving the function.

diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
viktormalik: @lenticularis39 could you please point to some code where copying of such attributes is already implemented in DiffKemp? I believe that you have written such a code.

diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
PSilling: Alright, will do.

diffkemp/simpll/passes/SeparateCallsToBitcastPass.cpp
lenticularis39: It's in `RemoveUnusedReturnValuesPass`.

PSilling: Thanks, I can see it there and will look into it.

PSilling: I believe that this code should be fully reusable (and not only in this scenario). Please, correct me if I am wrong. If this is true, however, I also propose to move this into `Utils.cpp`. https://github.com/viktormalik/diffkemp/blob/f321f58586fe0237e1bd8088800c92661f60d640/diffkemp/simpll/passes/RemoveUnusedReturnValuesPass.cpp#L209-L221

lenticularis39: That's a good idea - the function can take a `Instruction *`, copy the generic properties and then use `isa` to determine the instruction type for the purpose of copying the more specific ones.

PSilling:
Should be resolved. Please review the new function. Because there were no uses outside of call instructions, I have decided against the general approach. Therefore, the `copyCallProperties` function works only with call instructions instead (can be generalized in the future, if necessary).

lenticularis39:
I think the best solution is to drop `InvokeInst` support from `RemoveUnusedReturnValuesPass`. Invokes are a higher level operation that won't occur in the kernel code and SimpLL doesn't support them at any other place in SimpLL-specific code (moreover the implementation of e.g. invoke inlining would probably be non-trivial).

viktormalik:
> I think the best solution is to drop `InvokeInst` support from `RemoveUnusedReturnValuesPass`. Invokes are a higher level operation that won't occur in the kernel code and SimpLL doesn't support them at any other place in SimpLL-specific code (moreover the implementation of e.g. invoke inlining would probably be non-trivial).

I agree, remove support for `InvokeInst` and you can use `CallInst` instead of `CallBase`.

diffkemp/simpll/Utils.h
```diff
@@ -28,6 +29,9 @@ typedef std::pair<Function *, Function *> FunPair;
 typedef std::pair<const Function *, const Function *> ConstFunPair;
 typedef std::pair<const GlobalValue *, const GlobalValue *> GlobalValuePair;
 
+// Invalid attributes for void functions and calls.
+extern Attribute::AttrKind badVoidAttributes[13];
```
viktormalik, line 33: Do not use array, use `std::vector` instead.

PR #162, opened 2020-04-06
LLVM build: do not show sed warnings

viktormalik:
There may be warnings when disabling goto features if the altered header does not exist in the given kernel version. Therefore we redirect the sed stderr to /dev/null.

### DIFF
--- a/diffkemp/llvm_ir/build_llvm.py
+++ b/diffkemp/llvm_ir/build_llvm.py
@@ -41,6 +41,8 @@ class LlvmKernelBuilder:
         - disable usage of 'asm inline'
         """
         for header in self.compiler_headers:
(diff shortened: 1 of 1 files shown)


lenticularis39:
Is there a reason why we don't check if the header file exists before calling sed?

viktormalik:
Not really, I was following the Pyhton practice to run the command and then check for exceptions, but the truth is that checking for existence of the file is much more efficient.

lenticularis39:
If the operation wasn't performed by a separate process the exception approach would be more efficient, because it wouldn't have to look up the file twice, but here it has to look up and load the `sed` binary instead.

viktormalik:
True enough, thanks for the explanation, I added checks for file existence.

PR #165, opened 2020-04-08
Handle refactoring by inverting if conditions

viktormalik:
This change allows to handle refactorings of the form:

    if (C) {B1} {B2} -> if (!C) {B2} {B1}

including cases when `return` or `continue` is used in one of the branches.

The PR contains several features:
- Handling comparison of inverse branch conditions. This is done by collecting pairs of `cmp` instructions with inverse predicates and if such instructions are used as conditions in `br` instructions, successors for one of the branches are swapped.
- Improved comparison of PHI instructions that have reordered incoming value/block pairs. This is done by postponing PHI comparison into the end after all other values and blocks were compared.
- Running the `-mergereturn` pass in `opt`. This is to improve handling of situations when an additional `return` is added due to refactoring.

See individual commits for more implementation details.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -29,6 +29,20 @@
 std::set<std::string> ignoredMacroList = {
         "__COUNTER__", "__FILE__", "__LINE__", "__DATE__", "__TIME__"};
 
+/// Run comparison of PHI instructions after comparing everything else. This is
+/// to ensure that values and blocks incoming to PHIs are properly matched in
+/// time of PHI comparison.
+int DifferentialFunctionComparator::compare() {
+    int Res = FunctionComparator::compare();
+    if (Res == 0) {
+        for (auto &PhiPair : phisToCompare)
+            if (cmpPHIs(PhiPair.first, PhiPair.second))
+                return 1;
+        return 0;
+    }
+    return Res;
+}
+
 /// Compare GEPs. This code is copied from FunctionComparator::cmpGEPs since it
 /// was not possible to simply call the original function.
 /// Handles offset between matching GEP indices in the compared modules.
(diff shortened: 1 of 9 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -365,6 +411,18 @@ int DifferentialFunctionComparator::cmpOperations(
...
+        // Handle inverse conditions
+        if (isa<CmpInst>(L) && isa<CmpInst>(R)) {
+            auto CmpL = dyn_cast<CmpInst>(L);
+            auto CmpR = dyn_cast<CmpInst>(R);
+
+            // It is sufficient to compare the predicates here since the
+            // operands are compared in cmpBasicBlocks.
+            if (CmpL->getPredicate() == CmpR->getInversePredicate()) {
+                inverseConditions.emplace(L, R);
+                return 0;
```
lenticularis39, line 423: This seems to compare all `CmpInst`s with the same arguments and inverse predicates as equal, even those that are used at other places than `BranchInst`s. For example these two (clearly semantically different) programs now compare as equal:
```c
#include <stdio.h>

int main() {
    int a = 4, b = 5;
    int cmp = a <= b;
    printf("%d\n", cmp);
}
```
```c
#include <stdio.h>

int main() {
    int a = 4, b = 5;
    int cmp = a > b;
    printf("%d\n", cmp);
}
```
Going through uses of the `CmpInst` and making sure all of them are `BranchInst`s should fix this problem.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -299,6 +313,19 @@ int DifferentialFunctionComparator::cmpOperations(
...
+    // values, treat them as equal for now. They will be compared at the end,
+    // after all their incoming values have been compared and matched.
+    if (isa<PHINode>(L) && isa<PHINode>(R)) {
+        auto PhiL = dyn_cast<PHINode>(L);
+        auto PhiR = dyn_cast<PHINode>(R);
+        if (PhiL->getNumIncomingValues() == PhiR->getNumIncomingValues()) {
+            needToCmpOperands = false;
+            phisToCompare.emplace_back(PhiL, PhiR);
+        }
+        return 0;
```
lenticularis39, line 326: This returns `0` without putting the PHIs to `phisToCompare` when the number of incoming values is different. which probably isn't the correct behavior, moreover this causes an assert in `cmpBasicBlocks` on the equality of the number of operands after `cmpOperations`. I think the `return` should be inside the `if` condition.

PR #166, opened 2020-04-15
RPM: add %check phase

viktormalik:
Run SimpLL unit tests in the check phase (these do not need external
kernel sources present).

Also, Git is required for building because of GTest.

### DIFF
--- a/rpm/diffkemp.spec
+++ b/rpm/diffkemp.spec
@@ -13,6 +13,7 @@ Source1:        %pypi_source llvmcpy %{llvmcpy_version}
 BuildRequires:  gcc gcc-c++ cmake ninja-build
 BuildRequires:  llvm-devel
 BuildRequires:  python3-devel
+BuildRequires:  git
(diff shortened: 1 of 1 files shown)


lenticularis39:
I'm thinking about implementing conditional RPM build checks when the spec changes in CI to get some relevant results from testing - running regular DiffKemp and SimpLL tests doesn't make much sense when the code hasn't changed. (I tested this one manually, but in the future it could be useful to save time.)

viktormalik:
> I'm thinking about implementing conditional RPM build checks when the spec changes in CI to get some relevant results from testing - running regular DiffKemp and SimpLL tests doesn't make much sense when the code hasn't changed. (I tested this one manually, but in the future it could be useful to save time.)

Yeah, that would be nice, however, Travis runs on Ubuntu and it might be difficult to run rpmbuild from there. Maybe it would be possible to hook a Copr build of some diffkemp-testing project.

PR #170, opened 2020-04-23
 Move DiffKemp functions using LLVM IR to SimpLL.

lenticularis39:
Classes for working with SimpLL/LLVM `Module` and `Function` objects were added to the SimpLL Python interface along with their corresponding functions in the SimpLL FFI. 

Methods of `LlvmKernelModule` previously written in Python with the use of llvmcpy were rewritten in C++ and moved into a new SimpLL source file `library/DiffKempUtils.cpp`. The implementation part of `LlvmSysctlModule` was rewritten in C++ as a SimpLL class `SysctlTable` with a corresponding class in the SimpLL Python interface (`SimpLLSysctlTable`) using corresponding functions in the SimpLL FFI. `LlvmSysctlModule` now consists of an interface to `SimpLLSysctlTable`.

This enabled the removal of llvmcpy from DiffKemp dependences, because all functions using it were removed and replaced by the SimpLL library.

Also a new way of calling `runSimpLL` was added to provide support for comparing `SimpLLModule` objects (i.e. modules loaded and parsed using the SimpLL library) along with an experimental module caching option. If the `--enable-module-cache` parameter is enabled, modules used more than a set number of times (currently 3) are pre-loaded before the comparison and when functions in them are compared, they are cloned from the cache instead of being loaded from disk. This option requires SimpLL FFI to be enabled for the run.


diffkemp/simpll/library/FFI.h
```diff
@@ -0,0 +1,121 @@
...
+/// This file contains declarations of C functions and structure types used for
+/// interacting with DiffKemp.
+///
+//===----------------------------------------------------------------------===//
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// CFFI_DECLARATIONS_START
```
viktormalik, line 19: Add a comment (probably above this one) that this line must not be altered nor deleted.

diffkemp/simpll/library/FFI.h
```diff
@@ -0,0 +1,121 @@
...
+                       const char *ModLOut,
+                       const char *ModROut,
+                       const char *FunL,
+                       const char *FunR,
+                       struct config Conf,
+                       char *Output);
+
+void shutdownSimpLL();
+
+// CFFI_DECLARATIONS_END
```
viktormalik, line 117: Same as above - add a warning not to alter or delete this line.

diffkemp/simpll/library/FFI.cpp
```diff
@@ -0,0 +1,303 @@
...
+              ModLOut,
+              ModROut,
+              FunL,
+              FunR,
+              Conf,
+              Output);
+}
+
+void shutdownSimpLL() { llvm_shutdown(); }
+}
```
viktormalik, line 303: Add comment `// extern "C"` so that we know that the scope ends here.

PR #171, opened 2020-04-23
Refactor comparing of instructions and basic blocks

viktormalik:
This PR implements two related refactorings:
1. Move the code for comparison of instructions and of all their operands from `cmpBasicBlocks` into a newly created function `cmpOperationsWithOperands`. This enables the point 2, since the functions for finding syntax differences (differences in macros, inline assembly code, and in types) do not have to be called from two places (once from `cmpOperations` and once from `cmpBasicBlocks` for arguments) anymore.
2. Move all the code for finding syntax differences and for setting functions to be inlined into a single function `findDifferences` that is called from `cmpBasicBlocks` after `cmpOperationsWithOperands` returns a difference. 
Also, a separate function `findTypeDifferenceInFieldAccess` for finding differences in chained field abstractions is introduced since these have to be searched for even if the comparison of operations is equal.

Due to moving all the code for setting `toInline` into `cmpBasicBlocks`, the `CmpOperationsInlining` is merged into the `CmpBasicBlocksInlining`.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -1186,8 +1186,26 @@ int DifferentialFunctionComparator::cmpOperationsWithOperands(
     // convenient to have the code in a separate function.
 
     bool needToCmpOperands = true;
-    if (int Res = cmpOperations(L, R, needToCmpOperands))
+    if (int Res = cmpOperations(L, R, needToCmpOperands)) {
+        auto *CallL = dyn_cast<CallInst>(L);
+        auto *CallR = dyn_cast<CallInst>(R);
+        if (CallL || CallR) {
+            auto *CalledL = CallL ? getCalledFunction(CallL->getCalledValue())
+                                  : nullptr;
+            auto *CalledR = CallR ? getCalledFunction(CallR->getCalledValue())
+                                  : nullptr;
+            if (!(CalledL && CalledR)) {
+                // If just one of the instructions is a call, it is possible
+                // that some logic has been moved into a function. We'll try to
+                // inline that function and compare again.
+                if (CalledL && !isSimpllAbstractionDeclaration(CalledL))
+                    ModComparator->tryInline = {CallL, nullptr};
+                else if (CalledR && !isSimpllAbstractionDeclaration(CalledR))
+                    ModComparator->tryInline = {nullptr, CallR};
+            }
+        }
         return Res;
+    }
     if (needToCmpOperands) {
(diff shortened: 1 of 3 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -847,31 +708,20 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
...
                 return Res;
             }
         }
 
+        if (isa<CallInst>(&*InstL) && isa<CallInst>(&*InstR)) {
+            // Try to find a type difference in field accesses.
+            // This has to be done even if the operations are equal, since field
+            // accesses return a pointer and all pointers are compared as equal
+            // in FunctionComparator.
+            findTypeDifferenceInFieldAccess(dyn_cast<CallInst>(&*InstL),
```
lenticularis39, line 721: I think it should be clearly stated here that this handles chained field abstractions (both in the comment and in the function name) to distinguish it from the case handled in `processCallInstDifference` by calling `findTypeDifferences` directly (which also looks for "a type difference in field accesses").

viktormalik:
@lenticularis39 I added a new commit, please review it.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1186,8 +1186,24 @@ int DifferentialFunctionComparator::cmpOperationsWithOperands(
     // convenient to have the code in a separate function.
 
     bool needToCmpOperands = true;
-    if (int Res = cmpOperations(L, R, needToCmpOperands))
+    if (int Res = cmpOperations(L, R, needToCmpOperands)) {
+        auto *CallL = dyn_cast<CallInst>(L);
+        auto *CallR = dyn_cast<CallInst>(R);
+        if (CallL != CallR) {
```
lenticularis39, line 1192: While `!=` is logical non-equivalence (XOR) on bool operands, it behaves differently on non-bool operands in case they are two different non-zero objects (logical non-equivalence would return false, `!=` returns true).

Because of this, the code inside the if condition will execute even if both `L` and `R` are call instructions, since the pointers will always be different, which I suppose isn't the desired behavior.

`!!CallL != !!CallR`, `bool(CallL) != bool(CallR)` or simply `(CallL || CallR) && !(CallL && CallR)` should work.

viktormalik: Correct, thanks for noting this.

viktormalik: Fixed.

PR #172, opened 2020-04-30
Slicing w.r.t. to specific value of module params.

nikopatrik:


diffkemp/simpll/passes/VarValueDependencySlicer.h
```diff
@@ -0,0 +1,40 @@
...
+        : public PassInfoMixin<VarValueDependencySlicer> {
+  public:
+    PreservedAnalyses run(Module &Mod,
+                          ModuleAnalysisManager &mam,
+                          GlobalVariable *Var,
+                          std::string VarValue);
+
+  private:
+    Constant *getConstantFromString(Constant *initializer,
+                                    std::string VarValue);
```
viktormalik, line 35: This function should go to `Utils.h`

diffkemp/simpll/passes/VarValueDependencySlicer.cpp
```diff
@@ -0,0 +1,123 @@
...
+PreservedAnalyses VarValueDependencySlicer::run(Module &Mod,
+                                                ModuleAnalysisManager &mam,
+                                                GlobalVariable *Var,
+                                                std::string VarValue) {
+    Constant *initializer, *newConst;
+    SmallVector<Instruction *, 128> *listOfLoadInstructions =
+            new SmallVector<Instruction *, 128>();
+    SmallVector<std::pair<Instruction *, int>, 128>
+            *listOfLoadInstructionUsers =
+                    new SmallVector<std::pair<Instruction *, int>, 128>();
```
viktormalik, line 40: Several remarks:
1. Use local `std::vector` instead of dynamically allocating `SmallVector`.
2. Do we even need those vectors? Can't we just iterate all load users of the global variable and for each user directly iterate all its users and directly replace those (without using any intermediary vectors)? Especially getting rid of the second vector would simplify the code a lot - no need to have the `GetIndexOfGlobalVariableOperand` method.

diffkemp/simpll/passes/VarValueDependencySlicer.cpp
```diff
@@ -0,0 +1,123 @@
...
+                    int index = GetIndexOfGlobalVariableOperand(instrOfGv,
+                                                                loadInst);
+                    listOfLoadInstructionUsers->push_back(
+                            std::make_pair(instrOfGv, index));
+                }
+            }
+        }
+    }
+
+    while (!listOfLoadInstructionUsers->empty()) {
```
viktormalik, line 70: Prefer using range-based `for` instead of `while` loops.

diffkemp/simpll/passes/VarValueDependencySlicer.cpp
```diff
@@ -0,0 +1,123 @@
...
+        std::pair<Instruction *, int> pair =
+                listOfLoadInstructionUsers->pop_back_val();
+        Instruction *instruction = pair.first;
+        int index = pair.second;
+
+        instruction->setOperand(index, newConst);
+    }
+
+    while (!listOfLoadInstructions->empty()) {
```
viktormalik, line 79: Same as above.

diffkemp/simpll/passes/VarValueDependencySlicer.cpp
```diff
@@ -0,0 +1,123 @@
...
+        int valtmp;
+        try {
+            valtmp = std::stoi(Value);
+        } catch (const std::invalid_argument &ia) {
+            exit(10);
+        }
+        val = valtmp;
+        ConstantInt *CI2 = ConstantInt::get(CI->getType(), val, true);
+
+        return CI2;
```
viktormalik, line 100: Directly return `ConstantInt::get(CI->getType(), val, true)`, there's no need to use a temporary variable.

diffkemp/simpll/passes/VarValueDependencySlicer.cpp
```diff
@@ -0,0 +1,123 @@
...
+    if (const ConstantFP *CFP = dyn_cast<ConstantFP>(initializer)) {
+        double val;
+        try {
+            val = std::stod(Value);
+        } catch (const std::invalid_argument &ia) {
+            exit(10);
+        }
+        Constant *CFP2 = ConstantFP::get(CFP->getType(), val);
+
+        return CFP2;
```
viktormalik, line 112: Same as above.

diffkemp/simpll/passes/VarValueDependencySlicer.cpp
```diff
@@ -0,0 +1,123 @@
...
+    else
+        return PreservedAnalyses::all();
+
+    if (VarValue != "default") {
+        newConst = getConstantFromString(initializer, VarValue);
+
+        if (newConst == NULL)
+            return PreservedAnalyses::none();
+    } else
+        newConst = initializer;
```
viktormalik, line 53: I think that it would make more sense to move this (getting appropriate `Constant` from a `string`) to parsing of CLI arguments and that this pass should accept `Constant *` instead of `std::string` for parameter `VarValue`.

diffkemp/snapshot.py
```diff
@@ -205,15 +210,21 @@ def _from_yaml(self, yaml_file):
...
+                                self.snapshot_source.kernel_dir
+                            ),
+                            f["llvm"]
+                        ))
+                    if f["llvm"] else None,
+                    f["glob_var"],
+                    f.get("glob_var_value"),
+                    f["tag"],
+                    group
+                )
```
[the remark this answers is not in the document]
viktormalik: This comment still applies.

diffkemp/diffkemp.py
```diff
@@ -117,6 +120,8 @@ def generate(args):
     # Build sources for symbols from the list into LLVM IR
     with open(args.functions_list, "r") as fun_list_file:
         for line in fun_list_file.readlines():
+            # enable comments inside functions list
+            line = line[:line.find('#')]
```
viktormalik, line 124: Use `"` for string literals instead of `'` (many occurrences in other places, too).

diffkemp/diffkemp.py
```diff
@@ -188,6 +193,43 @@ def generate(args):
...
+                kernel_param = data_mod.find_param_var(param)
+                for data_fun in \
+                        data_mod.get_functions_using_param(kernel_param):
+                    snapshot.add_fun(
+                        name=data_fun,
+                        llvm_mod=data_mod,
+                        glob_var=kernel_param.name,
+                        glob_var_value=value,
+                        tag="function using module param variable \"{}\""
+                        .format(kernel_param.name)
```
viktormalik, line 223: Indent this line by 4 spaces.

diffkemp/simpll/Utils.cpp
```diff
@@ -686,6 +686,34 @@ std::string valueToString(const Value *Val) {
     return DumpStrm.str();
 }
 
+Constant *getConstantFromString(Constant *initializer, std::string Val) {
```
viktormalik, line 689: Why does this function use initializer? Just to get the variable type? In such case I think that it would be better to pass the type directly.

diffkemp/simpll/Config.cpp
```diff
@@ -93,8 +94,27 @@ Config::Config()
...
+            varValue = vopt.substr(varOptDelimiterIndex + 1, vopt.size());
+        }
+
+        FirstVar = First->getGlobalVariable(varName, true);
+        SecondVar = Second->getGlobalVariable(varName, true);
+
+        if (vopt == varName) {
+            VarValue = nullptr;
+        } else if (varValue.empty() || varValue == "default")
+            UseDefaultValue = true;
```
viktormalik, line 113: Wouldn't it be easier to directly pass the initializer to `VarValue`? It's true that you would need to have `VarValueFirst` and `VarValueSecond` but it seems somehow cleaner to me (and you could get rid of a block of code in `VarValueDependencySlicer`).

diffkemp/diffkemp.py
```diff
@@ -179,15 +190,55 @@ def generate(args):
...
+                            with_value_string = " with {} as value".format(value) \
+                                if value else ''
                             print(
-                                "  {}: {} (using data variable \"{}\")".format(
+                                "  {}: {} (using data variable \"{}\"{})".format(
                                     data_fun,
                                     os.path.relpath(data_mod.llvm,
                                                     args.kernel_dir),
-                                    data.name))
+                                    data.name, with_value_string))
```
viktormalik, line 204:
```suggestion
                                    data.name, 
                                    with_value_string))
```

diffkemp/diffkemp.py
```diff
@@ -179,15 +190,55 @@ def generate(args):
...
+                # For kernel module parameters, we compile the module,
+                # collect the list of functions using the parameter variable,
+                # and include it into the snapshot.
+                # Format of the entries in the function list is:
+                #    <module directory>/<module name>:<parameter> [ = <value> ]
+                line = [i.strip() for i in line.split(':')]
+                mod = line[0].split('/')
+                mod_dir = '/'.join(mod[:-1])
+                mod_name = mod[-1]
+                param_with_value = [i.strip() for i in line[1].split('=')]
```
viktormalik, line 215: Use `"` everywhere.

diffkemp/semdiff/function_diff.py
```diff
@@ -159,7 +159,7 @@ def functions_semdiff(first, second, fun_first, fun_second, config):
 
 def functions_diff(mod_first, mod_second,
                    fun_first, fun_second,
-                   glob_var, config,
+                   glob_var, config, glob_var_value=None,
```
viktormalik, line 162:
```suggestion
                   config,
                   glob_var, glob_var_value=None,
```
This params order would make more sense.

viktormalik (changes requested):

viktormalik:
This is stale and probably has too many conflicts, so will have to be eventually reworked from scratch.

PR #173, opened 2020-05-07
Better field access comparison and type difference finding

lenticularis39:
The moving of field access operations to separate functions (field access abstractions) was disabled. `cmpFieldAccess` was rewritten to compare the field accesses in place with the help of a new function `isConstantMemoryAccessToPtr`, moving the instruction iterators to their end if the operations are equal, causing them to be skipped in `cmpBasicBlocks`.

The code for type difference finding was modified to operate on the first instruction of the field access operation instead of calls to field access abstractions, iterating over the operation with the help of a new function `isFollowingFieldAccessInstruction` that is similar to `isConstantMemoryAccessToPtr`. Another new function called `getFieldAccessStart` is used to get the first instruction from an arbitrary one (typically the one where the difference is found).

A new file `FieldAccessUtils.cpp` was created for the new functions and `getFieldAccessSourceTypes`. Unit tests were written for the functions in `FieldAccessUtils.cpp` and for `cmpFieldAccess`.


lenticularis39:
I discovered that the comparison using the new `cmpFieldAccess` loops in certain cases (for example on `__alloc_pages_nodemask` between kernel versions `3.10.0-862.el7` and `3.10.0-957.el7`). This should be fixed before merging.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -904,39 +828,139 @@ bool DifferentialFunctionComparator::accumulateAllOffsets(
     return true;
 }
 
-/// Specific comparing of structure field access.
-int DifferentialFunctionComparator::cmpFieldAccess(const Function *L,
-                                                   const Function *R) const {
-    // First compute the complete offset of all GEPs in both functions.
+/// Finds all differences between source types in GEPs inside two field access
+/// abstractions and records them using findTypeDifference.
```
viktormalik, line 832: We do not use "abstractions" anymore.

viktormalik:
Shouldn't this PR completely remove the `FieldAccessFunctionGenerator` class?

lenticularis39:
I didn't want to remove it because of the possibility of it being useful at some point in the future, of course the question is how likely this is to happen.

lenticularis39:
Now I think it probably should be removed, because `DifferentialFunctionComparator` no longer supports field access abstraction.

viktormalik:
I agree, let's remove it.

PR #175, opened 2020-05-29
SimpLL: Ignore repetitive load instructions

### LINKED ISSUE #147: Handle repetitive load instructions without a store in between
If one of the compared modules contains an extra load instruction and another load from the same pointer occurred before (without a store to the same pointer between the two loads), allow to skip the extra load and use the first one instead.

This enhancement should remove the false positive difference reported in function `check_msg_timeout` that can be reproduced by running:

    bin/diffkemp compare snapshots/linux-3.10.0-327.el7/ snapshots/linux-3.10.0-514.el7/ --stdout --show-diff -f ipmi_register_smi

PSilling:
When using a load instruction during ignoration analysis, check whether the load instruction is repetitive. These are loads that use the same pointer as a previous load instruction that is reachable for all possible predecessor flows with no store in between. Pointer aliasing is not taken into account and may be added in the future if necessary.

The set of ignored instructions is also changed to an unordered map which holds references to replacement values. During value comparison, replacement values are used to avoid reanalysis.

Fixes #147


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -692,16 +789,20 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
...
                 sn_mapR.erase(&*InstR);
                 // One of the compared operations will be skipped and the
                 // comparison will be repeated.
-                if (mayIgnore(&*InstL))
+                if (ignoredL) {
                     InstL++;
-                else
+                } else {
                     InstR++;
+                }
```
viktormalik, line 805: Maybe a more readable style would be:

    if (mayIgnore(&*InstL, true)) {
        sn_mapL.erase(&*InstL);
        sn_mapR.erase(&*InstR);
        InstL++;
        continue;
    }
    if (mayIgnore(&*InstR, true)) {
        sn_mapL.erase(&*InstL);
        sn_mapR.erase(&*InstR);
        InstR++;
        continue;
    }

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -543,29 +552,117 @@ bool DifferentialFunctionComparator::mayIgnore(const User *Inst) const {
...
+    }
+    return canBeIgnored;
+}
+
+/// Check whether the given instruction is a repetitive variant of a previous
+/// load with no store instructions in between. Will return the first load
+/// if the replacement pointer is given.
+bool DifferentialFunctionComparator::mayIgnoreLoad(const LoadInst *Load,
+                                                   bool addToIgnored) const {
+    if (auto BB = Load->getParent()) {
```
viktormalik, line 591: Use inverse condition style:

    auto *BB = Load->getParent();
    if (!BB)
        return false;
    ....

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -543,29 +552,117 @@ bool DifferentialFunctionComparator::mayIgnore(const User *Inst) const {
...
+        // Check if all predecessor blocks have a path to a load
+        // from the same pointer with no stores to the same pointer
+        // in between.
+        bool first = true;
+        const LoadInst *PreviousLoad = nullptr;
+        std::deque<const BasicBlock *> blockQueue = {BB};
+        std::vector<const BasicBlock *> visitedBlocks;
```
viktormalik, line 598: Use `std::set`, the search will be faster. Also, it allows you to use the `find` method directly on the set (instead of `std::find`).

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -543,29 +552,117 @@ bool DifferentialFunctionComparator::mayIgnore(const User *Inst) const {
...
+                                   == visitedBlocks.end()) {
+                        blockQueue.push_back(PredBB);
                     }
                 }
-                return true;
             }
         }
-        return false;
+        // If the load is repeated without stores in between, skip
+        // it because it creates no semantic differences.
```
viktormalik, line 657: better comment: `... because it is redundant and its removal will cause no semantic difference.`

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -543,29 +552,117 @@ bool DifferentialFunctionComparator::mayIgnore(const User *Inst) const {
...
+}
+
+/// Check whether the given instruction is a repetitive variant of a previous
+/// load with no store instructions in between. Will return the first load
+/// if the replacement pointer is given.
+bool DifferentialFunctionComparator::mayIgnoreLoad(const LoadInst *Load,
+                                                   bool addToIgnored) const {
+    if (auto BB = Load->getParent()) {
+        // Check if all predecessor blocks have a path to a load
+        // from the same pointer with no stores to the same pointer
```
viktormalik, line 593: In fact, you're not checking for stores to the *same* pointer but to *any* pointer since checking the same pointer is not sound because of pointer aliasing. The comment should state that.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -543,29 +552,117 @@ bool DifferentialFunctionComparator::mayIgnore(const User *Inst) const {
...
+                    PreviousLoad = nullptr;
+                    searchPredecessors = false;
+                    blockQueue.clear();
+                    break;
+                }
+            }
+
+            if (searchPredecessors) {
+                auto BBPredecessors = predecessors(BB);
+                if (BBPredecessors.begin() == BBPredecessors.end()) {
```
viktormalik, line 637: You can use `BBPredecessors.empty()`.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -505,34 +507,41 @@ int DifferentialFunctionComparator::cmpAllocs(const CallInst *CL,
...
 /// Check if the given operation can be ignored (it does not affect semantics)
 /// for control flow only diffs.
-bool DifferentialFunctionComparator::mayIgnore(const User *Inst) const {
-    if (config.ControlFlowOnly)
-        return isa<AllocaInst>(Inst) || isCast(Inst);
-    else {
-        if (isa<AllocaInst>(Inst))
-            return true;
+bool DifferentialFunctionComparator::mayIgnore(const User *Inst,
+                                               bool addToIgnored) const {
```
viktormalik, line 511: Refactor this to `bool maySkipInstruction(const Instruction *Inst)`. The function should always insert into the `IgnoredInstructions` map if it returns true.

diffkemp/simpll/DifferentialFunctionComparator.cpp
viktormalik: Replace `mayIgnore` by `maySkipInstruction` that is proposed above.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -947,25 +1048,17 @@ int DifferentialFunctionComparator::cmpFieldAccess(const Function *L,
...
-        // The right value is a cast - use the original value of it in the
-        // comparison
-        return cmpValues(L, CR->getOperand(0));
+    // Use replacement references for ignored values.
+    auto ignoredL = ignoredInstructions.find(L);
+    auto ignoredR = ignoredInstructions.find(R);
+    auto replaceL = (ignoredL == ignoredInstructions.end()) ? nullptr
+                                                            : ignoredL->second;
+    auto replaceR = (ignoredR == ignoredInstructions.end()) ? nullptr
+                                                            : ignoredR->second;
```
viktormalik, line 1057: Implement a new method `Value *getReplacementValue(const Value *)` that returns the `IgnoredInstructions` map entry if it exists for the given parameter and call it from here.
Also, if the given Value is not an instruction, the `getReplacementValue` method should check if it is a skippable bitcast operator and if so, return its first operand.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1214,6 +1307,14 @@ int DifferentialFunctionComparator::cmpOperationsWithOperands(
             Value *OpR = R->getOperand(i);
 
             if (int Res = cmpValues(OpL, OpR)) {
+                auto *UserL = dyn_cast<User>(OpL);
```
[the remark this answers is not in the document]
viktormalik: Actually, if the new function `getReplacementValue` is implemented as proposed above, this block of code will become unnecessary since constant expressions will be handled by the `cmpValues` call.

diffkemp/simpll/DifferentialFunctionComparator.cpp
PSilling: Unfortunately, this is not possible. The iterator returned by `predecessors` implements only `begin` and `end`.

PSilling:
The proposed changes should now be implemented and ready for review.

diffkemp/simpll/DifferentialFunctionComparator.cpp
lenticularis39: Just to clarify this: `iterator_range::empty()` was added in LLVM 10, therefore it can't be used because of DiffKemp supporting lower versions of LLVM.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -503,75 +506,190 @@ int DifferentialFunctionComparator::cmpAllocs(const CallInst *CL,
...
 
-/// Check if the given operation can be ignored (it does not affect semantics)
-/// for control flow only diffs.
-bool DifferentialFunctionComparator::mayIgnore(const User *Inst) const {
-    if (config.ControlFlowOnly)
-        return isa<AllocaInst>(Inst) || isCast(Inst);
-    else {
-        if (isa<AllocaInst>(Inst))
+/// Check if the given instruction can be ignored (it does not affect
+/// semantics) for control flow only diffs. Replacements of ignorable
```
viktormalik, line 510: Remove the "control flow only diffs" part. It is legacy and not correct.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -503,75 +506,190 @@ int DifferentialFunctionComparator::cmpAllocs(const CallInst *CL,
...
-                    // comparator would detect a difference when the value is
-                    // used, since it would have a different type.
-                    return true;
-            }
+            return maySkipCast(Inst);
+        }
+        if (auto Load = dyn_cast<LoadInst>(Inst)) {
+            return maySkipLoad(Load);
+        }
+    }
```
viktormalik, line 530: I think that a better style of this would be:

    if (isCast(Inst)) {
        if (config.ControlFlowOnly)
            ...
        else
            ...
    }
    if (auto Load = ...) {
        ...
    }

to have all the top-level if conditions checking the instruction kind.

diffkemp/simpll/DifferentialFunctionComparator.h
```diff
@@ -151,9 +153,24 @@ class DifferentialFunctionComparator : public FunctionComparator {
     /// constant (if not, it returns false).
     bool accumulateAllOffsets(const BasicBlock &BB, uint64_t &Offset) const;
 
-    /// Check if the given operation can be ignored (it does not affect
-    /// semantics) for control flow only diffs.
-    bool mayIgnore(const User *Inst) const;
+    /// Check if the given instruction can be ignored (it does not affect
+    /// semantics) for control flow only diffs. Replacements of ignorable
```
viktormalik, line 157: Remove the part on "control flow only diffs".

PR #177, opened 2020-07-09
Difference in the number of elements of two enums

### LINKED ISSUE #146: Ignore difference in array size when index is an enum
If two GEPs for array element access are compared and the only difference is the number of elements of the accessed array, do not consider this a semantic change if the index has an enumeration type. 

In this situation, it is very likely that the enumeration type has additional elements in one of the versions, however, the element at the same index has the same semantics.

Resolving this issue should fix a false positive that occurs when running:

    bin/diffkemp compare snapshots/linux-4.18.0-80.el8 snapshots/linux-4.18.0-147.el8 -f si_meminfo --show-diff --stdout

tmalecova:
During comparison of two GEPs for array element access is inequality
only in the number of elements of the accessed array, its not considered
anymore as a semantic change if the index has an enumeration type.
This closes #146.

### DIFF
--- a/diffkemp/simpll/DebugInfo.cpp
+++ b/diffkemp/simpll/DebugInfo.cpp
@@ -479,3 +479,42 @@ void DebugInfo::removeFunctionsDebugInfo(Module &Mod) {
     for (auto &F : Mod)
         fpm.run(F, fam);
 }
+
+DICompositeType *getVariableTypeInfo(Value *Val) {
+    if (!Val->getType()->isIntegerTy())
+        return nullptr;
+
+    if (isa<Argument>(Val) || isa<AllocaInst>(Val)) {
+        // Val must be a parameter or a local variable so that we are able to
+        // extract its debug info
+        auto Fun =
+                isa<Argument>(Val)
+                        ? dyn_cast<Argument>(Val)->getParent()
+                        : dyn_cast<AllocaInst>(Val)->getParent()->getParent();
+
+        for (auto &Inst : instructions(Fun)) {
+            auto DbgInst = dyn_cast<DbgValueInst>(&Inst);
+            if (!DbgInst)
+                continue;
+
+            auto ValMD = dyn_cast<MetadataAsValue>(DbgInst->getOperand(0))
+                                 ->getMetadata();
+            auto DbgVal = dyn_cast<ValueAsMetadata>(ValMD)->getValue();
+
+            if (DbgVal != Val)
+                continue;
+
+            auto VarMD = dyn_cast<MetadataAsValue>(DbgInst->getOperand(1))
+                                 ->getMetadata();
+            auto LocVar = dyn_cast<DILocalVariable>(VarMD);
+
+            return dyn_cast<DICompositeType>(LocVar->getType());
+        }
+    }
+
+    auto Instr = dyn_cast<Instruction>(Val);
+    if (Instr && isCast(Instr))
+        return getVariableTypeInfo(Instr->getOperand(0));
+
+    return nullptr;
+}
--- a/diffkemp/simpll/DebugInfo.h
+++ b/diffkemp/simpll/DebugInfo.h
@@ -18,6 +18,7 @@
 #include "Utils.h"
 #include <llvm/IR/DebugInfo.h>
 #include <llvm/IR/Instructions.h>
+#include <llvm/IR/IntrinsicInst.h>
 #include <llvm/IR/PassManager.h>
 #include <llvm/IR/TypeFinder.h>
 #include <llvm/IR/ValueMap.h>
@@ -52,10 +53,6 @@ class DebugInfo {
         calculateMacroAlignments();
         collectLocalVariables(CalledFirst, LocalVariableMapL);
         collectLocalVariables(CalledSecond, LocalVariableMapR);
-        // Remove calls to debug info intrinsics from the functions - it may
-        // cause some non-equalities in FunctionComparator.
-        removeFunctionsDebugInfo(modFirst);
-        removeFunctionsDebugInfo(modSecond);
     };
 
     /// Maps structure type and index to struct member names
@@ -162,5 +159,10 @@ bool isDebugInfo(const Instruction &Instr);
 
 /// Get name of a struct type as it is specified in the C source.
 std::string getStructTypeName(const StructType *type);
+
+/// Get type of given Value
+/// \param Val Value of variable.
+/// \return Variable type debug info.
+DICompositeType *getVariableTypeInfo(Value *Val);
 
 #endif // DIFFKEMP_SIMPLL_DEBUGINFO_H
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -58,6 +58,43 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
         // The original function says the GEPs are equal - return the value
         return OriginalResult;
 
+    if (isa<ArrayType>(GEPL->getSourceElementType())
+        && isa<ArrayType>(GEPR->getSourceElementType())) {
+        // When the only difference is the size of the accessed array,
+        // it is not considered a semantic change if the index type is an enum
+
+        if (GEPL->getNumOperands() != 3 || GEPR->getNumOperands() != 3)
+            // We only handle GEPs that access exactly one array element
+            return OriginalResult;
+
+        if (GEPL->getSourceElementType()->getArrayNumElements()
+            == GEPR->getSourceElementType()->getArrayNumElements())
+            return OriginalResult;
+
+        auto *STyL = cast<SequentialType>(GEPL->getSourceElementType());
+        auto *STyR = cast<SequentialType>(GEPR->getSourceElementType());
+        if (int Res = cmpTypes(STyL->getElementType(), STyR->getElementType()))
+            // The array element type must be the same
+            return Res;
+
+        for (unsigned i = 0, e = GEPL->getNumOperands(); i != e; ++i) {
+            if (int Res = cmpValues(GEPL->getOperand(i), GEPR->getOperand(i)))
+                return Res;
+        }
+
+        auto TypeL = getVariableTypeInfo(GEPL->getOperand(2));
+        auto TypeR = getVariableTypeInfo(GEPR->getOperand(2));
+
+        if (!TypeL || !TypeR)
+            return OriginalResult;
+
+        if (TypeL->getTag() == dwarf::DW_TAG_enumeration_type
+            && TypeR->getTag() == dwarf::DW_TAG_enumeration_type)
+            return 0;
+
+        return OriginalResult;
+    }
+
     if (!isa<StructType>(GEPL->getSourceElementType())
         || !isa<StructType>(GEPR->getSourceElementType()))
         // One of the types in not a structure - the original function is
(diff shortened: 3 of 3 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -58,6 +58,25 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
...
+
+        if (int Res =
+                    cmpNumbers(GEPL->getNumOperands(), GEPR->getNumOperands()))
+            return Res;
+
+        for (unsigned i = 0, e = GEPL->getNumOperands(); i != e; ++i) {
+            if (int Res = cmpValues(GEPL->getOperand(i), GEPR->getOperand(i)))
+                return Res;
+        }
+        return 0;
```
viktormalik, line 77: We cannot simply return 0 here, we need to check that the array types are enums.

viktormalik:
There is a failing unit test for some LLVM versions. It needs to be checked.

viktormalik:
3 regression tests are failing - these need to be checked

diffkemp/simpll/DebugInfo.h
```diff
@@ -52,10 +52,6 @@ class DebugInfo {
         calculateMacroAlignments();
         collectLocalVariables(CalledFirst, LocalVariableMapL);
         collectLocalVariables(CalledSecond, LocalVariableMapR);
-        // Remove calls to debug info intrinsics from the functions - it may
-        // cause some non-equalities in FunctionComparator.
```
viktormalik, line 56: Make sure that the errors mentioned in this comment do not occur - add skipping of comparison of `llvm.dbg.*` instructions into `DifferentialFunctionComparator::cmpBasicBlocks`.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -58,6 +59,101 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
...
+            if (dbgValL != indexL) {
+                InstL++;
+                continue;
+            }
+            if (dbgValR != indexR) {
+                InstR++;
+                continue;
+            }
+            break;
+        }
```
viktormalik, line 135: This is a weird loop, I'd prefer to have a separate function that gets a value and returns its metadata, or even its type. It could be called e.g. `getVariableMetadata` and it should be located in `DebugInfo`.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -58,6 +59,101 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
...
+        if (int Res =
+                    cmpNumbers(GEPL->getNumOperands(), GEPR->getNumOperands()))
+            return Res;
+
+        for (unsigned i = 0, e = GEPL->getNumOperands(); i != e; ++i) {
+            if (int Res = cmpValues(GEPL->getOperand(i), GEPR->getOperand(i)))
+                return Res;
+        }
+        if (GEPL->hasAllConstantIndices() || GEPR->hasAllConstantIndices())
+            return OriginalResult;
```
viktormalik, line 79: This is not necessary, I believe

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -58,6 +59,101 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
...
+
+        auto instrL = dyn_cast<Instruction>(GEPL->getOperand(2));
+        auto instrR = dyn_cast<Instruction>(GEPR->getOperand(2));
```
viktormalik, line 82: You should check that the GEPs have at least 3 instructions. Also, let's limit this case for GEPs that have exactly 3 instructions.

diffkemp/simpll/DebugInfo.h
tmalecova: Sorry, I forgot to add it. This caused the failure of regression tests. Already fixed.

diffkemp/simpll/DifferentialFunctionComparator.cpp
tmalecova: Implemented `getVariableTypeInfo` function in `DebugInfo`, which returns the type of given value.

diffkemp/simpll/DifferentialFunctionComparator.cpp
tmalecova: Deleted.

diffkemp/simpll/DifferentialFunctionComparator.cpp
tmalecova: Not sure, but it's done.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -58,6 +58,69 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
         // The original function says the GEPs are equal - return the value
         return OriginalResult;
 
+    if (isa<ArrayType>(GEPL->getSourceElementType())
+        && isa<ArrayType>(GEPR->getSourceElementType())) {
+        // The number of elements of two enums can be different and
+        // its not considered as semantic change
```
viktormalik, line 64: Better comment:
"When the only difference is the size of the accessed array, it is not considered a semantic change if the index type is an enum"

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -58,6 +58,69 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
...
+        if (GEPL->getSourceElementType()->getArrayNumElements()
+            == GEPR->getSourceElementType()->getArrayNumElements())
+            return OriginalResult;
+
+        auto *STyL = cast<SequentialType>(GEPL->getSourceElementType());
```
viktormalik, line 69: Add comment: "The array element type must be the same"

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -58,6 +58,69 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
...
+                    cmpNumbers(GEPL->getNumOperands(), GEPR->getNumOperands()))
+            return Res;
+
+        for (unsigned i = 0, e = GEPL->getNumOperands(); i != e; ++i) {
+            if (int Res = cmpValues(GEPL->getOperand(i), GEPR->getOperand(i)))
+                return Res;
+        }
+
+        if (GEPL->getNumOperands() < 3 || GEPR->getNumOperands() < 3)
+            return OriginalResult;
```
viktormalik, line 84:
```suggestion
        if (GEPL->getNumOperands() != 3 || GEPR->getNumOperands() != 3)
            return OriginalResult;
```
Add a comment (e.g. "We only handle GEPs that access exactly one array element") and move this upwards to the beginning of this block.

diffkemp/simpll/DebugInfo.cpp
```diff
@@ -479,3 +479,27 @@ void DebugInfo::removeFunctionsDebugInfo(Module &Mod) {
     for (auto &F : Mod)
         fpm.run(F, fam);
 }
+
+DICompositeType *getVariableTypeInfo(Value *Val, const Function *Fun) {
+    for (const_inst_iterator I = inst_begin(Fun), E = inst_end(Fun); I != E;
+         ++I) {
```
viktormalik, line 485:
```suggestion
    for (auto &Inst : instructions(Fun)) {
```

diffkemp/simpll/DebugInfo.cpp
```diff
@@ -479,3 +479,27 @@ void DebugInfo::removeFunctionsDebugInfo(Module &Mod) {
     for (auto &F : Mod)
         fpm.run(F, fam);
 }
+
+DICompositeType *getVariableTypeInfo(Value *Val, const Function *Fun) {
```
viktormalik, line 483: Add Doxygen comment for the function

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -58,6 +58,69 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
...
+
+        auto InstrL = dyn_cast<Instruction>(GEPL->getOperand(2));
+        auto InstrR = dyn_cast<Instruction>(GEPR->getOperand(2));
+
+        if (InstrL && !isa<ZExtInst>(InstrL) && !isa<SExtInst>(InstrL)
+            && !isa<TruncInst>(InstrL))
+            return OriginalResult;
```
viktormalik, line 91: What if `InstrL` is an alloca instruction (i.e., a local variable)? The method wouldn't work because of this check.
Maybe it would be better to move resolving of casts into the `getVariableTypeInfo`.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -58,6 +58,69 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
...
+
+        auto IndexL = InstrL ? InstrL->getOperand(0) : GEPL->getOperand(2);
+        auto IndexR = InstrR ? InstrR->getOperand(0) : GEPR->getOperand(2);
+
+        if (!IndexL->getType()->isIntegerTy()
+            || !IndexR->getType()->isIntegerTy())
+            return OriginalResult;
+
+        const GetElementPtrInst *GEPiL = dyn_cast<GetElementPtrInst>(GEPL);
+        const GetElementPtrInst *GEPiR = dyn_cast<GetElementPtrInst>(GEPR);
```
viktormalik, line 104: This seems unnecessary - casting `GetElementPtrInst` to `GetElementPtrInst`.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -58,6 +58,47 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
...
+
+        auto *STyL = cast<SequentialType>(GEPL->getSourceElementType());
+        auto *STyR = cast<SequentialType>(GEPR->getSourceElementType());
+        if (int Res = cmpTypes(STyL->getElementType(), STyR->getElementType()))
+            // The array element type must be the same
+            return Res;
+
+        if (int Res =
+                    cmpNumbers(GEPL->getNumOperands(), GEPR->getNumOperands()))
+            return Res;
```
viktormalik, line 82: Not necessary since these numbers are always equal to 3 here.

diffkemp/simpll/DebugInfo.cpp
```diff
@@ -479,3 +479,40 @@ void DebugInfo::removeFunctionsDebugInfo(Module &Mod) {
...
+
+            return dyn_cast<DICompositeType>(LocVar->getType());
+        }
+    }
+
+    auto Instr = dyn_cast<Instruction>(Val);
+    if (Instr && isCast(Instr))
+        getVariableTypeInfo(Instr->getOperand(0));
+    else
+        return nullptr;
```
viktormalik, line 517:
```suggestion
    if (Instr && isCast(Instr))
        return getVariableTypeInfo(Instr->getOperand(0));
    
    return nullptr;
```

diffkemp/simpll/DebugInfo.cpp
```diff
@@ -479,3 +479,40 @@ void DebugInfo::removeFunctionsDebugInfo(Module &Mod) {
     for (auto &F : Mod)
         fpm.run(F, fam);
 }
+
+DICompositeType *getVariableTypeInfo(Value *Val) {
+    if (!Val->getType()->isIntegerTy())
+        return nullptr;
+
+    if (isa<Argument>(Val) || isa<AllocaInst>(Val)) {
```
viktormalik, line 487: Add comment: "Val must be a parameter or a local variable so that we are able to extract its debug info"

viktormalik (changes requested):

PR #179, opened 2020-10-17
SimpLL: Pointers to first differing instructions

tmalecova:
Added pair of const Instructions* to DifferentialFunctionComparator
where are stored first differing instructions of nonequal functions.
Also implemented function DifferentialFunctionComparator::equal that
compares two given instructions and decide whether they are equal
(returns true) or not (returns false).

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -41,6 +41,8 @@ int DifferentialFunctionComparator::compare() {
         for (auto &PhiPair : phisToCompare)
             if (cmpPHIs(PhiPair.first, PhiPair.second))
                 return 1;
+        // Functions are equal so we don't have differing instructions
+        DifferingInstructions = {nullptr, nullptr};
         return 0;
     }
     return Res;
@@ -198,6 +200,9 @@ int DifferentialFunctionComparator::cmpOperations(
         const Instruction *L,
         const Instruction *R,
         bool &needToCmpOperands) const {
+    // Need to store comparing instructions in DifferingInstructions pair
+    DifferingInstructions = {L, R};
+
     int Result = FunctionComparator::cmpOperations(L, R, needToCmpOperands);
 
     // Check whether the instruction is a call instruction.
@@ -1430,4 +1435,21 @@ void DifferentialFunctionComparator::findDifference(
             findTypeDifferences(FAL, FAR, L->getFunction(), R->getFunction());
         }
     }
+}
+
+bool DifferentialFunctionComparator::equal(const Instruction *InstL,
+                                           const Instruction *InstR) {
+    auto L = sn_mapL.find(InstL);
+    auto R = sn_mapR.find(InstR);
+
+    if (L == sn_mapL.end() || R == sn_mapR.end())
+        // Values are not in maps
+        return false;
+
+    if (std::make_pair(InstL, InstR) == DifferingInstructions)
+        // The instructions are the last instructions that were compared as
+        // non-equal.
+        return false;
+
+    return L->second == R->second;
 }
--- a/diffkemp/simpll/DifferentialFunctionComparator.h
+++ b/diffkemp/simpll/DifferentialFunctionComparator.h
@@ -39,6 +39,11 @@ class DifferentialFunctionComparator : public FunctionComparator {
               LayoutR(F2->getParent()->getDataLayout()), ModComparator(MC) {}
 
     int compare() override;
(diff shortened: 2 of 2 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1431,3 +1436,30 @@ void DifferentialFunctionComparator::findDifference(
...
     }
 }
+
+bool DifferentialFunctionComparator::equal(Instruction *InstL,
+                                           Instruction *InstR) {
+    const auto *ValL = dyn_cast<Value>(InstL);
+    const auto *ValR = dyn_cast<Value>(InstR);
+
+    if (!ValL || !ValR)
+        return false;
```
viktormalik, line 1446: This shouldn't be necessary (`Instruction` inherits from `Value` and so `Instruction *` can be used directly where `Value *` is expected).

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1431,3 +1436,30 @@ void DifferentialFunctionComparator::findDifference(
...
+    DenseMapIterator<const Value *, int> R = sn_mapR.find(ValR);
+
+    if (L == sn_mapL.end() && R == sn_mapR.end())
+        // Values are not in maps
+        return false;
+
+    if (L->second == sn_mapL.size() - 1 && R->second == sn_mapR.size() - 1)
+        // The last inserted values with the same serial number
+        // are the differing ones
+        return false;
```
viktormalik, line 1458: I think that this can be replaced by checking if the pair `{InstL, InstR}` is equal to `DifferingInstructions`. This would be also much more readable.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1431,3 +1436,30 @@ void DifferentialFunctionComparator::findDifference(
...
+
+    if (L->second != R->second)
+        // Check whether values in maps have different serial number
+        return false;
+
+    return true;
```
viktormalik, line 1464:
```suggestion
    return L->second == R->second;
```

viktormalik (changes requested):
Some minor modification suggestions that would improve the code:

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1431,3 +1436,20 @@ void DifferentialFunctionComparator::findDifference(
...
+    auto L = sn_mapL.find(InstL);
+    auto R = sn_mapR.find(InstR);
+
+    if (L == sn_mapL.end() || R == sn_mapR.end())
+        // Values are not in maps
+        return false;
+
+    if (std::make_pair(InstL, InstR) == DifferingInstructions)
+        // Comparing instructions with the same serial number are the differing
+        // ones
```
viktormalik, line 1451: Better comment: the instructions are the last instructions that were compared as non-equal.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1431,3 +1436,20 @@ void DifferentialFunctionComparator::findDifference(
...
+bool DifferentialFunctionComparator::equal(const Instruction *InstL,
+                                           const Instruction *InstR) {
+    auto L = sn_mapL.find(InstL);
+    auto R = sn_mapR.find(InstR);
+
+    if (L == sn_mapL.end() || R == sn_mapR.end())
+        // Values are not in maps
+        return false;
+
+    if (std::make_pair(InstL, InstR) == DifferingInstructions)
```
viktormalik, line 1449:
```suggestion
    if (DifferingInstructions == {InstL, InstR})
```

viktormalik (changes requested):

diffkemp/simpll/DifferentialFunctionComparator.cpp
tmalecova: Not working.

PSilling: Yes, I believe that this might be invalid. I don't think you can compare directly against an `std::initializer_list`.

lenticularis39: clang for example explicitly reports this:
```
$ clang test.cpp -o test
test.cpp:5:16: error: initializer list cannot be used on the right hand side of operator '=='
  return (test == {1, 2});
               ^  ~~~~~~
1 error generated.
```

viktormalik (approved):

PR #180, opened 2020-11-10
Add colored compilation with Clang support

PSilling:
Includes three small changes concerning the compilation process:

- Add colour diagnostic flags to CMakeLists for both GCC and Clang since a pseudo-terminal gets used with Ninja, which disables output colour coding by default.
- Fix warnings reported ba Clang. These include the removal of template nesting in `Output.cpp` and an unsigned integer subtraction fix in `DifferentialFunctionComparator.cpp`.
- Add Clang compilation into the CI configuration (for the latest version of LLVM only).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,13 +21,15 @@ jobs:
         - flake8 diffkemp --exclude=llreve --ignore=F403,F405,W504
         - flake8 tests
         - tools/check-clang-format.sh -d
-    - env: LLVM_VERSION=5.0
+    - env: LLVM_VERSION=5.0 CC=/usr/bin/gcc CXX=/usr/bin/g++
       if: type = cron
-    - env: LLVM_VERSION=6.0
-    - env: LLVM_VERSION=7
-    - env: LLVM_VERSION=8
-    - env: LLVM_VERSION=9
-    - env: LLVM_VERSION=10
+    - env: LLVM_VERSION=6.0 CC=/usr/bin/gcc CXX=/usr/bin/g++
+      if: type = cron
+    - env: LLVM_VERSION=7 CC=/usr/bin/gcc CXX=/usr/bin/g++
+    - env: LLVM_VERSION=8 CC=/usr/bin/gcc CXX=/usr/bin/g++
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -21,13 +21,14 @@ jobs:
...
         - tools/check-clang-format.sh -d
-    - env: LLVM_VERSION=5.0
+    - env: LLVM_VERSION=5.0 CC=/usr/bin/gcc CXX=/usr/bin/g++
       if: type = cron
-    - env: LLVM_VERSION=6.0
-    - env: LLVM_VERSION=7
-    - env: LLVM_VERSION=8
-    - env: LLVM_VERSION=9
-    - env: LLVM_VERSION=10
+    - env: LLVM_VERSION=6.0 CC=/usr/bin/gcc CXX=/usr/bin/g++
```
viktormalik, line 26:
```suggestion
    - env: LLVM_VERSION=6.0 CC=/usr/bin/gcc CXX=/usr/bin/g++
      if: type = cron
```
Travis runs 5 parallel jobs, so we keep the 5 most recent versions in normal builds (PR/branch) and the rest is run in Cron builds only.

viktormalik (changes requested):

.travis.yml
PSilling: Wouldn't it be better to make the LLVM 10 GCC build cron-only then? In reality, the compiler change should create no meaningful difference. I would say therefore say that it is more important to run tests on multiple LLVM versions (especially since there were quite a few problems caused by LLVM version compatibility already).

viktormalik: I don't think that this actually matters. LLVM 5 versions back is already quite old, so supporting it is not crucial. Truth is that changing a compiler should not make any difference, especially to the tests, however if it ever does, we'll spot it quickly if we run the same version with both GCC and Clang. And cron is run every day, so regressions will be seen quickly anyway.

PSilling: Alright, I have set LLVM 6.0 to cron-only.

viktormalik (approved):

PR #182, opened 2020-12-09
SimpLL: Equivalence slicer

tmalecova:
This pull request contains implementation of static slicer, that will be able to remove semantically equal and independent instructions of compared functions.

- [x] Remove prefix when the difference is in the first basic block.


diffkemp/simpll/ControlFlowGraphUtils.h
```diff
@@ -0,0 +1,106 @@
...
+#include "Utils.h"
+#include <llvm/IR/Module.h>
+#include <queue>
+#include <set>
+
+using namespace llvm;
+
+class ControlFlowSlicer {
+  public:
+    GlobalVariable *Variable = nullptr;
```
lenticularis39, line 32: Does this serve an actual purpose here? It seems to me like it's a leftover from VarDependencySlider.

diffkemp/simpll/ControlFlowGraphUtils.h
```diff
@@ -0,0 +1,106 @@
...
+    // Instructions directly dependent on the parameter
```
lenticularis39, line 33: If I understand it correctly this doesn't check dependency on the parameter, but on the differing instructions: the comments should reflect that (not only here, but in the whole file).

diffkemp/simpll/ControlFlowGraphUtils.cpp
```diff
@@ -0,0 +1,742 @@
...
+        if (isDependent(OpInst)) {
+            result = true;
+        }
+    } else if (auto OpOperator = dyn_cast<Operator>(Op)) {
+        for (auto &InnerOp : OpOperator->operands())
+            if (checkDependency(&InnerOp))
+                result = true;
+    }
+    return result;
+}
```
lenticularis39, line 248: This function is unused.

lenticularis39 (changes requested):
The general idea of the algorithm is good, but there are a few things that could possibly be improved in the implementation.

The ControlFlowSlicer class contains a lot of duplicate code from VarDependencySlicer, some of which is not even used (like checkDependency). The code that is used by both VarDependencySlicer and EquivalenceSlicer should be separated into a single class and used by both (e.g. by modifying VarDependencySlicer to use ControlFlowSlicer), not just duplicated.

As of the slicing itself - did you consider implementing the synchronization part in DifferentialFunctionComparator (extending the comparison already implemented there), leaving only the slicing in EquivalenceSlicer? (Also see my comments in EquivalenceSlicer.cpp about the comparison.)

diffkemp/simpll/ControlFlowGraphUtils.h
```diff
@@ -0,0 +1,106 @@
...
+#include "ResultsCache.h"
+#include "SourceCodeUtils.h"
+#include "Utils.h"
+#include <llvm/IR/Module.h>
+#include <queue>
+#include <set>
+
+using namespace llvm;
+
+class ControlFlowSlicer {
```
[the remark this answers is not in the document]
viktormalik: Nice catch, I didn't notice this at all. Candidate names: `DependenceSlicer`, `FunctionSlicer`, `CFGSlicer`. Any other ideas?

viktormalik:
> The ControlFlowSlicer class contains a lot of duplicate code from VarDependencySlicer, some of which is not even used (like checkDependency). The code that is used by both VarDependencySlicer and EquivalenceSlicer should be separated into a single class and used by both (e.g. by modifying VarDependencySlicer to use ControlFlowSlicer), not just duplicated.

I agree with this, move all the common code into `ControlFlowSlicer` (or whatever new name we use) and then make the two other slicers extend that class.

> As of the slicing itself - did you consider implementing the synchronization part in DifferentialFunctionComparator (extending the comparison already implemented there), leaving only the slicing in EquivalenceSlicer? (Also see my comments in EquivalenceSlicer.cpp about the comparison.)

I'd rather have this separated. `DifferentialFunctionComparator` is already large enough, there's no need to extend it with new functionality.

diffkemp/simpll/EquivalenceSlicer.cpp
```diff
@@ -0,0 +1,350 @@
...
+    std::set<BasicBlock *> pushed_QL;
+    std::set<BasicBlock *> pushed_QR;
+    // Get entry blocks
+    QL.push(&(config.FirstFun->getEntryBlock()));
+    QR.push(&(config.SecondFun->getEntryBlock()));
+
+    fComp->getSnMaps().first->erase(fComp->DifferingInstructions.first);
+    fComp->getSnMaps().second->erase(fComp->DifferingInstructions.second);
+
+    while (!QL.empty() && !QR.empty()) {
```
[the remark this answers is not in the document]
viktormalik: Not sure here, but I think that we came to an agreement that we need to use a different approach to the comparison from what is in `DifferentialFunctionComparator`.

On the other hand, I agree that we should try to resume the comparison from where the DFC left, but I'm not sure if that's possible here. Will have to think about this more.

viktormalik (changes requested):
I started with a high-level review only.

Probably the main problem that I can see now is that the slicing algorithm is very long and complex, and therefore hard to understand. What would probably help here would be to split it into multiple, conveniently named functions. 

Let's start with that (and with the current review from @lenticularis39) and then I'll do another round.

tmalecova:
I updated the PR according to your reviews. I am aware of failures in tests, and also of missing comments in the code. I will work on it,  but during that, feel free to make another round of reviews.

PR #184, opened 2021-03-16
Support LLVM 11

viktormalik:
A number of changes was required:
- `StringRef` is not implicitly casted to `std::string` (needs `.str()` call)
- `CallInst::getCalledValue` was replaced by `CallInst::getCalledOperand`
- `SequentialType` was replaced by `ArrayType`
- custom `AnalysisManager` needs to include `<llvm/IR/PassManagerImpl.h>`
- some functions have changed headers (`CallInst::Create`, `InlineFunction`)
- `AttributeList` does not have context field
- `CallInst::getParamAlignment` was replaced by `CallInst::getParamAlign`

The container image is updated with LLVM 11 and Python 3.8 (as Ubuntu 16.04 contains Python 3.6 which has a bug in pip).

### DIFF
--- a/diffkemp/simpll/Utils.cpp
+++ b/diffkemp/simpll/Utils.cpp
@@ -15,6 +15,7 @@
 #include "Config.h"
 #include <algorithm>
 #include <iostream>
+#include <llvm/IR/Intrinsics.h>
 #include <llvm/IR/Module.h>
 #include <llvm/IR/Operator.h>
 #include <llvm/IR/PassManager.h>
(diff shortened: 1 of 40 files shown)


diffkemp/simpll/Utils.cpp
```diff
@@ -50,25 +51,47 @@ static std::vector<Attribute::AttrKind> badVoidAttributes = {
         Attribute::AttrKind::Dereferenceable,
         Attribute::AttrKind::DereferenceableOrNull};
 
-/// Extract called function from a called value Handles situation when the
-/// called value is a bitcast.
-const Function *getCalledFunction(const Value *CalledValue) {
-    const Function *fun = dyn_cast<Function>(CalledValue);
+/// Convert a value to a function.
+/// Handles situation then the actual function is inside a bitcast or alias.
```
PSilling, line 55: Typo: should probably be `Handles situations when`.

diffkemp/simpll/Utils.cpp
```diff
@@ -50,25 +51,47 @@ static std::vector<Attribute::AttrKind> badVoidAttributes = {
...
+const Function *valueToFunction(const Value *Value) {
+    auto *fun = dyn_cast<Function>(Value);
```
PSilling, line 57: I believe that `Fun` would be better here, since other variable names are also capitalized.

PSilling (changes requested):
Other than the minor cosmetic changes, I do not see any issues. Since all tests passed, I would say that the PR is in a very good shape.

PSilling (approved):

tmalecova (approved):

PR #185, opened 2021-03-17
Adjusting compilation for macOS

tmalecova:
To compile Diffkemp in macOS, we need explicitly add system libraries for linking. 
Also, it is needed to explicitly define C++14 standard.
By default, insertion to map makes copy in macOS. `Class Result` do not
have copy constructor, so it caused errors. Therefore, we use `emplace` to
fix it.

### DIFF
--- a/diffkemp/simpll/simpll_build.py
+++ b/diffkemp/simpll/simpll_build.py
@@ -23,20 +23,25 @@ ffibuilder = FFI()
 
 ffibuilder.cdef(get_c_declarations("diffkemp/simpll/library/FFI.h"))
 
-llvm_libs = ["irreader", "passes", "support"]
+llvm_components = ["irreader", "passes", "support"]
 llvm_cflags = check_output(["llvm-config", "--cflags"])
-llvm_ldflags = check_output(["llvm-config", "--libs"] + llvm_libs)
+llvm_ldflags = check_output(["llvm-config", "--ldflags"])
+llvm_libs = check_output(["llvm-config", "--libs"] + llvm_components +
+                         ["--system-libs"])
 
 llvm_cflags = list(filter(lambda x: x != "",
-                          llvm_cflags.decode("ascii").strip().split(" ")))
+                          llvm_cflags.decode("ascii").strip().split()))
 llvm_ldflags = list(filter(lambda x: x != "",
-                           llvm_ldflags.decode("ascii").strip().split(" ")))
+                           llvm_ldflags.decode("ascii").strip().split()))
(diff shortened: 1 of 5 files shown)


diffkemp/simpll/simpll_build.py
```diff
@@ -23,20 +23,25 @@ def get_c_declarations(header_filename):
...
+llvm_libs = list(filter(lambda x: x != "",
+                        llvm_libs.decode("ascii").strip().split()))
 
 ffibuilder.set_source(
     "diffkemp.simpll._simpll", '#include <library/FFI.h>',
     libraries=['simpll-lib'],
     extra_compile_args=["-Idiffkemp/simpll"] + llvm_cflags,
-    extra_link_args=["-Lbuild/diffkemp/simpll", "-lstdc++"] + llvm_ldflags)
+    extra_link_args=["-Lbuild/diffkemp/simpll", "-lstdc++"] + llvm_ldflags +
+    llvm_libs)
```
viktormalik, line 44: Would PEP8 allow this?
```suggestion
    extra_link_args=["-Lbuild/diffkemp/simpll", "-lstdc++"] + llvm_ldflags +
                    llvm_libs)
```

tests/unit_tests/simpll/CMakeLists.txt
```diff
@@ -26,4 +26,9 @@ set_target_properties(runTests
   PROPERTIES
   RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 exec_program(llvm-config ARGS --libs irreader passes support OUTPUT_VARIABLE llvm_libs)
+exec_program(llvm-config ARGS --system-libs OUTPUT_VARIABLE system_libs)
+string(STRIP ${system_libs} system_libs)
```
viktormalik, line 30: Add newline after this line

viktormalik (changes requested):
Just a few small whitespace nits

diffkemp/simpll/simpll_build.py
tmalecova: According to `flake8 diffkemp --ignore=F403,F405,W504`, it is correct. Your suggestion will cause this error:
`diffkemp/simpll/simpll_build.py:44:21: E131 continuation line unaligned for hanging indent`.

viktormalik: Yeah, I thought so, let's leave it as it is, then.

viktormalik (approved):

PR #193, opened 2021-08-27
Add verbose levels

### LINKED ISSUE #36: Better verbosity options
Improve verbosity settings for DiffKemp. Enable multiple verbosity levels controlling various outputs (info about comparison results, builder output, various debugging information). This should be also implemented for SimpLL (introduce a verbosity CLI option instead of using `#ifdef DEBUG`).

Kuba314:
Added a new, more verbose verbosity level.

This new verbosity level can be selected by using a `-vv` parameter.
So far I've only moved each passes' debug output to this new level. Other debug messages stay on the same level as before.

Resolves #36

### DIFF
--- a/diffkemp/simpll/Config.cpp
+++ b/diffkemp/simpll/Config.cpp
@@ -46,9 +46,9 @@ cl::opt<bool> ControlFlowOpt(
 cl::opt<bool> PrintCallstacksOpt(
         "print-callstacks",
         cl::desc("Print call stacks for non-equal functions."));
-cl::opt<bool>
-        VerboseOpt("verbose",
-                   cl::desc("Show verbose output (debugging information)."));
+cl::opt<int> VerbosityOpt(
+        "verbosity",
+        cl::desc("Level of verbose output (debugging information)."));
 cl::opt<bool> VerboseMacrosOpt("verbose-macros",
                                cl::desc("Show debugging information for "
                                         "discovering macro differences"));
@@ -107,9 +107,14 @@ Config::Config()
     }
 
     std::vector<std::string> debugTypes;
-    if (VerboseOpt) {
-        // Enable debugging output in passes
-        debugTypes.emplace_back(DEBUG_SIMPLL);
+    if (VerbosityOpt > 0) {
+        // Enable debugging output in passes. Intended fallthrough
+        switch (VerbosityOpt) {
+        default:
+        case 1:
+            debugTypes.emplace_back(DEBUG_SIMPLL);
+            break;
+        }
     }
     if (VerboseMacrosOpt) {
         // Enable debugging output when finding macro differences
@@ -145,7 +150,7 @@ Config::Config(std::string FirstFunName,
                bool ControlFlowOnly,
                bool PrintAsmDiffs,
                bool PrintCallStacks,
-               bool Verbose,
+               int Verbosity,
                bool VerboseMacros)
         : First(std::move(FirstModule)), Second(std::move(SecondModule)),
           FirstFunName(FirstFunName), SecondFunName(SecondFunName),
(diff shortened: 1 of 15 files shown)


lenticularis39 (changes requested):
Please have a look at the code style CI run, there are some complaints from clang-format about code formatting.

diffkemp/simpll/Config.h
```diff
@@ -21,6 +21,7 @@
 #include <llvm/Support/SourceMgr.h>
 
 #define DEBUG_SIMPLL "debug-simpll"
+#define VERBOSE_SIMPLL "verbose-simpll"
```
viktormalik, line 24: `DEBUG_SIMPLL_VERBOSE` would be better here (to be consistent with the other debug types by using the `DEBUG` prefix).

diffkemp/simpll/Config.cpp
```diff
@@ -161,9 +169,17 @@ Config::Config(std::string FirstFunName,
     }
 
     std::vector<std::string> debugTypes;
-    if (Verbose) {
-        // Enable debugging output in passes
-        debugTypes.emplace_back(DEBUG_SIMPLL);
+    if (VerbosityOpt > 0) {
```
viktormalik, line 172: Should be like this, I believe:
```suggestion
    if (Verbosity > 0) {
```
`VerbosityOpt` is not relevant here as this constructor is used when SimpLL is **not** invoked from CLI.

diffkemp/simpll/Config.cpp
```diff
@@ -161,9 +169,17 @@ Config::Config(std::string FirstFunName,
...
+
+        // Enable debugging output in passes. Intended fallthrough
+        switch (VerbosityOpt) {
```
viktormalik, line 175:
```suggestion
        switch (Verbosity) {
```

viktormalik (changes requested):
I think that we should eventually merge the `--verbose-macros` option with the new verbosity levels. There are two options:
- add it to level 2
- create a new level for it (3)

I have a slight preference towards the second option as the output with `--verbose-macros` is quite long. WDYT @lenticularis39?

Anyways, I'm completely fine with doing this in a separate PR, just mentioning it here as it is related.

diffkemp/simpll/Config.h
Kuba314: What would be the 3rd level called? `DEBUG_SIMPLL_VERBOSE_MACROS` or `DEBUG_SIMPLL_VERBOSE_3`. I don't think either of those would be fitting. The first one says that only macros are at this level, and the third one wouldn't quite fit the naming scheme.

viktormalik: We could call it `DEBUG_SIMPLL_VERBOSE_EXTRA`, that way, each level has an added word.

Kuba314:
- Added third verbosity level (`-vvv`)
- Fixed code formatting
- Rebased onto master

viktormalik:
Oh, and please respect the 50/72 rule for the commit messages.

viktormalik (approved):

viktormalik:
@lenticularis39 could you please look at this? Thanks.

diffkemp/simpll/Config.cpp
```diff
@@ -107,13 +104,19 @@ Config::Config()
...
-        debugTypes.emplace_back(DEBUG_SIMPLL_MACROS);
+    if (VerbosityOpt > 0) {
+
+        // Enable debugging output in passes. Intended fallthrough
+        switch (VerbosityOpt) {
+        default:
+        case 3:
+            debugTypes.emplace_back(DEBUG_SIMPLL_VERBOSE_EXTRA);
+        case 2:
+            debugTypes.emplace_back(DEBUG_SIMPLL_VERBOSE);
```
lenticularis39, line 115: `[[fallthrough]]` attribute can be used to tell the compiler that the fallthough is intented: 
```suggestion
        case 3:
            debugTypes.emplace_back(DEBUG_SIMPLL_VERBOSE_EXTRA);
            [[fallthrough]];
        case 2:
            debugTypes.emplace_back(DEBUG_SIMPLL_VERBOSE);
            [[fallthrough]];
```

diffkemp/simpll/Config.cpp
```diff
@@ -161,13 +163,19 @@ Config::Config(std::string FirstFunName,
...
+        switch (Verbosity) {
+        default:
+        case 3:
+            debugTypes.emplace_back(DEBUG_SIMPLL_VERBOSE_EXTRA);
+        case 2:
+            debugTypes.emplace_back(DEBUG_SIMPLL_VERBOSE);
+        case 1:
+            debugTypes.emplace_back(DEBUG_SIMPLL);
+            break;
+        }
```
lenticularis39, line 178: The comments from the first occurrence of this code block also apply here.

lenticularis39 (approved):

Issue #67, opened 2019-07-23
SimpLL: print info about analysis phases


viktormalik:
For each phase of the analysis (pre-process, individual passes, analysis itself), write to stderr a message that the phase started. Indent all messages belonging to phases so that it is clear which message belongs to which phase.
For each phase and each pass, add a SimpLL CLI option to turn on/off messages for that phase. By default, only print messages for the analysis itself.
Remove the post-process phase which only does inlining that is not needed anymore.

viktormalik:
> Remove the post-process phase which only does inlining that is not needed anymore.

An I believe that this has been already removed.

viktormalik:
> I added debug prints at the places you mentioned. This is how it looks now.

This looks good, I think. Perhaps also add module names to `Generating function abstractions...` and `Removing unused return values...`.

> Also, do you have any ideas on what exactly to print on the second verbosity level, or should I figure that out by myself?

Some passes already print some info on the second level, so I'd leave that as it is.

PR #194, opened 2021-08-27
SimpLL: Make debug comparison report more readable

Kuba314:
I attempted to make the comparison process output more readable by changing text reporting such as `Functions are equal` to a shorter, simpler and more readble `==`. I also added some colors for clarity.

I need some opinions on the formatting, because I'm still not entirely satisfied with how it looks.

image

### DIFF
--- a/diffkemp/simpll/ModuleComparator.cpp
+++ b/diffkemp/simpll/ModuleComparator.cpp
@@ -115,6 +115,11 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
         DEBUG_WITH_TYPE(DEBUG_SIMPLL,
                         dbgs() << getDebugIndent() << "} "
                                << Color::makeRed("not equal\n")
+                               << Color::makeRed("========== ")
+                               << "Found difference between \""
+                               << FirstFun->getName() << "\" and \""
+                               << SecondFun->getName() << "\""
+                               << Color::makeRed(" ==========\n"););
         ComparedFuns.at({FirstFun, SecondFun}).kind = Result::NOT_EQUAL;
 
         while (tryInline.first || tryInline.second) {


viktormalik:
First of all - this looks much more readable than before, nice!

Several notes (and maybe suggestions):
- I don't like the braces too much, but the truth is that they make things much more readable (and the output can be now processed in an editor). Not sure if we can find something better here.
- I'd prefer using "equal" and "not equal" instead of "==" and "!=". It's longer, so it catches the eye more easily even if colours are not used (which will be the case for some terminals).
- Usually, you want to see which function is *not equal*. If colors are not available, this may be difficult to spot, so maybe we could add something like:
```
========== Found difference between "__vunmap" and "__vunmap" ==========
```
- I think that this PR could also add printing of information about analysis phases (see issue #67).

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -82,20 +89,30 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
...
-                });
+            DEBUG_SIMPLL,
+
+            decreaseDebugIndentLevel();
+            switch(ComparedFuns.at({FirstFun, SecondFun}).kind) {
+            case Result::EQUAL:
+                dbgs() << "declaration, names are "
+                       << Color::makeGreen("equal")
+                       << " }\n";
+                break;
```
viktormalik, line 100: I just realized that this case is useless as this code path never sets the `EQUAL` result (uses `ASSUMED_EQUAL` instead). Remove this case and use green colour for `ASSUMED_EQUAL`.

diffkemp/simpll/ModuleAnalysis.cpp
```diff
@@ -189,17 +189,22 @@ void simplifyModulesDiff(Config &config, OverallResult &Result) {
...
                 DEBUG_WITH_TYPE(DEBUG_SIMPLL,
-                                dbgs() << funPairResult.first.first->getName()
-                                       << " are syntactically different\n");
+                                dbgs() << "  " << Color::makeRed("==========")
+                                       << " Found difference between \""
+                                       << funPairResult.first.first->getName()
+                                       << "\" and \""
+                                       << funPairResult.first.second->getName()
+                                       << "\" "
+                                       << Color::makeRed("==========\n"););
```
viktormalik, line 198: I'd prefer leaving this part as it was and moving this new easy-to-spot notification to the point where function non-equality is detected (i.e. where `} not equal` is printed).

Kuba314: Like this?
image

viktormalik: Yes, great! Feel free to shift the line all the way to the left (i.e. no indent).

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -203,13 +217,15 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
...
+                                       << Color::makeGreen("equal")
+                                       << " after inlining\n");
                 ComparedFuns.at({FirstFun, SecondFun}).kind = Result::EQUAL;
             } else {
                 DEBUG_WITH_TYPE(DEBUG_SIMPLL,
-                                dbgs() << getDebugIndent() << "After inlining, "
-                                       << "the functions are not equal\n");
+                                dbgs() << getDebugIndent() << "still "
+                                       << Color::makeRed("not equal")
+                                       << " after inlining\n");
```
viktormalik, line 228: Sometimes, this can be quite far from the original function comparison. Could you add function names so that it's clearer? Something like:
```
"function1" and "function2" still not equal after inlining
```
with `not equal` being red.

PR #195, opened 2021-08-28
Merge occurences of EQUAL_SYNTAX with EQUAL

### LINKED ISSUE #191: Get rid of the equal-syntax result kind
Historically, *equal-syntax* was used when equality was shown by DiffKemp itself while *equal* was used when it had to call an external formal verification tool (LLReve). Since DiffKemp is now capable to handle many semantics-preserving changes, the *equal-syntax* result is not accurate anymore and should be removed. I suggest to merge it with *equal* and once we start supporting external tools again, we can come with a new result kind if necessary.

Kuba314:
Resolves #191

### DIFF
--- a/diffkemp/diffkemp.py
+++ b/diffkemp/diffkemp.py
@@ -331,7 +331,7 @@ def compare(args):
                         if pattern.search(called_res.diff):
                             break
                     else:
-                        fun_result.kind = Result.Kind.EQUAL_SYNTAX
+                        fun_result.kind = Result.Kind.EQUAL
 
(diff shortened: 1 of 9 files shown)


viktormalik (approved):

viktormalik:
Please rebase onto master to restart the CI.

viktormalik (changes requested):
Tests are failing since you need to replace `equal_syntax` in test specs, too.

viktormalik (approved):

Issue #202, opened 2021-12-10
Building with -Wall -Wextra yields a lot of warnings


viktormalik:
Running:
```
mkdir build && cd build
cmake .. -GNinja -DCMAKE_CXX_FLAGS="-Wall -Wextra"
ninja
```
prints a lot of warnings. It would be nice to fix those.

viktormalik:
Resolved via #203

PR #203, opened 2021-12-10
enable warnings on simpll build & fix warnings

parallelistic:
PR for #202 

I'm unsure of what would be the best approach for -Wunused-parameters, I don't think spamming `__attribute__((unused))` everywhere in passes would be very nice to look at...
Other than that, I haven't performed any big changes to tests, nor do I think they would be needed.

SimpLL tests passed, and haven't run pytest, so if someone with the different kernels downloaded could do that, I would be grateful.


viktormalik:
Thanks for the PR!

> I'm unsure of what would be the best approach for -Wunused-parameters, I don't think spamming `__attribute__((unused))` everywhere in passes would be very nice to look at... 

Making those arguments unnamed should be sufficient here.

> SimpLL tests passed, and haven't run pytest, so if someone with the different kernels downloaded could do that, I would be grateful.

CI took care of that ;-)

parallelistic:
> > I'm unsure of what would be the best approach for -Wunused-parameters, I don't think spamming `__attribute__((unused))` everywhere in passes would be very nice to look at...
> 
> Making those arguments unnamed should be sufficient here.

Can't believe i didn't think of that 👍

diffkemp/simpll/CMakeLists.txt
```diff
@@ -1,16 +1,20 @@
 # Create symlinks to a correct version of FunctionComparator based on LLVM version
-execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink
+execute_process(
+  COMMAND ${CMAKE_COMMAND} -E create_symlink
```
viktormalik, line 3: Please do not reformat lines that are not related to this PR, it makes `git blame` useless (this applies to almost the entire file).

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -600,7 +600,7 @@ bool DifferentialFunctionComparator::maySkipLoad(const LoadInst *Load) const {
                     searchPredecessors = false;
                     break;
                 }
-            } else if (auto Store = dyn_cast<StoreInst>(&*it)) {
+            } else if (/* auto Store = */ dyn_cast<StoreInst>(&*it)) {
```
viktormalik, line 603:
```suggestion
            } else if (isa<StoreInst>(&*it)) {
```

diffkemp/simpll/DebugInfo.cpp
```diff
@@ -148,8 +148,8 @@ void DebugInfo::extractAlignmentFromInstructions(GetElementPtrInst *GEP,
                 if (!TypeDISecond)
                     continue;
 
-                int indexSecond =
-                        getTypeMemberIndex(*TypeDISecond, elementName);
+                uint64_t indexSecond = static_cast<uint64_t>(
+                        getTypeMemberIndex(*TypeDISecond, elementName));
```
viktormalik, line 152: Just use `(uint64_t)` here (and everywhere else). LLVM does the same, so we follow its approach.

viktormalik (changes requested):

diffkemp/simpll/CMakeLists.txt
parallelistic: My mistake, my vim autocmd was at fault and I realised as much before, which is made me run clang-format but then I forgot about the CMakeLists

diffkemp/simpll/CMakeLists.txt
```diff
@@ -19,7 +23,7 @@ file(GLOB passes passes/*.cpp)
 file(GLOB library library/*.cpp)
 
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti -fpic -fvisibility=hidden")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti -fpic -fvisibility=hidden -Wall -Wextra -Wno-reorder")
```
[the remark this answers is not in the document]
parallelistic: It is in fact not, i had completely forgotten I had added it and therefor thought I finished fixing all errors :/

viktormalik (approved):
Looks good! Can you squash everything into a single commit which will describe a list of changes that were done?

parallelistic:
> Looks good! Can you squash everything into a single commit which will describe a list of changes that were done?

Yup will do, was thinking about doing exactly that :)

viktormalik:
Two last things related to the commit message:
1. please respect the 50/72 rule
2. if you don't mind, it'd be nice if you could use some actual (non-GitHub) email for your commits

parallelistic:
done :+1:

PR #206, opened 2021-12-15
SimpLL: Print info about analysis phases

### LINKED ISSUE #67: SimpLL: print info about analysis phases

Kuba314:
Resolves #67 
Adds these messages:
```
Analysing debugging information...
Preprocessing {mod_name}...
Slicing control-flow-only statements...
Generating function abstractions in module {mod_name}...
Removing unused return values in module {mod_name}...
Slicing {fun_name} w.r.t. value of {var_name}...
```

### DIFF
--- a/diffkemp/simpll/DebugInfo.cpp
+++ b/diffkemp/simpll/DebugInfo.cpp
@@ -12,7 +12,6 @@
 //===----------------------------------------------------------------------===//
 
 #include "DebugInfo.h"
-#include "Config.h"
 #include <llvm/IR/Constants.h>
 #include <llvm/Passes/PassBuilder.h>
 
--- a/diffkemp/simpll/passes/FunctionAbstractionsGenerator.cpp
+++ b/diffkemp/simpll/passes/FunctionAbstractionsGenerator.cpp
@@ -26,6 +26,10 @@ AnalysisKey FunctionAbstractionsGenerator::Key;
 /// and for each pair of assembly code and constraint.
 FunctionAbstractionsGenerator::Result FunctionAbstractionsGenerator::run(
         Module &Mod, AnalysisManager<Module, Function *> &mam, Function *Main) {
+    DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+                    dbgs() << "Generating function abstractions in "
+                           << Mod.getName() << "...\n";
+                    increaseDebugIndentLevel());
     FunMap funAbstractions;
     int i = 0;
     std::vector<Instruction *> toErase;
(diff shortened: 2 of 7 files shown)


diffkemp/simpll/DebugInfo.cpp
```diff
@@ -20,6 +20,8 @@ using namespace llvm;
 
 PreservedAnalyses RemoveDebugInfoPass::run(Function &Fun,
                                            FunctionAnalysisManager & /*fam*/) {
+    DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+                    dbgs() << "Analysing debugging information...\n");
```
viktormalik, line 24: This should go to `DebugInfo` (probably the best place would be its constructor).

diffkemp/simpll/passes/FunctionAbstractionsGenerator.cpp
```diff
@@ -26,6 +26,9 @@ AnalysisKey FunctionAbstractionsGenerator::Key;
 /// and for each pair of assembly code and constraint.
 FunctionAbstractionsGenerator::Result FunctionAbstractionsGenerator::run(
         Module &Mod, AnalysisManager<Module, Function *> &mam, Function *Main) {
+    DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+                    dbgs() << "Generating function abstractions in module "
```
viktormalik, line 30:
```suggestion
                    dbgs() << "Generating function abstractions in "
```
... same for other messages (the word "module" is redundant here).

viktormalik (changes requested):
Nice, thanks!

I'd suggest one more improvement - for each pass where you added this message, indent all other debug messages printed by that pass by one level (you can use functions `*DebugIndent*` from `Utils.h`). Similarly, for passes that are run from `preprocessModule`, indent their main info message by one level (and, consequently, all their inner messages by 2 levels).

diffkemp/simpll/DebugInfo.cpp
Kuba314: Does that mean creating a destructor and putting `decreaseDebugIndentLevel()` in there?

viktormalik: Most of the analysis is done in the constructor, so I'd put `decreaseDebugIndentLevel()` to its end.

Kuba314:
@viktormalik Could you rerun CI? Don't know why it failed exactly, but it seems like it wasn't my fault.

viktormalik (approved):

viktormalik:
Please rebase once the CI passes.

viktormalik (approved):

PR #214, opened 2022-02-15
Fix instruction ignoring bug

viktormalik:
When getting a replacement value for an ignored instruction in `cmpValues`, besides querying the `ignoredInstructions` map, we allow to skip bitcast and zero GEP operators. The problem is that the replacing value may not be already synchronized and hence an incorrect synchronisation may be created by `cmpValues`.

This fixes the above problem by checking that if the replacing value is an instruction, it is already synchronised.

See the added unit test for an example.


PSilling (approved):
Looks great!

PR #217, opened 2022-02-22
Replace uses of a global before erasing it (fix memory error)

viktormalik:
`SimplifyKernelGlobalsPass` removes global variables having the `__ksym` prefix. Before erasing such a global, we need to replace all its uses by `nullptr` so that there are no memory leaks.

This should also fix a bug that non-determinisically appears for the regression test `rhel-73-74_debugfs_create_u32` with LLVM 13.


lenticularis39 (approved):
Interesting, I remember working on this pass before and IIRC the idea was that there weren't any other uses of the `__ksym` symbols besides in the `llvm.used` intrinsic, it being only for having the symbol table in the resulting binary after compiling into native code.

Nevertheless that is a different issue, I agree with replacing the uses with null for now.

PR #219, opened 2022-02-28
Release version 0.3

viktormalik:
Changes in this PR mostly fix small bugs, cleanup the specfile, and update version numbers.

The list of released changes can be found in the changelog in specfile.


PSilling (approved):
Changes look great.

I just have a small note regarding "Eliminate all remaining build warnings". This depends on the compiler. With clang, I'm currently plagued by many other build warnings after `-Wall` got added in one of the more recent commits. Just an example:

`../diffkemp/simpll/Result.h:116:14: warning: private field 'kind' is not used [-Wunused-private-field]`

I don't think we need to deal with this now, but I'll look into it in the future so we can eliminate clang warnings as well.

viktormalik:
> I don't think we need to deal with this now, but I'll look into it in the future so we can eliminate clang warnings as well.

Thanks for the notice. There's also a bunch of warnings in the unit tests which I didn't fix.

PR #220, opened 2022-03-07
Save LLVM version when running diffkemp generate

### LINKED ISSUE #199: Add LLVM version to snapshots
There tend to be some incompatibilities between LLVM IR of different LLVM versions, leading to a snapshot generated by one version non-readable by DiffKemp built against other version. Currently this can cause a "mysterious" segmentation fault; if the LLVM version was added to the snapshot, a warning/error could be printed when attempting to use such a snapshot.

zacikpa:
This PR resolves #199.

When running `diffkemp generate`, saves the current LLVM **major** version into the snapshot YAML file.
When running `diffkemp compare`, checks whether the major version in use now is the same as in both snapshots. If not, fails.

There is no need to check for minor version differences, as [minor versions are never released](http://blog.llvm.org/2016/12/llvms-new-versioning-scheme.html). Furthermore, different patch versions should be compatible, so they are not checked either.

### DIFF
--- a/diffkemp/diffkemp.py
+++ b/diffkemp/diffkemp.py
@@ -15,6 +15,7 @@ import errno
 import os
 import re
 import sys
+import subprocess
 
 
 def __make_argument_parser():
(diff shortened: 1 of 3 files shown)


diffkemp/diffkemp.py
```diff
@@ -269,6 +270,22 @@ def compare(args):
     old_snapshot = Snapshot.load_from_dir(args.snapshot_dir_old)
     new_snapshot = Snapshot.load_from_dir(args.snapshot_dir_new)
 
+    # Check if snapshot LLVM versions are compatible with the current version
+    llvm_version = subprocess.check_output(
+        ["llvm-config", "--version"]).decode().rstrip().split(".")[0]
+
+    if llvm_version != old_snapshot.llvm_version:
+        sys.stderr.write(
+            "Error: old snapshot uses LLVM {}, current is {}.\n".format(
```
viktormalik, line 279:
```suggestion
            "Error: old snapshot was built with LLVM {}, current is {}.\n".format(
```
Same below.

viktormalik (changes requested):

diffkemp/diffkemp.py
zacikpa: Fixed.

viktormalik (approved):
Can you squash the commits, please?

PR #222, opened 2022-03-14
[WIP] Delete cscope.files when cscope database build fails

### LINKED ISSUE #212: Remove CScope files if creating database is interrupted
When using `--kernel-with-builder` to generate a snapshot, CScope is not run in case `cscope.files` exists in the kernel directory. That file is created by DiffKemp just before CScope database generating is run, however, if the generating fails or is interrupted, the file is not deleted and *generate* fails to find symbols the next time it is run.

This problem could be fixed either by deleting `cscope.files` in case CScope database generating fails or by changing the way to detect if the database was already generated (e.g. by checking for existence of files created during the generating).

zacikpa:
When merged, this PR resolves #212.

I've enclosed the call to _cscope_ into a `try .. except` block, removing _scope.files_ when _scope_ database build fails.
This should also cover the case when _cscope_ is interrupted since the same exception will be raised by _subprocess_.

Is it a viable solution?

### DIFF
--- a/diffkemp/llvm_ir/kernel_llvm_source_builder.py
+++ b/diffkemp/llvm_ir/kernel_llvm_source_builder.py
@@ -169,13 +169,14 @@ class KernelLlvmSourceBuilder(LlvmSourceFinder):
         Build a database for the cscope tool. It will be later used to find
         source files with symbol definitions.
         """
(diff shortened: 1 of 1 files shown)


viktormalik (commented):
Nice, this looks good, just adding one nit.

viktormalik (approved):

PR #223, opened 2022-03-28
WIP: Allow multiple builds using an environment variable

### LINKED ISSUE #211: Remove hard-coded build directory
Currently, we require that SimpLL is built in the `build` directory, otherwise the Python wrapper is not able to find it (this affects building SimpLL both as an [executable](https://github.com/viktormalik/diffkemp/blob/master/diffkemp/simpll/simpll.py#L112-L113) and [library](https://github.com/viktormalik/diffkemp/blob/master/diffkemp/simpll/simpll_build.py#L44)).

It'd be nice to get rid of this dependency to allow having multiple parallel builds. The build directory could be specified when installing the DiffKemp PyPI package (with the default dir set to `build`).

Once this is implemented, we can also update the development container entry point to not only install the PyPI package but to build SimpLL before that. This would remove an error that occurs when running the container without building SimpLL before.

zacikpa:
We could support multiple builds using an environment variable similarly to e.g., [Flask](https://flask.palletsprojects.com/en/2.0.x/config/). The usage of a default build directory would remain the same, usage of a custom one would be as follows:
```bash
mkdir custom_build                       # assume that the custom build is in this directory
export SIMPLL_BUILD_DIR=custom_build     # set the environment variable
pip install -e .                         # install normally (but within the same environment)
bin/diffkemp                             # use diffkemp normally (still in the same environment)
```

After researching the possible options, this seems like the only viable way of allowing multiple builds.
Let me know if this is acceptable.

If accepted, this PR resolves #211.

### DIFF
--- a/diffkemp/simpll/simpll.py
+++ b/diffkemp/simpll/simpll.py
@@ -113,10 +113,16 @@ def run_simpll(first, second, fun_first, fun_second, var, suffix=None,
         try:
             # Determine the SimpLL binary to use.
             # The manually built one has priority over the installed one.
-            if os.path.isfile("build/diffkemp/simpll/diffkemp-simpll"):
-                simpll_bin = "build/diffkemp/simpll/diffkemp-simpll"
+            # A custom build directory has priority over the default one.
+            build_dir_var = "SIMPLL_BUILD_DIR"
+            if build_dir_var in os.environ:
+                build_dir = os.environ[build_dir_var]
             else:
+                build_dir = "build"
+            simpll_bin = "{}/diffkemp/simpll/diffkemp-simpll".format(build_dir)
+            if not os.path.isfile(simpll_bin):
                 simpll_bin = "diffkemp-simpll"
+
             # SimpLL command
             simpll_command = list([simpll_bin, first, second,
                                    "--print-callstacks"])
--- a/diffkemp/simpll/simpll.py
+++ b/diffkemp/simpll/simpll.py
@@ -5,6 +5,7 @@ from diffkemp.llvm_ir.llvm_module import LlvmModule
 from diffkemp.semdiff.caching import ComparisonGraph
 from diffkemp.simpll._simpll import ffi, lib
 from diffkemp.simpll.library import SimpLLModule
+from diffkemp.simpll.simpll_build import get_simpll_build_dir
 import os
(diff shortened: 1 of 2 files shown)


viktormalik:
This is a nice solution, I like it. @lenticularis39 what do you think?

lenticularis39:
I agree with @viktormalik, I just have a small comment to the implementation. 

The code for getting the build directory is almost the same in simpll.py and simpll_build.py, including the constant with the environment variable name, therefore it could be moved into a function (e.g. `get_simpll_build_dir`) to avoid code duplication.

zacikpa:
I added a helper function `get_simpll_build_dir`, thanks for the suggestion. I was not sure in which file it would fit the most, feel free to comment on that.

I am still unsure about one thing (that's why a WIP PR):
- During the diffkemp installation, a `build_ext` command is invoked by _setuptools_, which compiles the `_simpll.abi3.so` shared library
- The `_simpll.abi3.so` is firstly created in the `build/` directory (a default value in `build_ext`), and then copied into the `diffkemp/simpll/` directory.
- However, the build directory setting cannot be changed for the `build_ext` command with our current CFFI setup, so it always ends up in `build/`.
- This is only an issue if the `build/` directory is inaccessible (e.g., has no write permissions). Otherwise it doesn't matter much, since the copying takes place.

I don't think this is a major issue, but it may be nice to know.

viktormalik:
> * The `_simpll.abi3.so` is firstly created in the `build/` directory (a default value in `build_ext`), and then copied into the `diffkemp/simpll/` directory.

So, IIUC, if I don't have a directory named `build` in the project root directory, the build will fail? Or it will be automatically created if it doesn't exist? Or this `build` directory is a subdir of the actual build directory?

zacikpa:
The `build` directory will be automatically created. It is not a subdirectory of the actual build directory. The build will only fail if the `build` directory already exists and it has no write access.

viktormalik:
> The `build` directory will be automatically created. It is not a subdirectory of the actual build directory. The build will only fail if the `build` directory already exists and it has no write access.

Which will effectively happen only if `build` has been created by a container and you're trying to rebuild from the local machine. I don't think that it's much of an issue, how difficult would it be to add a warning to `simpll_build.py`?

viktormalik (approved):

PR #226, opened 2022-04-22
Autobuild Diffkemp in Docker containers 

zacikpa:
The container can be run using a Python script `docker/diffkemp-devel/run-container.py`. The script takes several arguments:
- `--build-dir`: required
- `--llvm-version`: optional, default is 13
- `--image`: optional, default is `viktormalik/diffkemp-devel`, i.e. the official DockerHub image

Multiple containers with different build directories and different LLVM versions can run Diffkemp in parallel.

**Update:**
It should now be possible to run the container script from any host directory. In that case, another command-line option must be provided to determine where the diffkemp directory is located:
- `--diffkemp-dir` : optional, default is `.` (when running the script from the root diffkemp directory).


docker/diffkemp-devel/run-container.py
```diff
@@ -0,0 +1,39 @@
+import subprocess
+import os
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--build-dir")
+parser.add_argument("--diffkemp-dir", default=".")
+parser.add_argument("--llvm-version", default="13")
```
viktormalik, line 8: I wonder how difficult it'd be to set the default LLVM version to the latest currently supported version so that we don't have to change this with every LLVM release.

viktormalik (commented):
`docker/diffkemp-devel/run-container.py` should be executable.

docker/diffkemp-devel/run-container.py
```diff
@@ -0,0 +1,39 @@
+import subprocess
```
viktormalik, line 1:
```suggestion
#!/bin/env python3
import subprocess
```
+ make this file executable.

viktormalik (changes requested):

docker/diffkemp-devel/run-container.py
zacikpa: I removed the default option, transferring the choice to Dockerfile, where the default LLVM version is also specified in a [variable](https://github.com/zacikpa/diffkemp/blob/container-autobuild/docker/diffkemp-devel/Dockerfile#L3). It suppose it's OK to keep that one, since we need to change that file with every LLVM release nonetheless.

viktormalik (approved):
Looks great, thanks! Can you squash the commits a bit, please?

zacikpa:
> Looks great, thanks! Can you squash the commits a bit, please?

Done.

viktormalik (approved):

PR #229, opened 2022-05-09
Various small fixes

viktormalik:
This features a number of small fixes which should be quite easy to review, so posting them together.


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -316,13 +316,9 @@ int DifferentialFunctionComparator::cmpOperations(
...
     // after all their incoming values have been compared and matched.
     if (isa<PHINode>(L) && isa<PHINode>(R)) {
-        auto PhiL = dyn_cast<PHINode>(L);
-        auto PhiR = dyn_cast<PHINode>(R);
-        if (PhiL->getNumIncomingValues() == PhiR->getNumIncomingValues()) {
-            needToCmpOperands = false;
-            phisToCompare.emplace_back(PhiL, PhiR);
-            return 0;
-        }
+        needToCmpOperands = false;
```
lenticularis39, line 319: The comment above should also be updated (If PHI nodes are compared and they have the same number of incoming values...).

lenticularis39 (changes requested):

lenticularis39 (approved):

PR #231, opened 2022-05-19
Allow not saving path to source to the snapshot

viktormalik:
In some cases, it makes sense to deliberately omit storing the path to source dir in snapshot. An example is when the path would be the same for both compared snapshots (e.g. when comparing different Git refs) where storing the path could lead to false negative errors. This also facilitates moving snapshots independently from sources.


lenticularis39 (approved):
Makes sense, sorry for the delay, I somehow forgot about this PR.

PR #233, opened 2022-06-23
Handle comparison of relocated code

viktormalik:
A common refactoring is when a piece of code is moved into another part of the function (e.g. from inside the loop before the loop head). This helps to compare such a change as semantically equal.

Handling code relocations is done in 3 phases, they roughly correspond to states of the new enum `RelocationInfo::Status`.
- Relocation detection is run when a difference is found and no relocation has been found so far (status is `None`). Detection is done by trying to "skip" a block of code in one of the basic blocks and match the remaining parts of the blocks. If it succeeds, the relocation is stored.
- Relocation matching is run when a difference is found and there is a previously found potential relocation (state is `Stored`). In such a case, we reset the current instruction iterator to the head of the stored relocation (in the appropriate program) and continue the comparison (state is changed to `Matching`). Once all relocation instructions were successfully compared, the state is reset to `None` and the last phase is run.
- Relocation checking checks if the relocated code has no dependency on the skipped code. Direct data dependencies are resolved by the SSA property of LLVM IR, so this only checks indirect dependencies through load/store instructions. This is simply done by checking if both the relocated and the skipped code access the same pointer and one of the accesses is a write. In such a case, the relocation is evaluated as not equal.

See commit messages for more implementation details.

Note that relocation detection must be run after all other approaches to show equality (including custom pattern matching) are tried.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -965,6 +965,14 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
             if (Res)
                 return Res;
         } else {
+            if (Reloc.status == RelocationInfo::Stored) {
+                // If there is a dependency between the skipped instruction and
+                // the relocated code, fail the comparison
+                if (Reloc.prog == Program::First && isDependingOnReloc(*InstL))
+                    return 1;
+                if (Reloc.prog == Program::Second && isDependingOnReloc(*InstR))
+                    return 1;
+            }
             if (Reloc.status == RelocationInfo::Matching) {
                 // If the relocated code has been entirely matched, we can
                 // continue from the restore point.
@@ -1743,6 +1751,11 @@ bool DifferentialFunctionComparator::findMatchingOpWithOffset(
             Reloc.prog = prog_to_search;
             Reloc.tryInlineBackup = tryInlineBackup;
 
+            // Make sure that the first equal instruction is not depending on
+            // the relocation
+            if (isDependingOnReloc(*MovedInst))
+                return false;
+
             DEBUG_WITH_TYPE(DEBUG_SIMPLL,
                             dbgs() << getDebugIndent()
                                    << "Possible relocation found:\n"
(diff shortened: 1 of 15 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -898,6 +910,40 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
...
 
+            if (Reloc.status == RelocationInfo::Stored) {
+                // If there is an inequality found and we have previously found
+                // a possibly relocated block, try to match it now.
+                Reloc.status = RelocationInfo::Matching;
+                DEBUG_WITH_TYPE(DEBUG_SIMPLL,
+                                dbgs() << getDebugIndent()
+                                       << "Try to match the relocated block\n");
+                // Reset serial counters since the instructions are not equal
+                sn_mapL.erase(&*InstL);
```
PSilling, line 921: Maybe replace by a call to `DifferentialFunctionComparator::undoLastInstCompare`?

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1624,3 +1706,102 @@ const Value *
...
+    auto tryInlineBackup = ModComparator->tryInline;
+
+    // Mark the possible relocation beginning.
+    Reloc.begin = MovedInst;
+
+    auto BBEnd = MovedInst->getParent()->end();
+
+    // Reset the serial counters since InstL and InstR were already compared as
+    // non-equal and start from the following instruction.
+    sn_mapL.erase(&*InstL);
```
PSilling, line 1736: `DifferentialFunctionComparator::undoLastInstCompare` again.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1624,3 +1706,102 @@ const Value *
...
+    sn_mapR.erase(&*InstR);
+    ++MovedInst;
+    while (MovedInst != BBEnd) {
+        auto sn_mapL_backup = sn_mapL;
```
PSilling, line 1740: `mappedValuesBySn` should get backed up as well since `cmpValues` (i.e., `cmpOperationsWithOperands`) may change it as well.

PSilling (changes requested):
Great quality code. The only issue I have is that it ignores the recently introduced `mappedValuesBySn` map (necessary for custom pattern matching).

viktormalik:
> Great quality code. The only issue I have is that it ignores the recently introduced `mappedValuesBySn` map (necessary for custom pattern matching).

Good point, thanks. Should be fixed now.

PSilling (approved):
Indeed, looks great!

PR #247, opened 2022-11-02
Update README

viktormalik:
Several README updates:
- new usage description aligned with the new CLI
- added LLVM 15
- added NETYS paper reference

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,21 +2,43 @@
 
 # DiffKemp
 
-DiffKemp is a tool for automatic static analysis of semantic differences between
-different versions of the Linux kernel.
+DiffKemp is a framework for automatic static analysis of semantic differences
+between different versions of projects written in C, with main focus on the
+Linux kernel.
+
+The main use-case of DiffKemp is to compare selected functions and configuration
+options in two versions of a project and to report any discovered semantic
+differences.
+
+## About
+
+The main focus of DiffKemp is high scalability, such that it can be applied to
+large-scale projects containing a lot of code. To achieve that, the analysed
+functions are first compiled into LLVM IR, then several code transformations are
+applied, and finally the comparison itself is performed.
+
+Wherever possible, DiffKemp tries to compare instruction-by-instruction (on LLVM
+IR instructions) which is typically sufficient for most of the code. When not
+sufficient, DiffKemp tries to apply one of the built-in or user-supplied
+*semantics-preserving patterns*. If a semantic difference is still discovered,
+the relevant diffs are reported to the user.
+
+Note that DiffKemp is incomplete it its nature, hence may provide false positive
+results (i.e. claiming that functions are not equivalent even though they are).
+This especially happens with complex refactorings.
 
 ## Installation
 
-There are two options to install the project, either build from source or use a
+There are two options to install DiffKemp, either build from source or use a
 prepared RPM package for Fedora.
 
 ### Install from source
 
 Currently, DiffKemp runs on Linux and needs the following software installed:
-* Clang and LLVM (supported versions are 9, 10, 11, 12, 13, and 14)
+* Clang and LLVM (supported versions are 9, 10, 11, 12, 13, 14, 15)
 * Python 3 with CFFI (package `python3-cffi` in Fedora and Debian)
 * Python packages from `requirements.txt` (run `pip install -r requirements.txt`)
-* CScope
+* CScope (when comparing versions of the Linux kernel)
 
 Additionally, to build manually, you need to install the following tools:
 * CMake
@@ -32,6 +54,8 @@ Build can be done by running:
 
     pip install -e .
 
+The DiffKemp binary is then located in `bin/diffkemp`.
+
 ### Install from RPM
 
 Alternatively, you can use a prepared RPM package for Fedora that can be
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,45 +112,23 @@ Currently, these sysctl option groups are supported: `kernel.*`,
...
+important steps:
+- Simplification of the compared programs by applying multiple code
+  transformations such as dead code elimination, indirect calls abstraction,
+  etc.
+- Debug info parsing to collect information important for the comparison.
+  DiffKemp needs the analysed project to be built with debugging information in
+  order to work properly.
+- The comparison itself is mostly done instruction-by-instruction (on the level
+  of LLVM IR). In addition, DiffKemp handles semantics-preserving changes that
+  adhere to one the built-in patterns. Additional patterns can be specified
```
zacikpa, line 128:
```suggestion
  adhere to one of the built-in patterns. Additional patterns can be specified
```

README.md
```diff
@@ -32,47 +33,69 @@ Build can be done by running:
...
+    all the tools necessary for building kernel must be installed.
+
+  - ```
+    diffkemp llvm-to-snapshot PROJ_DIR LLVM_FILE SNAPSHOT_DIR SYMBOL_LIST
+    ```
+    can be used if the project is already compiled into a single LLVM IR file.
+    The file name is given in `LLVM_FILE` and must be relative to `PROJ_DIR`.
+    The remaining options are the same as for the other commands.
+
+- **semantic comparison** takes two snapshots and compares them for semantic
```
zacikpa, line 81:
```suggestion
- **Semantic comparison** takes two snapshots and compares them for semantic
```

zacikpa (commented):
Looks great, the structure is much cleaner compared to the previous version!

README.md
```diff
@@ -32,47 +33,69 @@ Build can be done by running:
...
+  and creates so-called *snapshots* which contain the relevant LLVM IR files and
+  additional metadata.
+
+  There are several options for snapshot generation:
+  - ```
+    diffkemp build PROJ_DIR SNAPSHOT_DIR SYMBOL_LIST
+    ```
+    is the default snapshot generation command for `make`-based projects. It
+    takes the project located in `PROJ_DIR`, builds it into LLVM IR, and creates
+    a snapshot for comparing semantics of functions from `SYMBOL_LIST`. The
```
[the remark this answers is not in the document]
viktormalik: Good point, I didn't realize that, fixed.

zacikpa (approved):

README.md
```diff
@@ -32,47 +33,70 @@ Build can be done by running:
 
     pip install -e .
 
+The DiffKemp binary is then located in `bin/diffkemp`.
```
lenticularis39, line 36: This might be a good place to mention how DiffKemp works in respect to current working directory and `SIMPLL_BUILD_DIR`.

README.md
```diff
@@ -3,7 +3,8 @@
 # DiffKemp
 
 DiffKemp is a tool for automatic static analysis of semantic differences between
-different versions of the Linux kernel.
+different versions of projects written in C, with main focus on the Linux
```
lenticularis39, line 6: Now when this first paragraph mentions "different versions of projects", it might be better to say "There are two options to install the *tool*" in the following Installation section, to avoid confusion between DiffKemp and the compared project.

Another idea: I'd also consider calling DiffKemp *framework* here instead of *tool*, given the recent and future changes, like in Petr's NETSYS 2022 talk.

lenticularis39 (commented):
I'm not sure if we talked about this before, but did you consider putting the About section right after the introductory sentence (before Installation section)? IMO the readme introducing DiffKemp in one sentence and then jumping straight to installation and command line usage is not a good way to introduce DiffKemp project to someone new, since DiffKemp presents an idea that is likely new to anyone who didn't arrive here from a talk about DiffKemp (and I can imagine even those who did would appreciate a refresher).

Also tips on how to achieve best results with DiffKemp (like build with debug information, filter out Clang-specific options in `build` phase, ...) might be useful, maybe in the future when there will be more of such information. (I'm not sure if the readme is the best place for this, though.)

README.md
viktormalik: We may rethink this soon, so I'd keep it as it is for now.

viktormalik:
> I'm not sure if we talked about this before, but did you consider putting the About section right after the introductory sentence (before Installation section)?

This is a great point. I created an "About" section at the top, although I put there slightly different information from what was in the original "About" section which is now renamed to "Important implementation details".

> Also tips on how to achieve best results with DiffKemp (like build with debug information, filter out Clang-specific options in `build` phase, ...) might be useful, maybe in the future when there will be more of such information. (I'm not sure if the readme is the best place for this, though.)

The mention about debug info is now in the "Important implementation details" section. I wouldn't put the rest at this point, we may reconsider in future.

README.md
```diff
@@ -87,47 +132,24 @@ once such as:
...
-      * Dead code elimination.
-      * Removal of code dependent on line numbers, file names, etc.
-      * ... and many others.
-  * Comparing programs for semantic equality. Programs are mostly compared
-    instruction-by-instruction.
+## Important implementation details
+
+The core of DiffKemp is the *SimpLL* component, written in C++ for performance
+reasons. The user inputs and comparison results are handled by Python for
+simpliucity. SimpLL performs several important steps:
```
lenticularis39, line 139:
```suggestion
simplicity. SimpLL performs several important steps:
```

lenticularis39 (changes requested):
There is just one typo, otherwise OK :)

lenticularis39 (approved):

PR #252, opened 2022-12-06
Output

PLukas2018:
This pull request contains implementation of new output file `diffkemp-out.yaml` for compare command.
The file will be created always when using compare command without `--stdout` option.
The file will be saved in output directory.

Created file is in yaml format and contains callstacks of non-equal functions and list of functions and types. 
Functions in the list contains their name, file location (relative to snapshot) , line and endline of function, result, list of called functions, list of used types.
Types in the list contains their name, file location, line and endline.


PLukas2018:
Thanks for the feedback.

Here is example of the yaml file (which is shortened and commented):

```yaml
# path to old snapshot
old-snapshot: /home/lukas/BC/diffkemp/kernel-src/SNAPSHOT-80
# path to new snapshot
new-snapshot: /home/lukas/BC/diffkemp/kernel-src/SNAPSHOT-147
# list of functions which are not equal
diff:
  # name of compared function (for example KABI function)
- function: __alloc_workqueue_key
  # list of functions in which difference is found
  diff:
    # name of function which is different
  - inner: worker
    # callstack of name of functions in old snapshot 
    # not including name of compared function
    first-callstack:
    - init_rescuer
    - # ...
    # the same but for new snapshot
    second-callstack:
    - init_rescuer
    - # ...
  - # ..
- # ...
# list of information about "definitions"
names:
  # name of function (dict key for better searching)
  process_one_work:
    # result of function
    result: not equal
    # function/type
    kind: function
    #  information about definition in old snapshot
    first:
      # name in old snapshot
      name: process_one_work
      # relative location of file in which is function/type defined
      file: kernel/workqueue.c
      # start line of function/type
      line: 2046
      # list of called functions
      calls:
        # name of function
      - function: _cond_resched
        # line were it is called
        line: 2179
      - # ... 
      # list of used types
      types:
        # name of type
      - name: worker
        # line where it is used
        line: 2046
      # line where function/type definition ends
      end-line: 2193
    # information about definition in new snapshot
    second:
      # ... similar to first
  # name of type
  worker:
    kind: type
    first:
      name: worker
      file: kernel/workqueue_internal.h
      line: 23
      end-line: 56
    second:
      # ... similar to first
```

diffkemp/diffkemp.py
```diff
@@ -468,7 +470,11 @@ def compare(args):
             old_fun_desc.mod.clean_module()
             new_fun_desc.mod.clean_module()
             LlvmModule.clean_all()
-
+        if output_dir is not None:
+            dump_compare_output(snapshot_dir_old=args.snapshot_dir_old, 
+                                snapshot_dir_new=args.snapshot_dir_new, 
+                                result=result,graph=result_graph,
```
lenticularis39, line 476: Doesn't the `result.graph` already have the graph, making it unnecessary to pass it in another argument?

diffkemp/semdiff/caching.py
```diff
@@ -150,6 +150,44 @@ def compare_vertex_priority(self, other):
                     return True
             return False
 
+        def to_yaml(self):
```
lenticularis39, line 153: I'm not sure if doing conversion to YAML inside ComparisonGraph is a good idea, most of result processing in DiffKemp is done elsewhere (in `function_diff.py` and `diffkemp.py`), the idea being that the `Result` class contains the comparison results. Also, it creates some confusion, since ComparisonGraph's `from_yaml` methods refer to a different YAML format (the one generated by SimpLL).

Wouldn't just extracting the required information from `Result` work? If I understand it correctly, the YAML contains mostly the same information as regular DiffKemp output, hence it should be already (maybe with a few expections) present in `Result`.

lenticularis39 (commented):
The `inner` field in the YAML is unclear, I'd just name it `function`, and perhaps the upper-level function field could be named `symbol`, since it can be also a global variable (IIRC that is not supported currently, but it was there at least in past and might be re-introduced).

The same applies for the `name` field; I'm not sure what to name it, maybe something like `definitions`?

viktormalik:
> ```yaml
> # path to old snapshot
> old-snapshot: /home/lukas/BC/diffkemp/kernel-src/SNAPSHOT-80
> # path to new snapshot
> new-snapshot: /home/lukas/BC/diffkemp/kernel-src/SNAPSHOT-147
> # list of functions which are not equal
> diff:

I'd prefer `results:` as it's in fact the list of compared functions evaluated as non-equal.

> ```yaml
>   # name of compared function (for example KABI function)
> - function: __alloc_workqueue_key
>   # list of functions in which difference is found
>   diff:

`diffs:`

> ```yaml
>     # name of function which is different
>   - inner: worker

`function: worker`

> ```yaml
>     # callstack of name of functions in old snapshot 
>     # not including name of compared function
>     first-callstack:
>     - init_rescuer
>     - # ...
>     # the same but for new snapshot
>     second-callstack:
>     - init_rescuer
>     - # ...
>   - # ..
> - # ...
> # list of information about "definitions"
> names:

`functions:` or `function-defs:`

> ```yaml
>   # name of function (dict key for better searching)
>   process_one_work:
>     # result of function
>     result: not equal
>     # function/type
>     kind: function
>     #  information about definition in old snapshot
>     first:

`old:`? since we use `old-snapshot` above

> ```yaml
>       # name in old snapshot
>       name: process_one_work
>       # relative location of file in which is function/type defined
>       file: kernel/workqueue.c
>       # start line of function/type
>       line: 2046
>       # list of called functions
>       calls:
>         # name of function
>       - function: _cond_resched
>         # line were it is called
>         line: 2179
>       - # ... 
>       # list of used types
>       types:
>         # name of type
>       - name: worker
>         # line where it is used
>         line: 2046
>       # line where function/type definition ends
>       end-line: 2193
>     # information about definition in new snapshot
>     second:
>       # ... similar to first
>   # name of type
>   worker:
>     kind: type
>     first:
>       name: worker
>       file: kernel/workqueue_internal.h
>       line: 23
>       end-line: 56
>     second:
>       # ... similar to first
> ```

Otherwise looks good to me. Also feel free to disagree with any of my suggestions and propose a better naming.

diffkemp/utils.py
```diff
@@ -19,3 +19,35 @@ def get_llvm_version():
...
+        # The end of the function is detected as a line that contains
+        # nothing but an ending curly bracket
+        line_index = start - 1
+        line = lines[line_index]
+        while line.rstrip() not in terminator_list:
+            line_index += 1
+            if line_index == len(lines):
+                raise EndLineError
+            line = lines[line_index]
+        return line_index + 1
```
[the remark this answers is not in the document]
PLukas2018: I think if it returns `line_index` it will return *index* of the line where the function ends. I originally meant it to return a *number* of the line where the function ends. I probably named the function badly.

diffkemp/diffkemp.py
PLukas2018: Thanks, it seems that you are correct.

diffkemp/semdiff/caching.py
PLukas2018: Thanks for the feedback. 

The idea was to create a serialized formatted output with the result of DiffKemp analysis which could be used for  visualization of the results and maybe for other purposes. I am not sure if the format, which I thought up, is the best way for how to do it. I thought that it would be good to print out *call stacks* (only name of functions) of the results (field `diff` in this YAML) and the *call graph* / `ComparisonGraph` (field `names` in this YAML) to not repeat information about functions (which can appear multiple times in call stacks).

I agree that having `from_yaml` and `to_yaml` (which refer to a different format) in the same class (`ComparisonGraph`) is confusing. I am thinking about moving the extraction (conversion to YAML) to a new class (or maybe to `Result` class).

Or maybe I could do it more differently and change the output format - I will try to explain my thoughs.  I could extract more information from `Result` class for creating call stacks (in YAML field `diff`), where I could add `line`, where the function is called. Then from functions which will appear in callstack I could extract from `ComparisonGraph` information about their deffinitions (both names, number of line where function starts, file) - field `names` in YAML. Then the new field `names` would not have to contain fields `calls` and `types`. The new yaml format, after these edits and after renaming some fields, could look something like this.
```
old-snapshot: /home/lukas/BC/diffkemp/kernel-src/SNAPSHOT-80
new-snapshot: /home/lukas/BC/diffkemp/kernel-src/SNAPSHOT-147
# list of functions which are not equal
results:
- function: __alloc_workqueue_key # name of compared function (for example KABI function)
  diffs: # list of functions in which difference is found
  - function: worker  # name of function which is different
    old-callstack: # callstack in old snapshot 
    - name: init_rescuer
      line: 4094 # line where the function is called
    - # ...
    new-callstack:
    - # similar to old-callstack
    - # ...
  - # ..
- # ...

definitions: # list of information about "definitions"
  process_one_work: # name of function/macro/type
    kind: function
    old: #  information about definition in old snapshot
      name: process_one_work
      file: kernel/workqueue.c
      line: 2046
      end-line: 2193
    # information about definition in new snapshot
    new:
      # ... similar to first
  # ...
```
But I am not sure, if the first version of the format where the field `names` represented `ComparisonGraph` would not be more useful in future for some purposes.

I could also remove the field `definitions` (`names`) completely and move these information to functions in call stack  (field `results`), but then i think lot of information would repeat and also I think it is useful to have definitions separately to know which pair of functions belongs together (if I am not mistaken they can have different names caused for example by renaming of funcion).

diffkemp/utils.py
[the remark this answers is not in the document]
viktormalik: Ah, right. Ok, since it's a utility function which may possibly be used elsewhere, it's reasonable to stick with line numbers. Maybe it'd be best to use line numbers also in the `syntax_diff` function and just use `lines[line - 1]` to access them.

diffkemp/semdiff/caching.py
viktormalik: > I agree that having `from_yaml` and `to_yaml` (which refer to a different format) in the same class (`ComparisonGraph`) is confusing. I am thinking about moving the extraction (conversion to YAML) to a new class (or maybe to `Result` class).

Agreed, moving it to `Results` sounds reasonable to me.

> [...]
> But I am not sure, if the first version of the format where the field `names` represented `ComparisonGraph` would not be more useful in future for some purposes.

I kind of like this new format better. It seems more compact and human-readable (though that's not its purpose for sure). There'll be some minor repetition (`line` in callstacks if there are mutliple same callstacks) but I'm ok with that. I'd suggest going with this one, it looks pretty extensible for future needs. Btw. shouldn't callstack entries contain also the file name (so that we don't have to pull that from definitions)?
 
> I could also remove the field `definitions` (`names`) completely and move these information to functions in call stack (field `results`), but then i think lot of information would repeat and also I think it is useful to have definitions separately to know which pair of functions belongs together (if I am not mistaken they can have different names caused for example by renaming of function).

Agreed, let's keep the definitions separated, they may come handy in future (in addition to the advantages that you listed).

PLukas2018:
I tried to edit some parts according to your comments and I tried to implement the new format of the YAML output discussed above. I made small changes in the format: entry `file` is included in call stacks and entry `name` in definitions is included only in the entry `new` and only if the names are different between versions. I moved the YAML creation to a class because I tried to split it to more smaller functions. You are welcome to review the code.

diffkemp/semdiff/caching.py
```diff
@@ -440,21 +440,20 @@ def mark_uncachable_from_assumed_equal(self):
...
-    def _edge_callstack_to_string(self, callstack):
-        """Converts a callstack consisting of Edge objects to a string
-        representation."""
-        return "\n".join(["{} at {}:{}".format(call.target_name,
-                                               call.filename,
-                                               call.line)
-                          for call in callstack])
-
-    def _yaml_callstack_to_string(self, callstack):
+    def _edge_callstack_to_dict(self, callstack):
```
zacikpa, line 443: Maybe this should be a static method if it stays here. Another option would be to create a `Callstack` class which would hold the dictionary and can be constructed from `Edge` objects.

diffkemp/semdiff/caching.py
```diff
@@ -440,21 +440,20 @@ def mark_uncachable_from_assumed_equal(self):
...
+        """Converts a callstack consisting of Edge objects to a representation
+        as a list of dictionaries with keys: name, file, line."""
+        return [{"name": call.target_name, "file": call.filename,
+                 "line": call.line}
+                for call in callstack]
+
+    def _yaml_callstack_to_dict(self, callstack):
```
zacikpa, line 450: Same as before.

diffkemp/semdiff/result.py
```diff
@@ -53,6 +56,15 @@ def __init__(self, name, filename=None, line=None, callstack=None,
             self.diff_kind = diff_kind
             self.covered = covered
 
+        def callstack_to_string(self):
```
zacikpa, line 59: I think the naming of this function may be bit confusing, since it does not convert the callstack to a string in-place, but rather returns its string representation. Something like `callstack_as_string` or `get_callstack_string` would be more clear.

zacikpa (commented):
I tried running `diffkemp compare` on some snapshots and everything seems to work well. The new output looks very good. I only have a couple of nitpicks.

diffkemp/diffkemp.py
```diff
@@ -580,3 +587,164 @@ def text_indent(text, width):
...
+
+    def _get_results(self, function_names, type_names, macro_names):
+        """Returns call stacks of not-equal functions.
+        Parameters are output arguments to save names of symbols that appear
+        in call stacks.
+        """
+        # list of all not-equal functions callstacks
+        results = []
+        for fun_name, fun_result in self.result.inner.items():
+            if fun_result.kind == Result.Kind.NOT_EQUAL:
```
viktormalik, line 633:
```suggestion
            if fun_result.kind != Result.Kind.NOT_EQUAL:
                continue
```
to decrease indentation of the following code

diffkemp/diffkemp.py
```diff
@@ -580,3 +587,164 @@ def text_indent(text, width):
...
+                function_names.add(fun_name)
+                # list of not-equal functions callstacks for a function
+                diffs = []
+                for called_res in sorted(fun_result.inner.values(),
+                                         key=lambda r: r.first.name):
+                    # information about a function which is different
+                    diff = {}
+                    diff["function"] = called_res.first.name
+                    # getting callstacks
+                    for entity in [called_res.first, called_res.second]:
```
viktormalik, line 643: Could you extract the below code into a function (something like `_get_callstack`) and then do
```python
diff["callstack-old"] = self._get_callstack(...)
diff["callstack-new"] = self._get_callstack(...)
```
Maybe if you create a separate `Callstack` class (as mentioned elsewhere), it could be a part of that (something like `Callstack.to_yaml()`).

viktormalik (commented):
The output looks good to me, just a couple of code improvement suggestions.

diffkemp/semdiff/caching.py
PLukas2018: Thanks. I have created the class `Callstack` with a field `calls`  which is the list of calls. I have also changed the method `_edge_callstack_to_dict` to class method of the class. I have also created a method for getting string representation and YAML representation of the call stack which accepts path of snapshot for removing the snapshot prefix of file paths.

diffkemp/semdiff/result.py
PLukas2018: Thanks. I have created `__str__` method in `Callstack` class for getting string representation of the call stack, I have also created a method `as_str_with_rel_paths` which will return string representation of the call stack with removed prefix of file paths.

diffkemp/syndiff/function_syntax_diff.py
```diff
@@ -15,37 +16,33 @@ def syntax_diff(first_file, second_file, name, kind, first_line, second_line):
     command = ["diff", "-C", "1", os.path.join(tmpdir, "1"),
                os.path.join(tmpdir, "2")]
 
-    if kind == "function":
-        terminator_list = ["}", ");"]
-    elif kind == "type":
-        terminator_list = ["};"]
+    DIFF_NOT_OBTAINED_MESSAGE = "  [could not obtain diff]\n"
```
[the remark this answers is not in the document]
PLukas2018: Thanks. Done.

diffkemp/utils.py
```diff
@@ -19,3 +19,37 @@ def get_llvm_version():
     """
     return int(subprocess.check_output(
         ["llvm-config", "--version"]).decode().rstrip().split(".")[0])
+
+
+class EndLineNotFound(Exception):
+    """
+    Error to inform that end line of function / type was not found.
+    """
+
```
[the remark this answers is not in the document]
PLukas2018: Thanks, yes it was a superfluous empty line. I have removed it.

diffkemp/diffkemp.py
```diff
@@ -580,3 +587,164 @@ def text_indent(text, width):
             if not output_dir:
                 output.write(text_indent("}}}\n", indent - 2))
             output.write("\n")
+
+
+class YamlOutput:
```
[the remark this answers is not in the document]
PLukas2018: For now i have moved it to file `output.py` .

diffkemp/diffkemp.py
```diff
@@ -580,3 +587,164 @@ def text_indent(text, width):
...
+        self.output["new-snapshot"] = self.new_dir
+
+        # Sets with symbol names
+        function_names = set()
+        macro_names = set()
+        # Note: type_names contains tuples (type name, parent name)
+        type_names = set()
+
+        self.output["results"] = self._get_results(function_names, type_names,
+                                                   macro_names)
```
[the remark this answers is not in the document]
PLukas2018: Thanks, I was thinking of making `function_names`, `type_names`, and `macro_names` fields of the class, but I was unsure what is better approach. I have changed it to fields of the class.
Also thanks for the advice about returning values, I have changed it.

diffkemp/diffkemp.py
PLukas2018: Thanks, that is clever.

diffkemp/diffkemp.py
PLukas2018: Thanks, I have created the `Callstack` class and moved getting of call stack to method `to_output_yaml_with_rel_path` and getting of "symbols" (from call stack for getting definitions) to method `get_symbol_names`.

diffkemp/diffkemp.py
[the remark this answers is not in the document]
lenticularis39: > I'd prefer moving this class into a separate module (source file), e.g. `output.py`.
> 
> Another option would be to transform this into a `Result.to_yaml()` method but I'm not sure if it would logically fit that. @lenticularis39 WDYT?

I think it makes sense to have it in `output.py`, we also have other formatting code like `print_syntax_diff` that can be moved into there.

lenticularis39 (changes requested):
I spotted a few typos, otherwise it looks good to me, I like the `Callstack` class and the idea of having output code to a separate file.

lenticularis39:
Now with the code organized into classes `YamlOutput` and `Callstack`, it would also make sense to add some unit tests.

diffkemp/semdiff/result.py
```diff
@@ -44,6 +45,94 @@ class Entity:
         or a parameter.
         If it is a function, it contains the file of the function.
         """
+        class Callstack:
```
viktormalik, line 48: Don't nest classes this much. Move `Callstack` at least one level up (i.e. directly inside `Result`).

diffkemp/semdiff/result.py
```diff
@@ -44,6 +45,94 @@ class Entity:
...
+
+            @classmethod
+            def from_simpll_yaml(cls, callstack):
+                """Creates a Callstack from a YAML representation of
+                a callstack (used in non-fun) diffs."""
+                return cls([{"name": call["function"], "file": call["file"],
+                             "line": call["line"]}
+                           for call in callstack])
+
+            def as_str_with_rel_paths(self, startpath):
```
viktormalik, line 76:
```suggestion
            def as_str_with_rel_paths(self, prefix):
```

diffkemp/semdiff/result.py
```diff
@@ -44,6 +45,94 @@ class Entity:
...
+                """Returns callstack as string with relative paths
+                to startpath."""
+                return str(self).replace(
+                    os.path.join(os.path.abspath(startpath), ""), "")
+
+            def to_output_yaml_with_rel_path(self, startpath):
```
viktormalik, line 82:
```suggestion
            def to_output_yaml_with_rel_path(self, prefix):
```

diffkemp/output.py
```diff
@@ -0,0 +1,141 @@
...
+import os
+import yaml
+
+
+class YamlOutput:
+    """Yaml representation of compare result."""
+
+    def __init__(self, snapshot_dir_old, snapshot_dir_new, result):
+        self.old_dir = os.path.join(
+            os.path.abspath(snapshot_dir_old), "")
```
viktormalik, line 13: Isn't `snapshot_dir_old` already an abspath? I think that `YamlOutput` shouldn't do the path transformation, it should be the caller's responsibility.

diffkemp/semdiff/result.py
```diff
@@ -44,6 +45,94 @@ class Entity:
...
+            def get_symbol_names(self, fun_name):
+                """
+                Returns tree sets of symbol names which appears in callstacks,
+                names does not contain kind.
+                First set contains function names, second macro names
+                and third type names.
+                :param fun_name: Name of function where the callstack starts.
+
+                Note: type names contains tuples (type name, name of function
+                in which it is used)
```
viktormalik, line 102: The comment is quite hard to understand. What about something like this:
```suggestion
                Returns a tuple containing three sets of symbol names which appear 
                in callstacks:
                - function names (set of strings)
                - macro names (set of strings)
                - type names (set of string pairs (type name, user function name))
                      
                :param fun_name: Name of the first callstack function.
```

diffkemp/output.py
```diff
@@ -0,0 +1,141 @@
...
+                    .callstack.to_output_yaml_with_rel_path(self.old_dir) \
+                    if called_res.first.callstack else []
+                diff["new-callstack"] = called_res.second \
+                    .callstack.to_output_yaml_with_rel_path(self.new_dir) \
+                    if called_res.second.callstack else []
+                # updating symbol names
+                if called_res.first.callstack:
+                    function_names, macro_names, type_names = called_res \
+                        .first.callstack.get_symbol_names(fun_name)
+                self.function_names.update(function_names)
```
zacikpa, line 64: This fails if the condition in the previous `if` statement is False, because no local variable  `function_names` is defined. Was it meant to be indented?

diffkemp/semdiff/result.py
```diff
@@ -44,6 +45,94 @@ class Entity:
...
+                """Creates a Callstack from callstack consisting
+                of Edge objects."""
+                return cls([{"name": call.target_name, "file": call.filename,
+                             "line": call.line}
+                            for call in callstack])
+
+            @classmethod
+            def from_simpll_yaml(cls, callstack):
+                """Creates a Callstack from a YAML representation of
+                a callstack (used in non-fun) diffs."""
```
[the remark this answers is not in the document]
PLukas2018: Thanks.

diffkemp/semdiff/result.py
PLukas2018: Thanks, I moved it to the `Result` class.

diffkemp/output.py
PLukas2018: Thanks. The received `snapshot_dir_old` was a relative path (argument of the program). I moved the path transformation to the caller function (`compare`).

I also moved path transformation in the `Callstack` class in methods `as_str_with_rel_paths` and `to_output_yaml_with_rel_path`. In the case of  `as_str_with_rel_paths` I moved it to the caller function (`print_syntax_diff`), in the case of `to_output_yaml_with_rel_path` I removed it completely because the received path was already absolute.

diffkemp/semdiff/result.py
PLukas2018: Thanks, I used most of your suggestion, but 
- for type names I edited the description of the second  of the pair to `name of function in which the type is used) `
- for param description I edited it to `Name of compared function`.
I have also renamed the param from `fun_name` to `compared_fun`, I think it better describe the param.

PLukas2018:
I attempted to write tests for the `Callstack` and `YamlOutput` classes. I will gladly receive some feedback.

For now, I have put the tests in to file `caching_test.py` but I assume that they should be in own file (I am thinking about `result_test.py` because they work with the `Result` class (`YamlOutput`) or with part of it - `Callstack`. Some of the tests use fixture `graph` so I am thinking to move the fixture to file `conftest.py` to be able to share/use it in multiple files or maybe I could write a new graph for these tests.

diffkemp/diffkemp.py
```diff
@@ -468,7 +469,13 @@ def compare(args):
             old_fun_desc.mod.clean_module()
             new_fun_desc.mod.clean_module()
             LlvmModule.clean_all()
-
+    # Create yaml output
+    if output_dir is not None and os.path.isdir(output_dir):
+        old_dir_abs = os.path.join(os.path.abspath(args.snapshot_dir_old), "")
+        new_dir_abs = os.path.join(os.path.abspath(args.snapshot_dir_new), "")
```
viktormalik, line 475: Why the `join` calls?

diffkemp/diffkemp.py
```diff
@@ -468,7 +469,13 @@ def compare(args):
...
+        YamlOutput(snapshot_dir_old=old_dir_abs, snapshot_dir_new=new_dir_abs,
+                   result=result).save(output_dir=output_dir,
+                                       file_name="diffkemp-out.yaml")
```
viktormalik, line 478: Nit: this is quite a complex command, could we split into two commands with an intermediate variable?

diffkemp/diffkemp.py
```diff
@@ -517,6 +524,9 @@ def text_indent(text, width):
         """Indent each line in the text by a number of spaces given by width"""
         return ''.join(" " * width + line for line in text.splitlines(True))
 
+    old_dir_abs_path = os.path.join(os.path.abspath(snapshot_dir_old), "")
+    new_dir_abs_path = os.path.join(os.path.abspath(snapshot_dir_new), "")
```
viktormalik, line 528: Same as above - are the `join` calls necessary?

tests/unit_tests/caching_test.py
```diff
@@ -243,31 +245,32 @@ def test_graph_to_fun_pair_list(graph):
...
         assert not macro.covered
         assert struct_file.filename == "include/file.h"
         assert struct_file.line == 121
-        assert struct_file.callstack == ("do_check at app/main.c:58\n"
-                                         "struct file (type) at "
-                                         "include/file.h:121")
+        assert str(struct_file.callstack) == (
+            "do_check at app/main.c:58\n"
+            "struct_file (type) at "
+            "include/file.h:121")
```
viktormalik, line 273: should fit one line now:
```suggestion
            "struct_file (type) at include/file.h:121")
```

viktormalik (commented):
Overall lgtm, just a couple of nits.

I agree that the tests should be in a separate file but considering that the PR is taking long time already, I suggest merging it in this state and doing the split later in a separate PR.

diffkemp/syndiff/function_syntax_diff.py
```diff
@@ -5,47 +5,44 @@
...
-            line_index = start - 1
-            line = lines[line_index]
-            while line.rstrip() not in terminator_list:
-                line_index += 1
-                if line_index == len(lines):
-                    return "Error: cannot get diff\n"
+                return DIFF_NOT_OBTAINED_MESSAGE
+
+            line_number = start
+            while line_number <= end:
```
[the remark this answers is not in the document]
PLukas2018: Thanks, sure it makes sense.

diffkemp/semdiff/result.py
```diff
@@ -38,12 +38,99 @@ def from_string(string):
...
+                                                   call["line"])
+                             for call in self.calls])
+
     class Entity:
         """
         Compared entity information. This can be e.g. a function, a module,
         or a parameter.
         If it is a function, it contains the file of the function.
         """
+
```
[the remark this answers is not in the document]
PLukas2018: Thanks, I removed it.

diffkemp/diffkemp.py
PLukas2018: Described below.

diffkemp/diffkemp.py
PLukas2018: Sure, done.

diffkemp/diffkemp.py
PLukas2018: I looked into that and it looks like paths of functions and macros are saved as absolute paths and paths of types are saved as relative paths. That is the reason why the function `replace` is used in call stacks (to get always a relative path even if the path is already relative).  
The `join` is "necessary" for adding `/` to the end of the snapshot path  (without that using `replace` on calls with abs path would return "relative path" starting with backslash `/`). I guess it could be done differently (like just adding a backslash to the snapshot path) but I used `join` because it was used like that before.

tests/unit_tests/caching_test.py
PLukas2018: Thanks, edited.

PLukas2018:
Sorry, I forgot to update this PR. I already split the tests using `conftest` file (hopefully there will be no problems with it). 
Also, I add skipping of empty diffs in `YamlOutput` class because I noticed it was done in the original output.

I updated the code according to your comments. I cleaned up the commits and rebased it on master. Hopefully, there will be no big problems anymore and the PR could be soon closed (merged). 

Also I compared one pair of kernels to check if there is a problem, it looks like there is no visible problem...

diffkemp/diffkemp.py
viktormalik: Makes sense, thanks.

lenticularis39 (approved):

viktormalik (approved):
Thanks for all the updates of this PR, LGTM! Also nice cleanup of the commits :-)

PR #255, opened 2023-03-11
Add a script that cleans up the environment after upgrading LLVM

zacikpa:
When upgrading LLVM,  the `.ll` files in `kernel/` and `tests/regression/kernel_modules/` become deprecated. Similarly, `diffkemp/simpll/_simpll.abi3.so` has to be replaced. Up to now, we were not able to detect the upgrade and rebuild them automatically.

I propose to execute a simple script during CMake configuration. The script reads the LLVM version from `diffkemp/simpll/_simpll.abi3.so` (i.e. the version of LLVM against which `simpll` is linked) and the current system LLVM version. If they differ, all of the mentioned files are deleted.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,4 +29,8 @@ if (${BUILD_LLREVE})
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/llreve/reve)
 endif ()
 
+execute_process(COMMAND
+  ${CMAKE_CURRENT_SOURCE_DIR}/tools/simpll_clean.sh
+  ${CMAKE_CURRENT_SOURCE_DIR})
+
 add_subdirectory(tests/unit_tests/simpll)
--- a/tools/simpll_clean.sh
+++ b/tools/simpll_clean.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+# A script which deletes the SimpLL shared library if it's linked against
+# a different LLVM version than the one currently installed.
+#
+# Usage:
+# ./simpll_clean.sh [DIFFKEMP_ROOT]
+#
+# When DIFFKEMP_ROOT is not provided, the script assumes that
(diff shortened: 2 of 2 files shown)


tools/llvm_clean.sh
```diff
@@ -0,0 +1,31 @@
...
+# the current working directory is the DiffKemp root directory.
+
+if [[ -z $1 ]]
+    then
+        DIFFKEMP_ROOT=$PWD
+    else
+        DIFFKEMP_ROOT=$1
+fi
+
+if [[ ! -f ${DIFFKEMP_ROOT}/diffkemp/simpll/_simpll.abi3.so ]]
```
lenticularis39, line 17: There might be an issue with the tool not printing anything when the path is wrong, e.g. because of a typo, since there will be no indication of an error in that case.

It might make sense to check for the existence of e.g. `${DIFFKEMP_ROOT}/diffkemp/simpll` first, and exit with an error if it doesn't exist.

zacikpa: Thanks, that's a good idea! It should be fixed now.

viktormalik:
I'm working on tests refactoring which will replace `tests/regression/kernel_modules` by multiple per-LLVM-version directories checked out into the tree. Also, the sources for tests will be required to be placed in `kernel/`.

That being said, I like this PR but would limit it to removing `_simpll.abi3.so`. In fact, that file is a part of a DiffKemp "build", so it makes sense to call this script from CMake (as opposed to `kernel` and `tests/regression/kernel_modules`).

zacikpa:
> That being said, I like this PR but would limit it to removing _simpll.abi3.so. In fact, that file is a part of a DiffKemp "build", so it makes sense to call this script from CMake (as opposed to kernel and tests/regression/kernel_modules).

Ok, that makes sense. Done.

viktormalik (approved):

PR #256, opened 2023-04-25
Comparison logger

xkucma00:
Implements logger, simplifying the way debug logging is done and adding conditional logging to comparisons in DifferentialFunctionComparator. For this kind of logging, the logger takes into the account the hierarchy of comparisons, logging only messages regarding comparisons that eventually lead to failure of instruction comparisons or those that are forced to log using higher debug level option.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -15,6 +15,7 @@
 #include "DifferentialFunctionComparator.h"
 #include "Config.h"
 #include "FieldAccessUtils.h"
+#include "Logger.h"
 #include "SourceCodeUtils.h"
 #include "passes/FunctionAbstractionsGenerator.h"
 #include <deque>
@@ -80,11 +81,13 @@ int DifferentialFunctionComparator::cmpValuesByMapping(const Value *L,
 /// Uses data saved in StructFieldNames.
 int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
                                             const GEPOperator *GEPR) const {
+    PREP_LOG("GEP", GEPL, GEPR);
+
     int OriginalResult = FunctionComparator::cmpGEPs(GEPL, GEPR);
 
     if (OriginalResult == 0)
         // The original function says the GEPs are equal - return the value
-        return OriginalResult;
+        RETURN_WITH_LOG(OriginalResult);
 
     if (isa<ArrayType>(GEPL->getSourceElementType())
         && isa<ArrayType>(GEPR->getSourceElementType())) {
@@ -93,57 +96,57 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
 
         if (GEPL->getNumOperands() != 3 || GEPR->getNumOperands() != 3)
             // We only handle GEPs that access exactly one array element
-            return OriginalResult;
+            RETURN_WITH_LOG(OriginalResult);
 
         if (GEPL->getSourceElementType()->getArrayNumElements()
             == GEPR->getSourceElementType()->getArrayNumElements())
-            return OriginalResult;
+            RETURN_WITH_LOG(OriginalResult);
 
         auto *STyL = cast<ArrayType>(GEPL->getSourceElementType());
         auto *STyR = cast<ArrayType>(GEPR->getSourceElementType());
         if (int Res = cmpTypes(STyL->getElementType(), STyR->getElementType()))
             // The array element type must be the same
-            return Res;
+            RETURN_WITH_LOG(Res);
 
         for (unsigned i = 0, e = GEPL->getNumOperands(); i != e; ++i) {
             if (int Res = cmpValues(GEPL->getOperand(i), GEPR->getOperand(i)))
-                return Res;
+                RETURN_WITH_LOG(Res);
         }
 
         auto TypeL = getVariableTypeInfo(GEPL->getOperand(2));
         auto TypeR = getVariableTypeInfo(GEPR->getOperand(2));
 
         if (!TypeL || !TypeR)
-            return OriginalResult;
+            RETURN_WITH_LOG(OriginalResult);
 
         if (TypeL->getTag() == dwarf::DW_TAG_enumeration_type
             && TypeR->getTag() == dwarf::DW_TAG_enumeration_type)
-            return 0;
+            RETURN_WITH_LOG(0);
 
-        return OriginalResult;
+        RETURN_WITH_LOG(OriginalResult);
     }
 
     if (!isa<StructType>(GEPL->getSourceElementType())
         || !isa<StructType>(GEPR->getSourceElementType()))
         // One of the types in not a structure - the original function is
         // sufficient for correct comparison
-        return OriginalResult;
+        RETURN_WITH_LOG(OriginalResult);
 
     if (getStructTypeName(dyn_cast<StructType>(GEPL->getSourceElementType()))
         != getStructTypeName(
                 dyn_cast<StructType>(GEPR->getSourceElementType())))
         // Different structure names - the indices may be same by coincidence,
         // therefore index comparison can't be used
-        return OriginalResult;
+        RETURN_WITH_LOG(OriginalResult);
 
     unsigned int ASL = GEPL->getPointerAddressSpace();
     unsigned int ASR = GEPR->getPointerAddressSpace();
 
     if (int Res = cmpNumbers(ASL, ASR))
-        return Res;
+        RETURN_WITH_LOG(Res);
 
     if (int Res = cmpNumbers(GEPL->getNumIndices(), GEPR->getNumIndices()))
-        return Res;
+        RETURN_WITH_LOG(Res);
 
     if (GEPL->hasAllConstantIndices() && GEPR->hasAllConstantIndices()) {
         std::vector<Value *> IndicesL;
@@ -153,7 +156,7 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
         const GetElementPtrInst *GEPiR = dyn_cast<GetElementPtrInst>(GEPR);
 
         if (!GEPiL || !GEPiR)
-            return OriginalResult;
+            RETURN_WITH_LOG(OriginalResult);
 
         for (auto idxL = GEPL->idx_begin(), idxR = GEPR->idx_begin();
              idxL != GEPL->idx_end() && idxR != GEPR->idx_end();
@@ -171,7 +174,7 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
                 // If the indexed type is not a structure type, the indices have
                 // to match in order for the instructions to be equivalent
                 if (int Res = cmpValues(idxL->get(), idxR->get()))
-                    return Res;
+                    RETURN_WITH_LOG(Res);
 
                 IndicesL.push_back(*idxL);
                 IndicesR.push_back(*idxR);
@@ -193,7 +196,7 @@ int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
                 || MemberNameR == DI->StructFieldNames.end()
                 || !MemberNameL->second.equals(MemberNameR->second))
                 if (int Res = cmpValues(idxL->get(), idxR->get()))
-                    return Res;
+                    RETURN_WITH_LOG(Res);
 
             IndicesL.push_back(*idxL);
             IndicesR.push_back(*idxR);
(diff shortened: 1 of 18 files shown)


viktormalik (commented):
Overall, this looks great! I did a first pass and have a couple of suggestions to improve usage.

Also, I think that we should document proper usage of the logger somewhere, especially the part with `PREP_LOG` and `RETURN_WITH_LOG(_FORCE)` is not very clear to me. Could you perhaps add a comment on that somewhere (`Logger.h`?) or at least explain it here in the PR so that we can discuss if it can be improved somehow? Thanks!

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -80,11 +81,14 @@ int DifferentialFunctionComparator::cmpValuesByMapping(const Value *L,
 /// Uses data saved in StructFieldNames.
 int DifferentialFunctionComparator::cmpGEPs(const GEPOperator *GEPL,
                                             const GEPOperator *GEPR) const {
+    PREP_LOG(<< getDebugIndent() << "L GEP: " << *GEPL << "\n"
+             << getDebugIndent() << "R GEP: " << *GEPR << "\n");
```
viktormalik, line 85: I'm wondering if we can get rid of the leading `<< getDebugIndent() <<` as in `LOG`. I assume that doing:
```
PREP_LOG("L GEP: " << *GEPL << "\n");
PREP_LOG("R GEP: " << *GEPR << "\n");
```
is not feasible b/c `prep_context` would be called twice.
Could we achieve this?
```
LOG_PREP("L GEP: " << *GEPL << "\n", "R GEP: " << *GEPR << "\n");
```
or maybe this?
```
LOG_PREP("L GEP: " << *GEPL << "\n" << "R GEP: " << *GEPR << "\n")
```
and the macro would magically apply the indent at the beginning of each line?

Just throwing ideas to make the usage simpler, feel free to comment them.

viktormalik (commented):
I'm still a bit confused about the difference between `RETURN_WITH_LOG` and `RETURN_WITH_LOG_FORCE`. Could you maybe explain the difference (especially when is each of them expcted to be used)? You can just explain it here in the PR and we can then reflect it in the code if necessary. Thanks!

diffkemp/simpll/Logger.h
```diff
@@ -0,0 +1,113 @@
...
+
+  protected:
+    void increaseDebugIndentLevel();
+    void decreaseDebugIndentLevel();
+
+  private:
+    // current level of indentation within the logger
+    size_t level{0};
+    // debug message buffer (force-kept, indent level, message)
+    std::vector<std::tuple<bool, size_t, std::string>> buffer{};
```
lenticularis39, line 104: I suggest using a struct type whenever you need to document the meaning of the fields of a tuple, e.g. from seeing `if (!iter->force_kept)` (in Logger.cpp) you immediately know what the if condition is checking, unlike `if (!std::get<bool>(*iter))`.

diffkemp/simpll/Logger.h
```diff
@@ -0,0 +1,113 @@
...
+    llvm::raw_ostream &prep_log();
+    // prepare for logging messages within the context of the last prepared
+    // message
+    void prep_context();
+    // log a prepared message
+    void log(bool keep = true, const char *force_keep_type = nullptr);
+    // dump all messages from the buffer
+    void dump();
+    // reset the logger indent level to 0 and return the original value
+    size_t storeIndent();
```
lenticularis39, line 92: The function doesn't _store_ anything, it just resets the ident level and returns the previous value (as explained in the comment). It is used to store the value (which is not clear without seeing the rest of the code), but that has nothing to do with the function itself.

Hence I believe the name should be different, the name `storeIndent` seems to imply that something is stored inside the class.

diffkemp/simpll/Logger.h
```diff
@@ -0,0 +1,113 @@
...
+// logging its children. Must be later followed by LOG_KEEP or RETURN_WITH_LOG
+// (or their _FORCE variants)
+#define PREP_LOG(msg)                                                          \
+    DEBUG_WITH_TYPE(LOGGER_BASE_LEVEL, logger.prep_log() msg;                  \
+                    logger.prep_context();)
+
+// Based on given value (interpreted as bool), either keep or erase prepared
+// message and its children. Exectuing keep on a lowest level message leads to
+// automatically printing it and all its children, clearing the message buffer.
+#define LOG_KEEP(keep) DEBUG_WITH_TYPE(LOGGER_BASE_LEVEL, logger.log(x))
```
lenticularis39, line 35: Is the `keep` argument here actually used somewhere? It looks to me that it is not, and `x`, which is passed as the `keep` value, is just passed from the context of the parent macro, in that case the argument would be redundant.

lenticularis39 (commented):
Just a few things I noticed so far, I haven't finished looking at the logic behind the buffer and the keeping of/discarding the messages yet.

diffkemp/simpll/Logger.h
xkucma00: Thanks for pointing that out, I meant to change the x to keep for this macro. It's meant to have it as an argument, so it can potentially be used instead of RETURN_WITH_LOG in cases where returning the given value isn't desired.

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -26,18 +27,15 @@
...
+        << "\" and \"" << SecondFun->getName() << "\" { ";
+        increaseDebugIndentLevel(););
     ComparedFuns.emplace(std::make_pair(FirstFun, SecondFun),
                          Result(FirstFun, SecondFun));
 
     // Check if the functions is in the ignored list.
     if (ResCache.isFunctionPairCached(FirstFun, SecondFun)) {
-        DEBUG_WITH_TYPE(DEBUG_SIMPLL,
-                        dbgs() << getDebugIndent() << "ignored }\n");
+        LOG(<< getDebugIndent() << "ignored }\n");
```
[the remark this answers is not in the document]
xkucma00: Makes sense, I ended up adding the indentation to LOG macros by default, but also implementing *_NO_INDENT variants for cases where it was not previously used.

diffkemp/simpll/DifferentialFunctionComparator.cpp
xkucma00: I've opted for the last option and added a function which automatically adds debug indent to each newline (except for the last one). The current implementation is a bit non-optimal in the way it handles the strings, if that's a problem I could re-do it a better way. 

I'd also like to just point out one small thing, including the leading << in the macro and not in the macro call leads to ugly alignment of the call in the code. I can't cite the source code from an outdated comment but basically this:

```
    PREP_LOG("L GEP: " << *GEPL << "\n"
                       << "R GEP: " << *GEPR << "\n");
```

xkucma00:
Thanks for the feedback, hopefully most/all of it should be addressed now. I'd like to ask one question in regards to GEP (and potentially other comparisons, like FieldAccess or Alignment) comparisons and logging. Currently the logger prints the arguments of the comparison, eg left and right GEP instruction, which basically just prints the instruction with the context that now it's being compared as GEP. Is that alright, or should this output be somehow specialized and display some specific information?

lenticularis39:
I don't think there is any more special information to display, the context can be always found within the LLVM IR files themselves, the key information is where the difference is. Recognizing why the instruction was compared as different is something that would be a nice feature, not only for debug logging, but possibly also for difference reporting, but that is an extra feature for a separate PR, if you or someone else wants to implement it some time in the future.

xkucma00:
I committed new changes, I hope the comments are clearer now.
In my case, the code seems to work both with and without `--disable-simpll-ffi`, though I'm using the development docker container if that's relevant. 
I haven't implemented storing the pointers instead of the converted strings yet, as that'll require bigger changes, but I'll be working on it.

lenticularis39:
> I committed new changes, I hope the comments are clearer now.

Looks good now, thanks.

>  In my case, the code seems to work both with and without `--disable-simpll-ffi`, though I'm using the development docker container if that's relevant.

Good to know, if it works in the DiffKemp development container, we can postpone the issue for after merging.

> I haven't implemented storing the pointers instead of the converted strings yet, as that'll require bigger changes, but I'll be working on it.

OK, of course having at least slow comparison logging is better than no comparison logging, I just wanted to point out that it is really slow, at least on my system:
```
$ time build/diffkemp/simpll/diffkemp-simpll snapshots/linux-3.10.0-862.el7/net/core/skbuff.ll snapshots/linux-3.10.0-957.el7/net/core/skbuff.ll --print-callstacks --fun ___pskb_trim --cache-dir /tmp/tmpytezidqh --verbosity 2
...
real    0m0,475s
user    0m0,432s
sys     0m0,036s
$ time build/diffkemp/simpll/diffkemp-simpll snapshots/linux-3.10.0-862.el7/net/core/skbuff.ll snapshots/linux-3.10.0-957.el7/net/core/skbuff.ll --print-callstacks --fun ___pskb_trim --cache-dir /tmp/tmpytezidqh --verbosity 2
...
real    0m34,030s
user    0m29,749s
sys     0m4,161s
```

lenticularis39 (approved):

diffkemp/simpll/DifferentialFunctionComparator.cpp
viktormalik: Ok, this is annoying but for the sake of consistency (avoiding leading `<<`), I guess that we don't have another option. In addition, since we'll eventually store just the pointers, this will likely change.

viktormalik:
Looks great!

I just have a couple of comments:
- IIUC, the difference between `RETURN_WITH_LOG` and `RETURN_WITH_LOG_FORCE` is that the latter logs the value also for an equality result, if the most verbose logging level is set. I'm not sure if the `_FORCE` suffix sufficiently captures that. What about swapping them around and naming them `RETURN_WITH_LOG_NEQ` and `RETURN_WITH_LOG`, respectively? The current `_FORCE` variant (i.e. the new non-suffix variant) seems to be used more often anyways.
- What is important in the end is that whatever naming we use, the usage is properly documented. So, I would love to see somewhere (header of  `Logger.h`?) a comment with a recommended way of using these (and maybe also other user-oriented) macros.
- Similarly, I would like to see a documentation on what is (and should be) logged for individual verbose/debug levels so that future developers know what to log under which level when adding new features. Maybe having this in one place with the previous point would make sense.

viktormalik:
> OK, of course having at least slow comparison logging is better than no comparison logging, I just wanted to point out that it is really slow, at least on my system:
> 
> ```
> $ time build/diffkemp/simpll/diffkemp-simpll snapshots/linux-3.10.0-862.el7/net/core/skbuff.ll snapshots/linux-3.10.0-957.el7/net/core/skbuff.ll --print-callstacks --fun ___pskb_trim --cache-dir /tmp/tmpytezidqh --verbosity 2
> ...
> real    0m0,475s
> user    0m0,432s
> sys     0m0,036s
> $ time build/diffkemp/simpll/diffkemp-simpll snapshots/linux-3.10.0-862.el7/net/core/skbuff.ll snapshots/linux-3.10.0-957.el7/net/core/skbuff.ll --print-callstacks --fun ___pskb_trim --cache-dir /tmp/tmpytezidqh --verbosity 2
> ...
> real    0m34,030s
> user    0m29,749s
> sys     0m4,161s
> ```

This is a significant performance drop. I'm ok merging this for now as long as we try to fix it as soon as possible. 

@zacikpa you're probably the one who uses the debug messages the most these days. How much pain would this bring you?

diffkemp/simpll/Logger.h
```diff
@@ -0,0 +1,149 @@
...
+#define LOG_INDENT() DEBUG_WITH_TYPE(DEBUG_SIMPLL, increaseDebugIndentLevel())
+#define LOG_UNINDENT() DEBUG_WITH_TYPE(DEBUG_SIMPLL, decreaseDebugIndentLevel())
+
+#define WITHOUT_LOGGER_INDENT(actions)                                         \
+    do {                                                                       \
+        size_t indent;                                                         \
+        DEBUG_WITH_TYPE(LOGGER_BASE_LEVEL, indent = logger.resetIndent());     \
+        actions;                                                               \
+        DEBUG_WITH_TYPE(LOGGER_BASE_LEVEL, logger.loadIndent(indent));         \
+    } while (false)
```
zacikpa, line 105: What exactly is the purpose of this macro? Does it have some future use? (currently it's not used anywhere)

zacikpa (commented):
@lenticularis39 Interestingly, for me, it works OK independently of the option (I'm building DiffKemp without the container).

@viktormalik I'm currently comparing rather small functions, so I hardly notice any delay. Fine by me to merge as is and fix the issue in a later PR.

xkucma00:
Thanks for the notes. The changes to the way comparison messages are stored are now implemented, I've noticed a significant speedup. I've also renamed the RETURN_WITH_LOG macros according to the suggestion and added an extended description of the usage and debug levels.

viktormalik (changes requested):
I found a couple of cases where the logging doesn't produce a nice output but I prefer to merge this PR first and will open separate issues afterwards.

The last thing that remains: there's a lot of commits, many of them repairing previous commits in the PR. Could you please do a cleanup and squash them into some reasonable amount?

xkucma00:
Okay, I've squashed the fixes and rebased it into 5 commits. According to `git diff`, it should be identical to the pre-rebase version.

viktormalik (approved):

PR #258, opened 2023-05-05
cc_wrapper: Always include -g option

lenticularis39:
Add -g to options always added to clang call, besides -S and -emit-llvm.

DiffKemp relies heavily on debug information for both finding source files to get diffs from and for handling structure types.


viktormalik:
@DanielKriz could you please check if this resolves your problem? Thanks!

DanielKriz:
@viktormalik Yes it does, thank you!

viktormalik (approved):

PR #264, opened 2023-06-27
Add --full-diff and --no-show-diff command-line options

zacikpa:
This PR adds two new command-line options to Diffkemp:
- `--no-show-diff`: The current default option is to **not** print syntactic differences. I propose that the default behaviour is that we **do** print them, and this option disables it.
- `--full-diff`:  This option causes Diffkemp to show syntactic differences even for functions that are semantically equal. This is useful when analyzing which functions with syntactic differences were successfully evaluated as semantically equal.

### DIFF
--- a/diffkemp/config.py
+++ b/diffkemp/config.py
@@ -9,7 +9,7 @@ class ConfigException(Exception):
 class Config:
     def __init__(self, snapshot_first, snapshot_second, show_diff,
                  output_llvm_ir, pattern_config, control_flow_only,
-                 print_asm_diffs, verbosity, use_ffi, semdiff_tool):
+                 print_asm_diffs, verbosity, use_ffi, full_diff, semdiff_tool):
         """
         Store configuration of DiffKemp
         :param snapshot_first: First snapshot representation.
@@ -31,6 +31,7 @@ class Config:
         self.print_asm_diffs = print_asm_diffs
         self.verbosity = verbosity
         self.use_ffi = use_ffi
+        self.full_diff = full_diff
 
         # Semantic diff tool configuration
         self.semdiff_tool = semdiff_tool
--- a/diffkemp/config.py
+++ b/diffkemp/config.py
@@ -1,10 +1,7 @@
 """Configuration of the tool."""
 from diffkemp.semdiff.pattern_config import PatternConfig
 from diffkemp.snapshot import Snapshot
-from diffkemp.utils import get_llvm_version
 import os
-import sys
-import errno
 
 
 class ConfigException(Exception):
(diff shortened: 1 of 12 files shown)


diffkemp/diffkemp.py
```diff
@@ -529,7 +497,8 @@ def text_indent(text, width):
     old_dir_abs_path = os.path.join(os.path.abspath(snapshot_dir_old), "")
     new_dir_abs_path = os.path.join(os.path.abspath(snapshot_dir_new), "")
 
-    if fun_result.kind == Result.Kind.NOT_EQUAL:
+    any_diff = any([x.diff for x in fun_result.inner.values()])
```
viktormalik, line 500: This looks like a potentially expensive operation and should be done lazily. Either inline the `any` call into the condition below or define `any_diff` as a lambda.

diffkemp/config.py
```diff
@@ -41,3 +61,60 @@ def __init__(self, snapshot_first, snapshot_second, show_diff,
...
+            sys.exit(errno.EINVAL)
+
+        if llvm_version != snapshot_second.llvm_version:
+            sys.stderr.write(
+                "Error: new snapshot was built with LLVM {}, "
+                "current version is LLVM {}.\n".format(
+                    snapshot_second.llvm_version, llvm_version
+                )
+            )
+            sys.exit(errno.EINVAL)
```
viktormalik, line 95: Could we move this code into `Snapshot.load_from_dir`? It would simplify this method and we could avoid the duplication here.

diffkemp/config.py
```diff
@@ -41,3 +61,60 @@ def __init__(self, snapshot_first, snapshot_second, show_diff,
...
+        return cls(
+            snapshot_first=snapshot_first,
+            snapshot_second=snapshot_second,
+            show_diff=not args.no_show_diff or args.show_diff,
+            full_diff=args.full_diff,
+            pattern_config=pattern_config,
+            control_flow_only=args.control_flow_only,
+            output_llvm_ir=args.output_llvm_ir,
+            print_asm_diffs=args.print_asm_diffs,
+            verbosity=args.verbosity,
```
viktormalik, line 117: This CLI option is now called `verbose`
```suggestion
            verbosity=args.verbose,
```

viktormalik (changes requested):

diffkemp/config.py
```diff
@@ -41,3 +61,60 @@ def __init__(self, snapshot_first, snapshot_second, show_diff,
...
+            use_ffi=args.use_ffi,
```
viktormalik, line 118: This is now a different CLI option
```suggestion
            use_ffi=not args.disable_simpll_ffi,
```

diffkemp/semdiff/function_diff.py
```diff
@@ -282,8 +284,12 @@ def functions_diff(mod_first, mod_second,
...
                 fun_result.first = fun_pair[0]
                 fun_result.second = fun_pair[1]
-                if fun_result.kind == Result.Kind.NOT_EQUAL:
-                    if fun_result.first.diff_kind in ["function", "type"]:
+                if fun_result.kind == Result.Kind.NOT_EQUAL or \
+                   config.full_diff:
+                    if fun_result.first.diff_kind in ["function", "type"] or (
+                            config.full_diff and
+                            fun_result.first.filename is not None and
+                            fun_result.second.filename is not None):
```
viktormalik, line 292: IIUC, this part of the condition adds some syntactic differences (i.e. macros diffs) which were not previously shown. Does that make sense here? Shouldn't we just keep the condition as it was since `config.full_diff` is already checked above?

zacikpa: You are right. For some reason, I thought that `diff_kind` is `None` when the compared functions are semantically equal, but that is not true.

viktormalik (approved):

viktormalik:
Thanks! I'll squash the commits into 1 if that's ok.

zacikpa:
Yes, that's OK.

viktormalik:
Done, I used the description from the PR.

Issue #263, opened 2023-06-20
`test_build_cscope_database` fails in the CI


viktormalik:
CI job: https://github.com/viktormalik/diffkemp/actions/runs/5320381768/jobs/9634075578

Failure log:
```
builder = <diffkemp.llvm_ir.kernel_llvm_source_builder.KernelLlvmSourceBuilder object at 0x7f0dcf682b80>

    @pytest.mark.parametrize("builder", versions, indirect=True)
    def test_build_cscope_database(builder):
        """Test building CScope database."""
        builder._build_cscope_database()
        for file in ["cscope.files", "cscope.in.out", "cscope.out",
                     "cscope.po.out"]:
>           assert os.path.isfile(os.path.join(builder.source_dir, file))
E           AssertionError: assert False
E            +  where False = <function isfile at 0x7f0dda7da700>('/home/runner/work/diffkemp/diffkemp/kernel/linux-3.10/cscope.in.out')
E            +    where <function isfile at 0x7f0dda7da700> = <module 'posixpath' from '/opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/posixpath.py'>.isfile
E            +      where <module 'posixpath' from '/opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/posixpath.py'> = os.path
E            +    and   '/home/runner/work/diffkemp/diffkemp/kernel/linux-3.10/cscope.in.out' = <function join at 0x7f0dda7daee0>('/home/runner/work/diffkemp/diffkemp/kernel/linux-3.10', 'cscope.in.out')
E            +      where <function join at 0x7f0dda7daee0> = <module 'posixpath' from '/opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/posixpath.py'>.join
E            +        where <module 'posixpath' from '/opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/posixpath.py'> = os.path
E            +      and   '/home/runner/work/diffkemp/diffkemp/kernel/linux-3.10' = <diffkemp.llvm_ir.kernel_llvm_source_builder.KernelLlvmSourceBuilder object at 0x7f0dcf682b80>.source_dir
```

All other tests pass to the database is built correctly. Maybe just the `cscope.in.out` file was renamed or so.

zacikpa:
[...]

PR #265, opened 2023-07-17
Update CI workflow to save space

zacikpa:
Apparently, issue #263 (which has been recently occurring non-deterministically in multiple branches) is caused by the lack of space in GitHub runners.

Github only [guarantees 14 GB of disk space ](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) in the runners. In practice, the available space is usually higher, but it does not suffice in some runs (just the `kernel` directory has 21 GB after all tests are run).

This PR attempts to solve the issue using the following workarounds and upgrades of the CI workflow:
-  At the beginning of `build-and-test` jobs, some unused software is deleted, namely .NET, Haskell, CodeQL, and Android SDK. This frees up to 20 GB of disk space. The solution is taken from [this action](https://github.com/marketplace/actions/maximize-build-disk-space).
- Kernel caching is moved to a separate job, which is a prerequisite of each `build-and-test` job. Using a new feature of the cache action, `build-and-test` only restores the cache, without attempting to submit it. This way, the jobs in the `build-and-test` matrix save the space necessary to create the compressed archive. Arguably, the solution is also prettier, because it is clear which job creates the cache (previously, it was the job which got to the post step first).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,38 @@ env:
     4.11
 
 jobs:
+  cache-kernels:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Cache Linux Kernels
+        id: kernel-cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ github.workspace }}/kernel
+          key: kernels-${{ env.KERNELS }}
+          lookup-only: true
+
+      - name: Install Dependencies
+        if: steps.kernel-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


lenticularis39 (approved):
Looks good to me, thanks for the investigation, I didn't even know that the CI builds include big "random" things like Android Studio!  IMO let's merge it and we'll see if the issue is solved eventually.

viktormalik (approved):

PR #266, opened 2023-07-21
Add the option to turn off built-in patterns

zacikpa:
This PR allows turning built-in patterns on/off in both Diffkemp and SimpLL.

Most commits are taken from [this old unmerged branch](https://github.com/viktormalik/diffkemp/tree/pattern-cli-options), rebased onto the current master branch, and refactored heavily. The configuration of relocation checking and inverse branch condition checking is added anew since these two patterns did not exist at the time.

I also decided to clearly distinguish between built-in patterns and custom patterns implemented by Petr (see commit [ad83f1](https://github.com/viktormalik/diffkemp/commit/ad83f1ddf8eeb5aa39be5205d4f61a64e3f22e3e)). Feel free to comment on that.

The PR is ready for review at this point, but I should probably add some tests in the meantime.


diffkemp/simpll/simpll.py
```diff
@@ -21,52 +21,100 @@ def add_suffix(file, suffix):
...
-    if verbosity == 0:
+    if config.verbosity == 0:
         stderr = open(os.devnull, "w")
 
     first_out_name = add_suffix(first, suffix) if suffix else first
     second_out_name = add_suffix(second, suffix) if suffix else second
 
-    if use_ffi:
+    if config.use_ffi:
+        builtin_pattern_args = ffi.new("struct builtin_pattern_args *")
```
viktormalik, line 40: The below is quite a chunk of code. Separate into function?

viktormalik (commented):
Looks good functionality-wise. Just a couple of comments.

diffkemp/simpll/ModuleAnalysis.cpp
```diff
@@ -174,7 +177,8 @@ void simplifyModulesDiff(Config &config, OverallResult &Result) {
                  mam.getResult<CalledFunctionsAnalysis>(*config.First,
                                                         config.FirstFun),
                  mam.getResult<CalledFunctionsAnalysis>(*config.Second,
-                                                        config.SecondFun));
+                                                        config.SecondFun),
+                 config);
```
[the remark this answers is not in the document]
zacikpa: In the end, I decided to pass just the built-in pattern configuration structure instead of the whole config - it's explicit which `Config` fields the `DebugInfo` constructor actually uses.

diffkemp/simpll/simpll.py
zacikpa: I resolved this using the method `as_ffi_struct` of the `BuiltinPatterns` class in the updated version of the PR.

diffkemp/simpll/simpll.py
```diff
@@ -21,52 +21,100 @@ def add_suffix(file, suffix):
...
+               cache_dir=None, module_cache=None, modules_to_cache=None):
     """
     Simplify modules to ease their semantic difference. Uses the SimpLL tool.
     :return A tuple containing the two LLVM IR files generated by SimpLL
             followed by the result of the comparison in the form of a graph and
             a list of missing function definitions.
     """
     stderr = None
-    if verbosity == 0:
+    if config.verbosity == 0:
```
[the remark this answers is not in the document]
viktormalik: Ah, ok. Let's leave it as it is now and if we do another rebase of this PR, you can do that.

diffkemp/simpll/simpll.py
```diff
@@ -132,34 +136,53 @@ def run_simpll(first, second, fun_first, fun_second, var, suffix=None,
...
-                simpll_command.append("--control-flow")
-
-            if output_llvm_ir:
+            if config.custom_pattern_config:
+                simpll_command.extend(["--custom-pattern-config",
+                                       config.custom_pattern_config.path])
+
+            # Builtin pattern configuration
+            if config.builtin_patterns.struct_alignment:
+                simpll_command.append("--struct-alignment")
```
viktormalik, line 151: Should we do:
```suggestion
            if config.builtin_patterns.struct_alignment:
                simpll_command.append("--struct-alignment")
            else:
                simpll_command.append("--struct-alignment=0
                ")
```
Otherwise, when a pattern is not enabled in `config.builtin_patterns`, the default SimpLL option will apply (which would be "on" in most cases and we don't want that).

diffkemp/simpll/Config.h
```diff
@@ -51,13 +64,15 @@ class Config {
...
     // Cache file directory.
     std::string CacheDir;
-    // Path to LLVM IR pattern configuration.
-    std::string PatternConfigPath;
+    // Path to custom LLVM IR differential pattern configuration.
+    std::string CustomPatternConfigPath;
+
+    // The following structure specifies which built-in patterns
+    // should be treated as semantically equal.
+    struct BuiltinPatterns BIPatterns;
```
viktormalik, line 72: It could be just me but I would prefer just `Patterns`. It's not obvious what `BIPatterns` stands for anyways, so you still have to look up the type (`BuiltinPatterns`) if you're new to this.

viktormalik (approved):
Looks more elegant now, thanks :-)

viktormalik:
Can you rebase the commits?

zacikpa:
> Can you rebase the commits?

Done.

PR #274, opened 2023-08-29
 Viewer: Add tests for Callstack appearance

PLukas2018:
This PR adds tests for testing if calls in `Callstack` are properly aligned.
The tests use:
- [`getBoundingClientRect`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) method which provides info about DOM elements' position/coordinates,
- [Cypress](https://www.cypress.io/): FE testing tool which runs in/on the browser, it allows testing of components' visualisation (appearance, style).


viktormalik (approved):
Beautiful piece of work, thanks!

Issue #277, opened 2023-09-04
Install viewer to system


viktormalik:
Currently, viewer requires the `view` directory to be located relatively to the `diffkemp.py` file (see [diffkemp.py](https://github.com/viktormalik/diffkemp/blob/2db77b5759bd5b3f889d078af2932a2db1affbd8/diffkemp/diffkemp.py#L582)). This is not possible when DiffKemp is installed to the system, e.g. from the RPM file or using `ninja install -C build && pip install .`

We should figure out a way to install the viewer to the system and update `diffkemp view` to use the installed version, if the local one doesn't exist (something similar to what we do for the [SimpLL binary](https://github.com/viktormalik/diffkemp/blob/master/diffkemp/simpll/simpll.py#L118-L123)).

IIUC, one of the problems is that `diffkemp view` needs to write source and diff files into `view/build/`, which will be problematic if `view/` is installed in a system path. Perhaps we could update the viewer such that we can pass it the `build/` directory upon executing `npm prod run`.

It would be also nice if we could build the viewer as a static site but that sounds even more complicated w.r.t. accessing local files.

cc @PLukas2018

PLukas2018:
I came up with 3 possible solutions:
1. **Build the viewer as a static site** (without using the server): There is the problem with accessing local files.
    - Possibilities to access local files:
        1. Using **AJAX** (fetch, XMLHttpRequest) is not an option ([CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSRequestNotHttp?utm_source=devtools&utm_medium=firefox-cors-errors&utm_campaign=default)).
        2. Using a **folder picker** ([File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API), [drag and drop](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/getAsFileSystemHandle), [`<input type="file">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file)) -- the disadvantage is that user has to pick a folder (which would contain src files and diffs) in the browser.
        3. Other options?
2. **More clever backend/server**:
    - Create a more clever server that would run the result viewer and serve him src and diff files from a specified local directory.
    - The server and the `build` folder would need to be probably installed (should be enough as read-only).  
    - The server could read the src files of the compared project from the project folder and provide it to the browser (the result viewer).
[...]

PR #278, opened 2023-09-07
Viewer: Install build to system

PLukas2018:
This PR makes it possible to run the result viewer when DiffKemp is installed to system (see #277).

Build of the viewer is installed directly to `/var/lib/diffkemp/view/build`, originally I wanted to use `CMAKE_INSTALL_LOCALSTATEDIR`  (= `var`) variable, but the problem would be in locating the directory when running `diffkemp view` because the exact location would be dependant on `CMAKE_INSTALL_PREFIX` variable.

When DiffKemp is installed, `diffkemp view` runs `serve` directly via `npx`, the impact is that when first running `diffkemp view` the user will be asked if it is ok to install `serve` package (if it is not installed).

Also, I was thinking (probably overthinking) when copying src files of a compared project to `build` directory for the result viewer, if the copied files should not have read permission only for the user -- so it cannot be read by other users (for security reasons). Right now the copied files have the same permissions as the original ones.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,11 +52,11 @@ if (${BUILD_VIEWER})
     FILE_PERMISSIONS OWNER_READ OWNER_WRITE
                      GROUP_READ
                      WORLD_READ
-    # Setting 777 for directories to make it possible to create
-    # `src` and `diffs` folder by user (`diffkemp view`).
+    # Setting 711 for directories to make it possible 
+    # to access files/folders in the directory. 
     DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
-                     GROUP_READ GROUP_WRITE GROUP_EXECUTE
-                     WORLD_READ WORLD_WRITE WORLD_EXECUTE
+                          GROUP_EXECUTE
+                          WORLD_EXECUTE
     # Do not install files describing results of comparison.
     PATTERN "diffkemp-out.yaml" EXCLUDE
     PATTERN "src" EXCLUDE
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,27 @@ if (${BUILD_VIEWER})
     PATTERN "diffkemp-out.yaml" EXCLUDE
     PATTERN "src" EXCLUDE
     PATTERN "diffs" EXCLUDE
-  ) # install
-endif ()
+  ) # install - copy `build` dir
+  # Creating `src`, `diffs` directories
+  # and setting the permissions to 777 to make able to copy there files and remove them.
+  install(DIRECTORY DESTINATION /var/lib/diffkemp/view/src
+    DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                          GROUP_READ GROUP_WRITE GROUP_EXECUTE
+                          WORLD_READ WORLD_WRITE WORLD_EXECUTE
+  ) # install src dir
+  install(DIRECTORY DESTINATION /var/lib/diffkemp/view/diffs
+    DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                          GROUP_READ GROUP_WRITE GROUP_EXECUTE
+                          WORLD_READ WORLD_WRITE WORLD_EXECUTE
+  ) # install diffs dir
+  # Creating diffkemp-out.yaml
+  # and setting its permission to 666 to make able edit its content by `diffkemp view`.
+  install(CODE "file(TOUCH /var/lib/diffkemp/view/diffkemp-out.yaml)")
+  install(CODE "file(CHMOD /var/lib/diffkemp/view/diffkemp-out.yaml
+                           PERMISSIONS OWNER_READ OWNER_WRITE
+                                       GROUP_READ GROUP_WRITE
+                                       WORLD_READ WORLD_WRITE
+  )") # install - file chmod
+endif () # ${BUILD_VIEWER}
 
 add_subdirectory(tests/unit_tests/simpll)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,22 @@ if (${BUILD_VIEWER})
     COMMAND ${NPM} run build
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/view
   )
+  install(DIRECTORY view/build DESTINATION /var/lib/diffkemp/view
+    # Setting 755 for files.
+    FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                     GROUP_READ GROUP_EXECUTE
+                     WORLD_READ WORLD_EXECUTE
+    # Setting 777 for directories to make it possible to create
+    # `src` and `diffs` folder by user (`diffkemp view`).
+    DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                     GROUP_READ GROUP_WRITE GROUP_EXECUTE
+                     WORLD_READ WORLD_WRITE WORLD_EXECUTE
+    # Making it possible to remove `diffkemp-out.yaml`
+    # by a user (`diffkemp view`) if it existed during installation.
+    PATTERN "diffkemp-out.yaml"
+         PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                     GROUP_READ GROUP_WRITE GROUP_EXECUTE
+                     WORLD_READ WORLD_WRITE WORLD_EXECUTE
+  ) # install
 endif ()
-
 add_subdirectory(tests/unit_tests/simpll)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,9 @@ if (${BUILD_VIEWER})
     COMMAND ${NPM} run build
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/view
   )
-  install(DIRECTORY view/build/ DESTINATION /var/lib/diffkemp/view
+
+  set(VIEW_INSTALL_DIR /var/lib/diffkemp/view)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -49,6 +49,22 @@ if (${BUILD_VIEWER})
     COMMAND npm run build
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/view
   )
+  install(DIRECTORY view/build DESTINATION /var/lib/diffkemp/view
+    # Setting 755 for files.
+    FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                     GROUP_READ GROUP_EXECUTE
+                     WORLD_READ WORLD_EXECUTE
+    # Setting 777 for directories to make it possible to create
+    # `src` and `diffs` folder by user (`diffkemp view`).
```
viktormalik, line 58: These is quite a permissive access, maybe we could create empty `src/` and `diffs/` subdirs and only grant those 777 (and 755 to the rest)?

Related: do not install contents of `src/` and `diffs/`.

CMakeLists.txt
```diff
@@ -49,6 +49,22 @@ if (${BUILD_VIEWER})
...
+    DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                     GROUP_READ GROUP_WRITE GROUP_EXECUTE
+                     WORLD_READ WORLD_WRITE WORLD_EXECUTE
+    # Making it possible to remove `diffkemp-out.yaml`
+    # by a user (`diffkemp view`) if it existed during installation.
```
viktormalik, line 63: The file should not exist during installation. If it does, let's not install it.

viktormalik (changes requested):
> Build of the viewer is installed directly to `/var/lib/diffkemp/view/build`

Can we install it to `/var/lib/diffkemp/view`? No need to have the `build` subdir when installed to the system path.

viktormalik:
> Build of the viewer is installed directly to `/var/lib/diffkemp/view/build`, originally I wanted to use `CMAKE_INSTALL_LOCALSTATEDIR` (= `var`) variable, but the problem would be in locating the directory when running `diffkemp view` because the exact location would be dependant on `CMAKE_INSTALL_PREFIX` variable.

You'd have to introduce a new option (or env variable) which could be used to pass the installed directory location to `diffkemp view`. But I don't think it's necessary at this point.

> When DiffKemp is installed, `diffkemp view` runs `serve` directly via `npx`, the impact is that when first running `diffkemp view` the user will be asked if it is ok to install `serve` package (if it is not installed).

I noticed, I don't think that it's a problem. We can let the RPM file install the package.

> Also, I was thinking (probably overthinking) when copying src files of a compared project to `build` directory for the result viewer, if the copied files should not have read permission only for the user -- so it cannot be read by other users (for security reasons). Right now the copied files have the same permissions as the original ones.

A good point, I thought of the same. I think that it's ok to leave the original permissions for now, if the user wants extra security, he can manage those permissions himself. IMHO, we'll need to revisit the permissions when we think about deploying the viewer to a server, so no need to worry about it now.

diffkemp/diffkemp.py
```diff
@@ -654,6 +670,9 @@ def view(args):
...
 
     if args.devel:
-        os.system(f"cd {VIEW_DIRECTORY} && npm start")
+        os.system(f"cd {view_directory} && npm start")
+    # run serve directly if diffkemp is installed
+    elif view_directory == VIEW_INSTALL_DIR:
+        os.system(f"npx serve {PUBLIC_DIRECTORY}")
     else:
-        os.system(f"cd {VIEW_DIRECTORY} && npm run prod")
+        os.system(f"cd {view_directory} && npm start")
```
viktormalik, line 678: Why not `npm run prod` as was here before?

viktormalik:
I'd like to merge #280 first and then rebase this to also use `http.server` instead of `npx serve` to run the production build.

diffkemp/diffkemp.py
viktormalik: Not relevant after #280 is merged.

PLukas2018:
I am thinking that it could be possible to make separately the `build` of the viewer (e.g. commit it to the repository, not sure if good idea because it will often change), and the installation would only move the `build` directory to `/var/lib/diffkemp/view/`.

viktormalik:
> I am thinking that it could be possible to make separately the `build` of the viewer (e.g. commit it to the repository, not sure if good idea because it will often change), and the installation would only move the `build` directory to `/var/lib/diffkemp/view/`.

I believe that we already do that. With this PR, you can build the viewer with `cmake -DBUILD_VIEWER=On` and then do the installation independently by `ninja install` (which basically copies the necessary files).

CMakeLists.txt
```diff
@@ -49,6 +49,22 @@ if (${BUILD_VIEWER})
     COMMAND npm run build
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/view
   )
+  install(DIRECTORY view/build DESTINATION /var/lib/diffkemp/view
+    # Setting 755 for files.
```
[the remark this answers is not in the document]
PLukas2018: Yes it does.

CMakeLists.txt
PLukas2018: 711 for the rest (not `diffs/` and `src/`) looks like is enough.

viktormalik: I'm not sure, you usually need R+E to do anything reasonable with a directory. I'd stick with the standard 755 for directories.

CMakeLists.txt
PLukas2018: There was a problem with creating `diffkemp-out.yaml` because of the set directory permissions, I decided to create this file (empty) when installing, and `diffkemp view` will write to it and after it exits it will make the file again empty.

CMakeLists.txt
PLukas2018: Ok, to be safe I changed it to 755, but when I tested 711 did work too.

CMakeLists.txt
viktormalik: Ah, I see, that's the one file created by `diffkemp view `. Ok, let's stick with an empty file for now but I'd like to handle that properly later.

diffkemp/diffkemp.py
PLukas2018: Thanks, you were right, it was my mistake...

PLukas2018:
> > I am thinking that it could be possible to make separately the `build` of the viewer (e.g. commit it to the repository, not sure if good idea because it will often change), and the installation would only move the `build` directory to `/var/lib/diffkemp/view/`.
> 
> I believe that we already do that. With this PR, you can build the viewer with `cmake -DBUILD_VIEWER=On` and then do the installation independently by `ninja install` (which basically copies the necessary files).

Thanks that makes more sense. I am not sure how you plan to 'distribute' it using rpm, but I changed the installation of the viewer not dependent on the `-DBUILD_VIEWER` option of `cmake` but rather check if the `build` directory exists. Now it could be possible to 'distribute' the srcs with `view/build/` dir already created and using `cmake` without the option `-DBUILD_VIEWER=On` to install the viewer.

PLukas2018:
I have tested the `diffkemp view` command (before [the dirs permission change to 755](https://github.com/viktormalik/diffkemp/pull/278/commits/0e89f2b10ef75d54e7582b13c04450aac9d307d0) commit)  with `-DBUILD_VIEWER` `On` and `Off` and without installation and with installation and it looked like it did what it was supposed to do.

CMakeLists.txt
```diff
@@ -47,6 +47,43 @@ if (${BUILD_VIEWER})
     COMMAND ${NPM} run build
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/view
   )
-endif ()
+endif () # ${BUILD_VIEWER}
+if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/view/build)
+  install(DIRECTORY view/build/ DESTINATION /var/lib/diffkemp/view
```
viktormalik, line 52: Create a variable for `/var/lib/diffkemp/view` so that we can conveniently change it later if we decide to.

viktormalik (commented):
I apologize for creating more work but I just got an idea: what if we placed `src/`, `diffs/`, and `diffkemp-out.yaml` to a separate directory within the viewer `build` directory (something like `data`)? I feel like that would simplify a lot of code:

- just create an empty `data` dir with 777 during installation, nothing else,
- no need to deal with deleting `src/` and `diffs/` contents without deleting the directories themselves - we'd just delete both dirs and `diffkemp-out.yaml` at the end of `diffkemp view`

OFC we'd need some changes to the code but hopefully it's just the matter of setting `SOURCE_DIRECTORY` and `DIFF_DIRECTORY` in `view()`.

viktormalik:
> Thanks that makes more sense. I am not sure how you plan to 'distribute' it using rpm, but I changed the installation of the viewer not dependent on the `-DBUILD_VIEWER` option of `cmake` but rather check if the `build` directory exists. Now it could be possible to 'distribute' the srcs with `view/build/` dir already created and using `cmake` without the option `-DBUILD_VIEWER=On` to install the viewer.

I don't think that's a good idea. Creation of the `build` directory (or, better, a static build) should be a prerequisite of installing the viewer so they should be under the same CMake option so that we can be sure that the `view/build` directory was created in a correct way.

PLukas2018:
> I apologize for creating more work but I just got an idea: what if we placed `src/`, `diffs/`, and `diffkemp-out.yaml` to a separate directory within the viewer `build` directory (something like `data`)?

Sure, I will do it, I think it makes more sense.

PLukas2018:
I addressed the comments and retested the `diffkemp view` command, it should work as expected.

viktormalik (approved):
Beautiful! Hope you don't mind if I squash the commits.

PLukas2018:
Thanks. Yes, squash the commits, please.
The description of commit could be something like:
```
Viewer: Install build to system

Making possible to run the result viewer when DiffKemp is installed to system.
Also `diffs`, `src` and `diffkemp-out.yaml` are now placed in `data` directory.
```

viktormalik:
Thanks! I updated it a bit and squashed.

PR #279, opened 2023-09-11
Add creating snapshots for a single C file

PLukas2018:
This PR extends `diffkemp build` to be able to make snapshots from a single C file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,9 @@ DiffKemp runs in two phases:
     also has additional options to configure the project build, see `diffkemp
     build --help` for the complete list.
 
+    The command can be also used to generate a snapshot from a single C file.
+    In this case, the path to the file should be given in place of `PROJ_DIR`.
+
   - ```
     diffkemp build-kernel KERNEL_DIR SNAPSHOT_DIR SYMBOL_LIST
     ```
--- a/diffkemp/llvm_ir/single_c_builder.py
+++ b/diffkemp/llvm_ir/single_c_builder.py
@@ -1,0 +1,42 @@
+"""
+LLVM source builder for single C file.
+"""
+from diffkemp.llvm_ir.single_llvm_finder import SingleLlvmFinder
+from diffkemp.utils import get_functions_from_llvm
+import os
+from subprocess import check_call
+
+
+class SingleCBuilder(SingleLlvmFinder):
+    """
+    LLVM source builder for single C file.
+    Extends the SingleLlvmFinder class by compiling C file to LLVM IR file.
+    """
+    def __init__(self, source_dir, c_file_name, clang="clang",
+                 clang_append=[]):
+        """
+        :param clang: clang compiler to be used
+        :param clang_append: list of args to add when compiling
+        """
+        llvm_file_name = os.path.splitext(c_file_name)[0] + ".ll"
+        SingleLlvmFinder.__init__(self, source_dir, llvm_file_name)
+
+        self.c_file_name = c_file_name
+        self.clang = clang
+        self.clang_append = clang_append
+        self.initialize()
+
+    def str(self):
+        return "single_c_file"
+
+    def initialize(self):
+        # Using clang to compile c file to llvm file.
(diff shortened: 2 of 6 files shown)


PLukas2018:
I used the same clang flags that are used by `diffkemp build` now
https://github.com/viktormalik/diffkemp/blob/2db77b5759bd5b3f889d078af2932a2db1affbd8/diffkemp/building/cc_wrapper.py#L176 

but should not it be somehow unified with flags that are used for `diffkemp build-kernel`
https://github.com/viktormalik/diffkemp/blob/2db77b5759bd5b3f889d078af2932a2db1affbd8/diffkemp/llvm_ir/kernel_llvm_source_builder.py#L347-L348
?

diffkemp/llvm_ir/wrapper_build_finder.py
```diff
@@ -38,16 +38,7 @@ def initialize(self):
         # Read LLVM IR files, get all LLVM functions that satisfy C naming from
         # there and save them into self.functions along with the files there
         # were found in
```
viktormalik, line 40: This comment is not relevant anymore, let's remove it from here.

README.md
```diff
@@ -92,6 +92,10 @@ DiffKemp runs in two phases:
     also has additional options to configure the project build, see `diffkemp
     build --help` for the complete list.
 
+    The command can be also used for a snapshot generation for a single C file,
+    in this case, the `PROJ_DIR` represents the path to the file. 
+    If `SYMBOL_LIST` is not given, all functions from the file are considered.
```
viktormalik, line 97: Small rewording:
```suggestion
    The command can be also used to generate a snapshot from a single C file.
    In this case, the path to the file should be given in place of `PROJ_DIR`. 
    
    If `SYMBOL_LIST` is not given, all functions from the project/file are considered.
```

diffkemp/llvm_ir/single_c_builder.py
```diff
@@ -0,0 +1,42 @@
...
+    LLVM source builder for single C file.
+    Extends the SingleLlvmFinder class by compiling C file to LLVM IR file.
+    """
+    def __init__(self, source_dir, c_file_name, clang="clang",
+                 clang_append=[]):
+        """
+        :param clang: clang compiler to be used
+        :param clang_append: list of args to add when compiling
+        """
+        llvm_file_name = c_file_name.rsplit(".", 1)[0] + ".ll"
```
viktormalik, line 21: Use `os.path.splitext[0] + .ll`.

viktormalik, line 21: Also shouldn't be?
```suggestion
        self.llvm_file_name = c_file_name.rsplit(".", 1)[0] + ".ll"
```

viktormalik (commented):
Can we add a unit test for this?

viktormalik:
> I used the same clang flags that are used by `diffkemp build` now
>
> but should not it be somehow unified with flags that are used for `diffkemp build-kernel`
> ?

Good point. The `build-kernel` args were chosen like this b/c we wanted to have control over what LLVM passes are being applied. I agree that we should do the same for `build` but that's a different discussion from what this PR is about so let's keep it as it is for now.

diffkemp/llvm_ir/single_c_builder.py
viktormalik: I missed that, correct.

README.md
PLukas2018: I omitted the last line because `SYMBOL_LIST` is already mentioned in the lines above
https://github.com/viktormalik/diffkemp/blob/abab4e4bdf8ec485d944a2d8b552a2a736dff257/README.md?plain=1#L85-L86

PLukas2018:
I addressed the comments, fixed a mistake I made (returning a dict instead of a list) and added tests. I guess I could add more tests (testing `SYMBOL_LIST` or `build` command in general...).

viktormalik:
Thanks! Can you squash the commits? Looks like leaving only the first two is the best.

viktormalik (approved):
Works nicely, thanks!

PR #283, opened 2023-09-20
Update name of rhel-kernel-get script

PLukas2018:
I discovered a few places where was used the old name (with `.py` extension) of `rhel-kernel-get` script, the most problematic is in `ci.yml`. This PR updates the name.


viktormalik (approved):
Good catch, thanks!

PR #284, opened 2023-09-21
SimpLL unit test improvements

viktormalik:
While working on another feature (better development environment), a couple of small improvements for SimpLL unit tests popped up and I realized that they can be useful in general, so posting them straight away.

The list of improvements (see commits for details):
- make vendoring of GoogleTest optional via the `-DVENDOR_GTEST` option of CMake,
- move `runTests` binary back to its default location (i.e. inside the build directory),
- add a new target `ninja test` for conveniently running the unit tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,7 @@ jobs:
       - name: Run Unit Tests
         working-directory: ${{ github.workspace }}
         run: |
-          tests/unit_tests/simpll/runTests
+          build/tests/unit_tests/simpll/runTests
 
       - name: Run Regression Tests
         if: matrix.regression-tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,7 @@ jobs:
         env: ${{ matrix.env }}
         working-directory: ${{ github.workspace }}/build
         shell: bash
-        run: cmake $GITHUB_WORKSPACE -GNinja -DSANITIZE_ADDRESS=${{ matrix.asan }}
+        run: cmake $GITHUB_WORKSPACE -GNinja -DSANITIZE_ADDRESS=${{ matrix.asan }} -DVENDOR_GTEST=On
 
       - name: Build
         env: ${{ matrix.env }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,10 +132,9 @@ jobs:
         working-directory: ${{ github.workspace }}
         run: pip3 install -e .
 
-      - name: Run Unit Tests
+      - name: Run SimpLL Unit Tests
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -135,7 +135,7 @@ jobs:
       - name: Run Unit Tests
         working-directory: ${{ github.workspace }}
         run: |
-          tests/unit_tests/simpll/runTests
+          build/tests/unit_tests/simpll/runTests
```
PLukas2018, line 138: I guess it could be "simplified" to just `ninja test` if working directory would be changed to `${{ github.workspace }}/build`.

PLukas2018 (commented):
Just to learn more about DiffKemp I tried the changes you made. `ninja test` with `-DVENDOR_GTEST` works well, I tried also without `-DVENDOR_GTEST` on Fedora 38 and after installing `gtest-devel` it works ok too.

.github/workflows/ci.yml
viktormalik: Good idea, will do that.

viktormalik:
Fixed.

@PLukas2018 feel free to approve the PR, I added you to collaborators so your approve will enable merging.

PLukas2018 (approved):
Just thought which occurred me, if it was not possible to check with something like `find_library` if GoogleTests are already installed within the system, if they are not then vendor it with DiffKemp. But I think your approach does make more sense because you do not always need to test DiffKemp, so no need to always have GoogleTests. Otherwise looks good to me.

PLukas2018:
> @PLukas2018 feel free to approve the PR, I added you to collaborators so your approve will enable merging.

Thanks, I noticed, I just was worried that I am not experienced enough to approve PRs.

viktormalik:
> Just thought which occurred me, if it was not possible to check with something like `find_library` if GoogleTests are already installed within the system, if they are not then vendor it with DiffKemp.

Sometimes, you may want to vendor GTest even if it is installed on the system (e.g. when the installed version is too old), so having a config option offers a better control.

> But I think your approach does make more sense because you do not always need to test DiffKemp, so no need to always have GoogleTests.

You need GTest in any way b/c otherwise the compilation would fail. Perhaps it'd make sense to add an option not to build testing but we can always add that later.

> Otherwise looks good to me.

Thanks!

lenticularis39:
> Thanks, I noticed, I just was worried that I am not experienced enough to approve PRs.

You are the fourth most significant contributor to DiffKemp per commit, and the third per added lines, I'd say you don't have to worry about that :)

viktormalik:
> Thanks, I noticed, I just was worried that I am not experienced enough to approve PRs.

No worries, just approve those that you feel comfortable about (such as this one) :-)

Issue #281, opened 2023-09-13
Logger: compilation error for CMAKE_BUILD_TYPE=Release


PLukas2018:
When compiling DiffKemp using `cmake` with `-DCMAKE_BUILD_TYPE=Release` (as specified in `README.md`) and then runing `ninja`, the compilation ends with error:

```
[10/57] Building CXX object diffkemp/s...CMakeFiles/simpll-lib.dir/Logger.cpp.o
FAILED: diffkemp/simpll/CMakeFiles/simpll-lib.dir/Logger.cpp.o 
/usr/bin/c++  -I/home/lukas/BC/diffkemp/build/diffkemp/simpll -I/home/lukas/BC/diffkemp/diffkemp/simpll -isystem /usr/lib/llvm-14/include -fno-rtti -fpic -Wall -Wextra -O3 -DNDEBUG   -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fdiagnostics-color=always -std=gnu++17 -MD -MT diffkemp/simpll/CMakeFiles/simpll-lib.dir/Logger.cpp.o -MF diffkemp/simpll/CMakeFiles/simpll-lib.dir/Logger.cpp.o.d -o diffkemp/simpll/CMakeFiles/simpll-lib.dir/Logger.cpp.o -c /home/lukas/BC/diffkemp/diffkemp/simpll/Logger.cpp
In file included from /home/lukas/BC/diffkemp/diffkemp/simpll/Logger.h:101,
                 from /home/lukas/BC/diffkemp/diffkemp/simpll/Logger.cpp:14:
/home/lukas/BC/diffkemp/diffkemp/simpll/Logger.cpp: In member function ‘void Logger::log(bool, const char*)’:
/home/lukas/BC/diffkemp/diffkemp/simpll/Logger.cpp:32:34: error: expected unqualified-id before ‘(’ token
   32 |                       && ::llvm::isCurrentDebugType(force_keep_type);
      |                                  ^~~~~~~~~~~~~~~~~~
``` 

I am not sure if it is not problem on my end, my `c++` is `c++ (Ubuntu 9.5.0-1ubuntu1~22.04) 9.5.0`.
Without the option `-DCMAKE_BUILD_TYPE=Release`  the compilation is without error.

PR #286, opened 2023-09-26
Fix build & build instructions

FrNecas:
More details in the commits. As for the README change, there doesn't seem to be a way to build without tests right now unless I am missing something.

Fixes: #281

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ Currently, DiffKemp runs on Linux and needs the following software installed:
 * Python 3 with CFFI (package `python3-cffi` in Fedora and Debian)
 * Python packages from `requirements.txt` (run `pip install -r requirements.txt`)
 * CScope (when comparing versions of the Linux kernel)
+* GoogleTest (gtest) for running the C++ tests (can be vendored by using
+  `-DVENDOR_GTEST=On` in the `cmake` command)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -39,6 +39,7 @@ Currently, DiffKemp runs on Linux and needs the following software installed:
 * Python 3 with CFFI (package `python3-cffi` in Fedora and Debian)
 * Python packages from `requirements.txt` (run `pip install -r requirements.txt`)
 * CScope (when comparing versions of the Linux kernel)
+* GoogleTest (gtest) for running the C++ tests
```
viktormalik, line 42: Maybe add information that GTest can be vendored?
```suggestion
* GoogleTest (gtest) for running the C++ tests (can be vendored by using `-DVENDOR_GTEST=On` in CMake)
```

viktormalik (commented):
Please fix the clang-format issue, too.

> As for the README change, there doesn't seem to be a way to build without tests right now unless I am missing something.

You're right, there's not. It'd make sense to allow something like `-DBUILD_TESTS=Off` in CMake.

README.md
FrNecas: Good idea, fixed

viktormalik (approved):

PR #288, opened 2023-10-06
Docker-dev: using shutil.which instead of command

PLukas2018:
When running `diffkemp-devel/run-container.py` on Ubuntu it does not recognise command `command` (probably because it is a shell built-in command). This PR fixes it using `shutil.which` instead.


viktormalik (approved):
Looks good, thanks :-)

PR #290, opened 2023-10-07
Match PHI nodes on first use, not based on their order in a basic block

zacikpa:
After some refactorings, PHI nodes may get reordered. Currently, we do not handle this since PHI nodes are matched in the order in which they appear at the beginning of a basic block.

This simple PR solves the problem by using the following approach:
- During `cmpBasicBlocks`, all comparisons of PHI nodes are skipped.
- When a pair of PHI nodes is used for the first time (by some non-PHI instructions), it is matched.
- After all non-PHI instructions are compared, the matched pairs of PHI nodes are compared.

The PR also fixes a bug in `cmpPHIs` which could cause `cmpPHIs` to compare invalid pairs of PHI-incoming values/blocks. The bug caused the `scnprintf` function to be compared as non-equal between RHEL 8.0 and RHEL 8.1.


viktormalik (approved):
LGTM, thanks!

PR #291, opened 2023-10-11
CI speedup: move CScope db build to the cache-kernels job

viktormalik:
When running regression tests on freshly downloaded kernels, it is usually very slow b/c it needs to build the CScope database for each used kernel version.
    
In CI, this is done for every job in the matrix, despite the fact that building CScope database does not depend on LLVM version.
    
To speed things up, this PR moves building of the CScope db to the 'cache-kernels' job which downloads and caches the kernels. This is done by running `make cscope` for each downloaded kernel.
    
Adding some comparison of CI runtimes below.
    
```
Current master:
- with cached kernels:    30m 56s
- without cached kernels: 43m 43s (11m 18s in cache-kernels)
    
This PR:
- with cached kernels:    19m 2s
- without cached kernels: 40m 8s (21m 4s in cache-kernels)
```
    
When considering some measurement errors, we can see that it takes roughly the same time when CI needs to refetch the kernels but there's a significant speedup when using pre-cached kernels.

This also required a small change in one unit test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,9 +27,19 @@ jobs:
           key: kernels-${{ env.KERNELS }}
           lookup-only: true
 
+      - name: Delete unused software
+        if: steps.kernel-cache.outputs.cache-hit != 'true'
+        # This is necessary for running CScope database build as it takes
+        # a lot of disk space and the runners die silently when out of space.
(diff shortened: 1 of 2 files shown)


zacikpa (approved):
Looks good, thanks!

PR #293, opened 2023-10-18
Extend statistics

FrNecas:
This PR adds the following statistics:
 - number of instructions compared
 - number of instructions where 1:1 comparison was used (i.e. semantic preserving pattern was not used)
 - number of C code lines analyzed

Tracking the lines analyzed requires a set and the calculations on top of this data-structure resulted in a 10 second (260 -> 270) increase when analyzing the RHEL KABI. This feels quite significant, therefore it has to be explicitly enabled using `--extended-stat` option. The other statistics are just simple counters, therefore, they are included in the base statistics.

Example output comparing RHEL8 kernels:
```
Lines compared:          36537
Instructions compared:   109760
1:1 equal instructions:  107380 (98%)
```

### DIFF
--- a/diffkemp/cli.py
+++ b/diffkemp/cli.py
@@ -105,6 +105,10 @@ def make_argument_parser():
     compare_ap.add_argument("--report-stat",
                             help="report statistics of the analysis",
                             action="store_true")
+    compare_ap.add_argument("--extended-stat",
+                            help="report detailed statistics about the "
+                                 "analysis (may be more expensive to compute)",
+                            action="store_true")
     compare_ap.add_argument("--source-dirs",
                             nargs=2,
                             help="specify root dirs for the compared projects")
--- a/diffkemp/semdiff/caching.py
+++ b/diffkemp/semdiff/caching.py
@@ -44,6 +44,19 @@ class ComparisonGraph:
             else:
                 return ComparisonGraph.DependencyKind.STRONG
 
+    class FunctionStats:
+        """Contains statistics about analysis of a single function."""
+        def __init__(self, left, right):
+            self.inst_cnt = (left.get("inst-cnt", 0), right.get("inst-cnt", 0))
+            self.inst_equal_cnt = (left.get("inst-equal-cnt", 0),
+                                   right.get("inst-equal-cnt", 0))
+
+        def compared_inst_cnt(self):
+            return self.inst_cnt[0] + self.inst_cnt[1]
+
+        def compared_inst_equal_cnt(self):
+            return self.inst_equal_cnt[0] + self.inst_equal_cnt[1]
+
     class Vertex:
         """
         Vertex in the comparison graph.
@@ -55,13 +68,14 @@ class ComparisonGraph:
         Note: names, files and lines are tuples containing the values for both
         modules.
         """
-        def __init__(self, names, result, files=None, lines=None):
+        def __init__(self, names, result, files=None, lines=None, stats=None):
             self.names = names
             self.result = result
             self.nonfun_diffs = []
             self.successors = ([], [])
             self.files = files
             self.lines = lines
+            self.stats = stats
             # Vertices are by default cachable, but there are some cases when
             # it is necessary to run the comparison again.
             self.cachable = True
@@ -90,10 +104,10 @@ class ComparisonGraph:
             vertex = cls(
                 (res_left["function"], res_right["function"]),
                 Result.Kind.from_string(fun_result["result"]),
-                (res_left["file"] if "file" in res_left else None,
-                 res_right["file"] if "file" in res_right else None),
-                (res_left["line"] if "line" in res_left else None,
-                 res_right["line"] if "line" in res_right else None)
+                (res_left.get("file", None), res_right.get("file", None)),
+                (res_left.get("line", None), res_right.get("line", None)),
+                ComparisonGraph.FunctionStats(
+                    res_left.get("stats", {}), res_right.get("stats", {}))
             )
             if "calls" in res_left:
                 for res, side in [(res_left,
--- a/diffkemp/semdiff/caching.py
+++ b/diffkemp/semdiff/caching.py
@@ -50,12 +50,17 @@ class ComparisonGraph:
             self.inst_cnt = (left.get("inst-cnt", 0), right.get("inst-cnt", 0))
             self.inst_equal_cnt = (left.get("inst-equal-cnt", 0),
                                    right.get("inst-equal-cnt", 0))
+            self.lines_cnt = (left.get("lines-cnt", 0),
+                              right.get("lines-cnt", 0))
 
         def compared_inst_cnt(self):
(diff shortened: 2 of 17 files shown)


diffkemp/semdiff/caching.py
```diff
@@ -55,13 +55,17 @@ class Vertex:
...
+                     inst_cnt=None, inst_equal_cnt=None, lines_cnt=None):
             self.names = names
             self.result = result
             self.nonfun_diffs = []
             self.successors = ([], [])
             self.files = files
             self.lines = lines
+            self.inst_cnt = inst_cnt
+            self.inst_equal_cnt = inst_equal_cnt
+            self.lines_cnt = lines_cnt
```
viktormalik, line 68: I'm wondering if it's worth enclosing these into a separate object like `stats`. It could save some lines, especially if we add more stats in future, and also better represents the purpose of these fields.

The encapsulation would probably have to be done in SimpLL, too, so it's a bit of extra work.

FrNecas: Good idea, I will change this. I am not really sure about saving some lines. The `Stats` object/class in the frontend could load the sub-yaml dynamically but I think it will still be a bit more maintainable to explicitly state the `Stats` attributes in the frontend.

FrNecas: Done

viktormalik (commented):
LGTM, thanks!

One thing I'm thinking about is to move instructions counts behind `--extended-stats`, too. Not that it'd hard to compute, it's just that the numbers are intended for our experiments and don't seem much useful for regular users. In fact, the same applies for detailed statistics about compared objects (see `Result.report_object_stat`). Maybe we could move those under `--extended-stats`, too, as a part of this PR?

Also, could you please squash the first and the third commit together and add some better commit messages?

Thanks!

diffkemp/cli.py
```diff
@@ -105,6 +105,10 @@ def make_argument_parser():
     compare_ap.add_argument("--report-stat",
                             help="report statistics of the analysis",
                             action="store_true")
+    compare_ap.add_argument("--extended-stat",
+                            help="compute extended statistics that are more "
+                                 "expensive to compute (e.g. line count). ",
```
viktormalik, line 110: Maybe rephrase this?
```suggestion
                            help="report detailed statistics about the analysis "
                                 "(may be more expensive to compute). ",
```

diffkemp/simpll/ModuleComparator.cpp
```diff
@@ -104,6 +104,18 @@ void ModuleComparator::compareFunctions(Function *FirstFun,
...
+    ComparedFuns.at({FirstFun, SecondFun}).First.stats.instEqualCnt =
+            fComp.InstEqual;
+    ComparedFuns.at({FirstFun, SecondFun}).First.stats.linesCnt =
+            fComp.ComparedLinesL.size();
+    ComparedFuns.at({FirstFun, SecondFun}).Second.stats.instCnt =
+            fComp.ComparedInstR;
+    ComparedFuns.at({FirstFun, SecondFun}).Second.stats.instEqualCnt =
+            fComp.InstEqual;
+    ComparedFuns.at({FirstFun, SecondFun}).Second.stats.linesCnt =
+            fComp.ComparedLinesR.size();
```
viktormalik, line 118: Could we enclose this into a method/function? Also repeats below.

viktormalik (commented):
I'm getting division by zero:
```
$ bin/diffkemp compare snapshots/linux-4.18.0-80.el8 snapshots/linux-4.18.0-147.el8 -f __alloc_pages_nodemask --stdout --report-stat --extended-stat
[...]
Traceback (most recent call last):
  File "/home/vmalik/src/diffkemp/diffkemp/bin/diffkemp", line 6, in <module>
    diffkemp.run_from_cli()
  File "/home/vmalik/.local/lib/python3.11/site-packages/diffkemp/diffkemp.py", line 37, in run_from_cli
    args.func(args)
  File "/home/vmalik/.local/lib/python3.11/site-packages/diffkemp/diffkemp.py", line 503, in compare
    result.report_stat(args.show_errors, args.extended_stat)
  File "/home/vmalik/.local/lib/python3.11/site-packages/diffkemp/semdiff/result.py", line 292, in report_stat
    self.report_object_stat()
  File "/home/vmalik/.local/lib/python3.11/site-packages/diffkemp/semdiff/result.py", line 250, in report_object_stat
    equal_percent = equal_instructions / compared_instructions * 100
                    ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
ZeroDivisionError: division by zero
```

FrNecas:
I am not really sure what to output if there were no instructions compared. I've used `0 (nan%)` for now. Let me know if you think of something better. Or maybe we could leave out this statistic completely if the number of instructions is 0? I slightly lean towards being explicit and consistent between all runs (i.e. always output all statistics).

diffkemp/semdiff/result.py
```diff
@@ -259,6 +269,10 @@ def __hash__(self):
             print("Elapsed time:            {:.2f} s".format(
                 self.stop_time - self.start_time))
         print("Functions compared:      {}".format(compared))
+        print("Lines compared:          {}".format(compared_lines))
+        print("Instructions compared:   {}".format(compared_instructions))
+        print("1:1 equal instructions:  {0} ({1:.0f}%)".format(
+            equal_instructions, equal_percent))
```
viktormalik, line 275: Let's insert one blank line here (before `Total differences:`) to visually separate the following block.

viktormalik (commented):
> I am not really sure what to output if there were no instructions compared. I've used 0 (nan%) for now. Let me know if you think of something better. Or maybe we could leave out this statistic completely if the number of instructions is 0? I slightly lean towards being explicit and consistent between all runs (i.e. always output all statistics).

Agreed, let's be explicit. My example works now, I'm just wondering why the bug occurred when there's:
```
Lines compared:          0
Instructions compared:   10570
1:1 equal instructions:  10390 (98%)
```
Was there a bug before? Also I'm curious why the number of lines compared is 0.

Anyways, the update LGTM, just one small nit for better readability of the extended stats.

FrNecas:
> Was there a bug before? Also I'm curious why the number of lines compared is 0.

Are you sure that you are running the same example? I am still getting 0 compared instructions and 0 analyzed lines. Weird...

viktormalik:
> > Was there a bug before? Also I'm curious why the number of lines compared is 0.
> 
> Are you sure that you are running the same example? I am still getting 0 compared instructions and 0 analyzed lines. Weird...

Hm, I'm getting 0 everywhere now, too, but I'm sure that I ran the same example (just checked my terminal history). Strange... Anyways, there are 466 functions compared and no instructions so something is clearly off here.

viktormalik (changes requested):

viktormalik (approved):
Wonderful!

PR #294, opened 2023-10-19
Introduce Nix-based builds and environments

viktormalik:
This introduces new build and development environments based on [Nix flakes](https://nixos.wiki/wiki/Flakes). In short, Nix is a package manager with focus on replicability and consistency.

This provides several Nix flakes:
- Flake for building DiffKemp with `nix build` or `nix build .#diffkemp-llvm14` for building with a specific LLVM version. This creates a new directory `result/` containing DiffKemp with all dependencies managed by Nix. You can then run it with `result/bin/diffkemp ...`.
- Flake for development environment which can be entered by `nix develop` or `nix develop .#diffkemp-llvm14` (or any other supported LLVM version). This enters a new shell with all dependencies required for building and running DiffKemp, as well as dependencies for building and analysing reasonably modern Linux kernels.
- Flake for development environment suitable for building CentOS 7 kernels which are needed for regression tests. Can be entered by `nix develop .#test-kernel-buildenv`.

The main aim is to replace the Docker container as the default development environment. The workflow with Nix is:

1. Install Nix (see instructions in updated README).
2. Download kernels for tests:
```
[localhost]$ nix develop .#test-kernel-buildenv
[test-kernel-buildenv]$ rhel-kernel-get 3.10.0-514.el7 --output-dir kernel --kabi
[...] # repeat for all necessary kernels
[test-kernel-buildenv]$ exit
```
3. Build DiffKemp and run tests, either in the development flake or even on localhost (I believe that the latter requires the same version of glibc as is in Nix).
```
[localhost]$ nix develop        # sometimes not even necessary
[diffkemp-llvm16/localhost]$    # build DiffKemp
[diffkemp-llvm16/localhost]$ pytest tests
```
I'd love if someone could try and confirm that this works.

See commit messages for more implementation details.

I've also update CI to use Nix. The advantage is that we no longer depend on Ubuntu 20.04 which was the last version featuring GCC 7 necessary for building kernels for tests. Another big advantage is that the CI should be now deterministic and fully replicable, and therefore much easier to debug. The disadvantage is that tests runtime with Nix is remarkably slower, CI running 30+ minutes with caching (roughly the same time before #291. I'll try to debug that but I feel like the advantages slightly overrule the slowdown in this case.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,57 +7,58 @@
     let
       system = "x86_64-linux";
       pkgs = import nixpkgs { inherit system; };
-    in
-    {
-      formatter.${system} = pkgs.nixpkgs-fmt;
 
-      packages.${system}.default = with pkgs;
-        python3Packages.buildPythonPackage {
-          pname = "diffkemp";
-          version = "0.5.0";
+      mkDiffkemp =
+        llvmPackages:
+          with pkgs;
+          python3Packages.buildPythonPackage {
+            pname = "diffkemp";
+            version = "0.5.0";
 
-          src = self;
+            src = self;
 
-          nativeBuildInputs = with llvmPackages_16; [ cmake gcc libllvm ninja ];
+            nativeBuildInputs = with llvmPackages; [ cmake gcc libllvm ninja ];
 
-          buildInputs = with llvmPackages_16; [
-            clangNoLibcxx
-            cscope
-            diffutils
-            gtest
-            gnumake
-          ];
+            buildInputs = with llvmPackages; [
+              clangNoLibcxx
+              cscope
+              diffutils
+              gtest
+              gnumake
+            ];
 
-          propagatedBuildInputs = with python3Packages; [
-            cffi
-            pyyaml
-            setuptools
-          ];
+            propagatedBuildInputs = with python3Packages; [
+              cffi
+              pyyaml
+              setuptools
+            ];
 
-          # Including cmake in nativeBuildInputs automatically runs it during
-          # configurePhase so we just need to set correct flags.
-          cmakeFlags = [ "-DCMAKE_BUILD_TYPE=Debug" "-GNinja" ];
+            # Including cmake in nativeBuildInputs automatically runs it during
+            # configurePhase so we just need to set correct flags.
+            cmakeFlags = [ "-DCMAKE_BUILD_TYPE=Debug" "-GNinja" ];
 
-          # We're mixing Ninja and Python build here so we need to manually define
-          # buildPhase and installPhase to make sure both are built. CMake has
-          # switched dir to build/ so let's switch back and define ninjaFlags.
-          ninjaFlags = [ "-C" "build" ];
+            # We're mixing Ninja and Python build here so we need to manually
+            # define buildPhase and installPhase to make sure both are built.
+            # CMake has switched dir to build/ so let's switch back and define
+            # ninjaFlags.
+            ninjaFlags = [ "-C" "build" ];
 
-          buildPhase = ''
-            cd ..
-            ninjaBuildPhase
-            setuptoolsBuildPhase
-          '';
+            buildPhase = ''
+              cd ..
+              ninjaBuildPhase
+              setuptoolsBuildPhase
+            '';
 
-          installPhase = ''
-            ninjaInstallPhase
-            pipInstallPhase
-            install -m 0755 bin/diffkemp $out/bin/diffkemp
-          '';
-        };
+            installPhase = ''
+              ninjaInstallPhase
+              pipInstallPhase
+              install -m 0755 bin/diffkemp $out/bin/diffkemp
+            '';
+          };
 
-      devShells.${system} = {
-        default = with pkgs;
+      mkDiffkempDevShell =
+        diffkemp-pkg:
+          with pkgs;
           let
             rhel_kernel_get = python3Packages.buildPythonApplication {
               pname = "rhel-kernel-get";
@@ -72,7 +73,7 @@
             };
           in
           mkShell {
-            inputsFrom = [ self.packages.${system}.default ];
+            inputsFrom = [ diffkemp-pkg ];
 
             buildInputs = [
               bc
@@ -94,43 +95,70 @@
               pytest-mock
             ];
           };
+    in
+    {
+      formatter.${system} = pkgs.nixpkgs-fmt;
+
+      packages.${system} = rec {
+        default = diffkemp-llvm16;
+
+        diffkemp-llvm16 = mkDiffkemp pkgs.llvmPackages_16;
+        diffkemp-llvm15 = mkDiffkemp pkgs.llvmPackages_15;
+        diffkemp-llvm14 = mkDiffkemp pkgs.llvmPackages_14;
+        diffkemp-llvm13 = mkDiffkemp pkgs.llvmPackages_13;
+        diffkemp-llvm12 = mkDiffkemp pkgs.llvmPackages_12;
+        diffkemp-llvm11 = mkDiffkemp pkgs.llvmPackages_11;
+        diffkemp-llvm10 = mkDiffkemp pkgs.llvmPackages_10;
+        diffkemp-llvm9 = mkDiffkemp pkgs.llvmPackages_9;
       };
 
-      # Environment for downloading and preparing test kernels (RHEL 7 and 8).
-      # Contains 2 changes from the default env necessary for RHEL 7:
-      #  - gcc 7
-      #  - make 3.81
-      # It should be sufficient to use this to download the kernels (with
-      # rhel-kernel-get) and the tests can be run in the default dev shell.
-      test-kernel-buildenv = with pkgs;
-        let
-          oldmake = import
-            (builtins.fetchTarball {
-              url = "https://github.com/NixOS/nixpkgs/archive/92487043aef07f620034af9caa566adecd4a252b.tar.gz";
-              sha256 = "00fgvpj0aqmq45xmmiqr2kvdir6zigyasx130rp96hf35mab1n8c";
-            })
-            { inherit system; };
-          gnumake381 = oldmake.gnumake381;
+      devShells.${system} = rec {
+        default = diffkemp-llvm16;
 
-          default = self.devShells.${system}.default;
-        in
+        diffkemp-llvm16 = mkDiffkempDevShell self.packages.${system}.diffkemp-llvm16;
+        diffkemp-llvm15 = mkDiffkempDevShell self.packages.${system}.diffkemp-llvm15;
+        diffkemp-llvm14 = mkDiffkempDevShell self.packages.${system}.diffkemp-llvm14;
+        diffkemp-llvm13 = mkDiffkempDevShell self.packages.${system}.diffkemp-llvm13;
+        diffkemp-llvm12 = mkDiffkempDevShell self.packages.${system}.diffkemp-llvm12;
+        diffkemp-llvm11 = mkDiffkempDevShell self.packages.${system}.diffkemp-llvm11;
+        diffkemp-llvm10 = mkDiffkempDevShell self.packages.${system}.diffkemp-llvm10;
+        diffkemp-llvm9 = mkDiffkempDevShell self.packages.${system}.diffkemp-llvm9;
 
-        gcc7Stdenv.mkDerivation {
-          # gcc7Stdenv provides GCC 7, however it doesn't provide the mkShell
-          # command so we need to use mkDerivation and clear phases (that's what
-          # mkShell does).
-          name = "test-kernel-buildenv";
-          phases = [ ];
+        # Environment for downloading and preparing test kernels (RHEL 7 and 8).
+        # Contains 2 changes from the default env necessary for RHEL 7:
+        #  - gcc 7
+        #  - make 3.81
+        # It should be sufficient to use this to download the kernels (with
+        # rhel-kernel-get) and the tests can be run in the default dev shell.
+        test-kernel-buildenv =
+          with pkgs;
+          let
+            oldmake = import
+              (builtins.fetchTarball {
+                url = "https://github.com/NixOS/nixpkgs/archive/92487043aef07f620034af9caa566adecd4a252b.tar.gz";
+                sha256 = "00fgvpj0aqmq45xmmiqr2kvdir6zigyasx130rp96hf35mab1n8c";
+              })
+              { inherit system; };
+            gnumake381 = oldmake.gnumake381;
+          in
 
-          nativeBuildInputs = lib.lists.remove gcc default.nativeBuildInputs;
+          gcc7Stdenv.mkDerivation {
+            # gcc7Stdenv provides GCC 7, however it doesn't provide the mkShell
+            # command so we need to use mkDerivation and clear phases (that's what
+            # mkShell does).
+            name = "test-kernel-buildenv";
+            phases = [ ];
 
-          buildInputs = lib.lists.remove gnumake default.buildInputs ++ [
-            gnumake381
-          ];
+            nativeBuildInputs = lib.lists.remove gcc default.nativeBuildInputs;
 
-          propagatedBuildInputs = default.propagatedBuildInputs;
+            buildInputs = lib.lists.remove gnumake default.buildInputs ++ [
+              gnumake381
+            ];
 
-          dontUseSetuptoolsShellHook = true;
-        };
+            propagatedBuildInputs = default.propagatedBuildInputs;
+
+            dontUseSetuptoolsShellHook = true;
+          };
+      };
     };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -8,6 +8,10 @@
       system = "x86_64-linux";
       pkgs = import nixpkgs { inherit system; };
 
+      llvmVersionMin = 9;
+      llvmVersionMax = 16;
+      llvmVersions = pkgs.lib.lists.range llvmVersionMin llvmVersionMax;
+
       mkDiffkemp =
         llvmPackages:
           with pkgs;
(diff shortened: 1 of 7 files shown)


zacikpa:
I haven't yet studied the source code much, but I tried running it, and I'm a bit confused about what the `nix develop` command does.

- Say I would like to develop using the latest LLVM. I haven't yet created any `build` directory.
- When I run `nix develop`, it creates a *binary executable file* named `build` (I have no idea what it contains) and subsequently tries to install diffkemp. This fails, because it wants to create a *directory* named `build`.
- Even if the `build` directory is created before running `nix develop`, I don't understand why the command attempts to install diffkemp when SimpLL is not yet built.

viktormalik:
Ah, that's a problem of Setuptools. One of the problems of Nix is that `/nix/store/` is an immutable file system, therefore you cannot do `pip install` inside the development shell. To overcome this, there's a built-in function `setuptoolsShellHook` which does some tricks and installs the `diffkemp` Python package. The problem is that the installation builds the CFFI shared object in the `build` directory and if that directory doesn't exist, it apparently creates the shared object with name `build`. Since `setuptoolsShellHook` is run by default with each `nix develop`, there are problems when DiffKemp is not built in the `build` directory at the time of running `nix develop`.

There are several solutions to that:
- We don't run `setuptoolsShellHook` by default. This is probably a good approach b/c it has to be run manually after building SimpLL anyways, but I'm not sure if it will be ok with the CI (I wasn't able to do `nix develop --command setuptoolsShellHook`).
- We keep running `setuptoolsShellHook` by default and document the workflow, which would be:
```
$ nix develop
$ rm -rf build && mkdir build  # not necessary if build/ existed before
$ cmake -B build . -GNinja ...
$ ninja -C build
$ setupToolsShellHook
```

zacikpa:
Oh, I understand now.

> We keep running setuptoolsShellHook by default and document the workflow, which would be:
> ```
> $ nix develop
> $ rm -rf build && mkdir build  # not necessary if build/ existed before
> $ cmake -B build . -GNinja ...
> $ ninja -C build
> $ setupToolsShellHook
> ```

I'm ok with this approach, at least until we find some cleaner solution.

I will try to make a more thorough review soon.

viktormalik:
> I'm ok with this approach, at least until we find some cleaner solution.
> 
> I will try to make a more thorough review soon.

Thanks! I'll try to come up with a solution for the CI do that we can get rid of the default `setuptoolsShellHook`.

viktormalik:
So I've come up with a bit clumsy solution but it's the best I have for now. It disables `setuptoolsShellHook` by default but performs some magical export commands which will make sure that it can be called manually from inside the shell as well as by `nix develop --command bash -c setuptoolsShellHook` (for running in CI). See the last commit for details.

I've also updated README with instructions to use `setuptoolsShellHook` instead of `pip install .` inside Nix dev shell.

PLukas2018:
> The main aim is to replace the Docker container as the default development environment. The workflow with Nix is:
> 
>    1. Install Nix (see instructions in updated README).
> 
>    2. Download kernels for tests:
> 
> ```
> [localhost]$ nix develop .#test-kernel-buildenv
> [test-kernel-buildenv]$ rhel-kernel-get 3.10.0-514.el7 --output-dir kernel --kabi
> [...] # repeat for all necessary kernels
> [test-kernel-buildenv]$ exit
> ```
>
>    3. Build DiffKemp and run tests, either in the development flake or even on localhost (I believe that the latter requires the same version of glibc as is in Nix).
> 
> 
> ```
> [localhost]$ nix develop        # sometimes not even necessary
> [diffkemp-llvm16/localhost]$    # build DiffKemp
> [diffkemp-llvm16/localhost]$ pytest tests
> ```
>  I'd love if someone could try and confirm that this works.

I tried this and it worked for me (all tests passed). Downloading the kernels using `nix develop .#test-kernel-buildenv` and running the tests on my local machine did not work for me.

flake.nix
```diff
@@ -0,0 +1,177 @@
...
+            buildInputs = lib.lists.remove gnumake default.buildInputs ++ [
+              gnumake381
+            ];
+
+            propagatedBuildInputs = default.propagatedBuildInputs;
+
+            dontUseSetuptoolsShellHook = true;
+          };
+      };
+    };
```
zacikpa, line 176: I was playing with Nix a bit and found out that we could define the set of all development environments like this, and we wouldn't have to update multiple lines with each new LLVM version:
```suggestion
    in {
      formatter.${system} = pkgs.nixpkgs-fmt;

      packages.${system} = (builtins.listToAttrs (map (ver: {
        name = "diffkemp-llvm${ver}";
        value = mkDiffkemp pkgs."llvmPackages_${ver}";
      }) llvmVersionStrings)) // {
        default = mkDiffkemp pkgs."llvmPackages_${llvmLastVersionString}";
      };

      devShells.${system} = (builtins.listToAttrs (map (ver: {
        name = "diffkemp-llvm${ver}";
        value =
          mkDiffkempDevShell self.packages.${system}."diffkemp-llvm${ver}";
      }) llvmVersionStrings)) // {
        default = mkDiffkempDevShell
          self.packages.${system}."diffkemp-llvm${llvmLastVersionString}";
      } // {

        # Environment for downloading and preparing test kernels (RHEL 7 and 8).
        # Contains 2 changes from the default env necessary for RHEL 7:
        #  - gcc 7
        #  - make 3.81
        # It should be sufficient to use this to download the kernels (with
        # rhel-kernel-get) and the tests can be run in the default dev shell.
        test-kernel-buildenv = with pkgs;
          let
            oldmake = import (builtins.fetchTarball {
              url =
                "https://github.com/NixOS/nixpkgs/archive/92487043aef07f620034af9caa566adecd4a252b.tar.gz";
              sha256 = "00fgvpj0aqmq45xmmiqr2kvdir6zigyasx130rp96hf35mab1n8c";
            }) { inherit system; };
            gnumake381 = oldmake.gnumake381;

          in gcc7Stdenv.mkDerivation {
            # gcc7Stdenv provides GCC 7, however it doesn't provide the mkShell
            # command so we need to use mkDerivation and clear phases (that's what
            # mkShell does).
            name = "test-kernel-buildenv";
            phases = [ ];

            nativeBuildInputs = lib.lists.remove gcc default.nativeBuildInputs;

            buildInputs = lib.lists.remove gnumake default.buildInputs
              ++ [ gnumake381 ];

            propagatedBuildInputs = default.propagatedBuildInputs;

            dontUseSetuptoolsShellHook = true;
          };
      };
    };
}
```
Of course, we would have to define some variables:
```nix
llvmVersions = pkgs.lib.lists.range 9 16;
llvmVersionStrings = map toString llvmVersions;
llvmLastVersionString = pkgs.lib.lists.last llvmVersionStrings;
```
With each new LLVM version being packaged in Nix, we would only update the range in `llvmVersions`.

On the other hand, when I started playing with this, I thought the end result would look much cleaner than it does :smile: So feel free to stick with the original version if you don't like this.

zacikpa:
> So I've come up with a bit clumsy solution but it's the best I have for now. It disables `setuptoolsShellHook` by default but performs some magical export commands which will make sure that it can be called manually from inside the shell as well as by `nix develop --command bash -c setuptoolsShellHook` (for running in CI). See the last commit for details.
> 
> I've also updated README with instructions to use `setuptoolsShellHook` instead of `pip install .` inside Nix dev shell.

Great, the workflow is much cleaner this way!

I tried building Diffkemp with two different LLVM versions, downloading the kernels, and running the tests - everything worked fine for me.

The only issue I encountered was with `RPython` - I had it installed locally, and `CMake` tried to run this local installation from within the Nix environment to build the `cc-wrapper`, which failed. When I uninstall it, CMake detects that it's not available and does not even try building the wrapper, so the build succeeds (as in the CI).

flake.nix
viktormalik: This actually looks quite nice (and easier to maintain when we add new LLVM versions), thanks! Let me give this a try.

viktormalik:
> The only issue I encountered was with `RPython` - I had it installed locally, and `CMake` tried to run this local installation from within the Nix environment to build the `cc-wrapper`, which failed. When I uninstall it, CMake detects that it's not available and does not even try building the wrapper, so the build succeeds (as in the CI).

Right, I forgot about RPython, let me fix that.

viktormalik:
> Downloading the kernels using `nix develop .#test-kernel-buildenv` and running the tests on my local machine did not work for me.

What exactly failed? Could you try to run various DiffKemp commands (after Nix build) and report the errors? Also don't forget to `pip install`, that's always done only locally in Nix env.

PLukas2018:
> What exactly failed? Could you try to run various DiffKemp commands (after Nix build) and report the errors? Also don't forget to `pip install`, that's always done only locally in Nix env.

I will check again and inform you, maybe I just did something wrong or it could be as you mentioned that I do not have same glibc version as nix (I have `2.35` , nix has `2.37`).

Also, I could/will try to run the EqBench benchmark using `result/bin/diffkemp ...` to check if I get the same (or similar) results as running it normally.

Also, I am thinking if we do not want to include in the nix libraries required for running/developing the result viewer -- `node`, `npm`?

viktormalik:
> > What exactly failed? Could you try to run various DiffKemp commands (after Nix build) and report the errors? Also don't forget to `pip install`, that's always done only locally in Nix env.
> 
> I will check again and inform you, maybe I just did something wrong or it could be as you mentioned that I do not have same glibc version as nix (I have `2.35` , nix has `2.37`).
> 
> Also, I could/will try to run the EqBench benchmark using `result/bin/diffkemp ...` to check if I get the same (or similar) results as running it normally.

Thanks, that would be useful!

> Also, I am thinking if we do not want to include in the nix libraries required for running/developing the result viewer -- `node`, `npm`?

My plan was to add dependencies for the viewer later but it's true that it may be as simple as adding those two packages. Let me give it a try.

flake.nix
DanielKriz: I also think, that this looks nicer and more maintainable. I was thinking and some packages in `nixpkgs` are using some kind of "latest" version variable, perhaps we also could use something like that here?

```nix
llvmLatestVersion = 16;
llvmVersions = pkgs.lib.lists.range 9 llvmLatestVersion;
llvmVersionStrings = map toString llvmVersions;
```

And if I am correct (but I am not sure at the moment), then we should be able to use this `llvmLatestVersion` in the format strings without casting.

I have not tried running it, so its more of an idea/suggestion.

DanielKriz:
I am able to build the DiffKemp with nix without any problem. However, I seem to have some issue with `rhel-kernel-get`, it may be unrelated to nix.

It also should be noted that I have been using `podman` with `nixos/nix` and also `nixos` virtual machine for the testing, as installation of `nix` on fedora requires some changes in default SELinux.

I ma going to try this in some other machine/VM/container.

Here is the output of `rhel-kernel-get` command. (I did check whether the download haven't failed, but it did not, the kernel sources were indeed located in `/tmp/nix-shell.3pfEsS/tmp4vfzrt6p/linux-3.10.0-514.el7`).

[rhel_kernel_get_out.log](https://github.com/viktormalik/diffkemp/files/13079637/rhel_kernel_get_out.log)

flake.nix
viktormalik: Thanks, I've implemented this (with some modifications).

viktormalik:
> > The only issue I encountered was with `RPython` - I had it installed locally, and `CMake` tried to run this local installation from within the Nix environment to build the `cc-wrapper`, which failed. When I uninstall it, CMake detects that it's not available and does not even try building the wrapper, so the build succeeds (as in the CI).
> 
> Right, I forgot about RPython, let me fix that.

So, the thing is that RPython is not in nixpkgs because it's for Python 2 and that has been deprecated a while ago. Trying to workaround this seems to require a lot of (probably hacky) work so, instead, I've introduced a new env variable `WITHOUT_RPYTHON` which, when set, disables RPython in CMake completely. The variable is now set in Nix envs.

viktormalik:
> > Also, I am thinking if we do not want to include in the nix libraries required for running/developing the result viewer -- `node`, `npm`?
> 
> My plan was to add dependencies for the viewer later but it's true that it may be as simple as adding those two packages. Let me give it a try.

It turns out that adding `node` and `npm` is the matter of a single package so I've added it in this PR. On my setup, running both production and development version of the viewer from `nix develop` worked.

viktormalik:
> Here is the output of `rhel-kernel-get` command. (I did check whether the download haven't failed, but it did not, the kernel sources were indeed located in `/tmp/nix-shell.3pfEsS/tmp4vfzrt6p/linux-3.10.0-514.el7`).
> 
> [rhel_kernel_get_out.log](https://github.com/viktormalik/diffkemp/files/13079637/rhel_kernel_get_out.log)

Interesting, this looks like the source tree is corrupted somehow. Could you try to go into the source directory and run `make olddefconfig` directly from there?

flake.nix
```diff
@@ -0,0 +1,189 @@
...
+              cffi
+              pyyaml
+              setuptools
+            ];
+
+            WITHOUT_RPYTHON = true;
+
+            # Including cmake in nativeBuildInputs automatically runs it during
+            # configurePhase so we just need to set correct flags.
+            cmakeFlags = [ "-DCMAKE_BUILD_TYPE=Debug" "-GNinja" "-DBUILD_VIEWER=On" ];
```
PLukas2018, line 50:
```suggestion
            cmakeFlags = [ "-DCMAKE_BUILD_TYPE=Debug" "-GNinja" ];
```

PLukas2018 (commented):
> It turns out that adding node and npm is the matter of a single package so I've added it in this PR. On my setup, running both production and development version of the viewer from nix develop worked.

I tried the viewer and for the `nix develop` both the production and development versions of the viewer worked for me too, but I was unable to do `nix build`, it failed in installation ([nix-build.log](https://github.com/viktormalik/diffkemp/files/13118569/nix-build.log)) -- I think this could be caused by the installation destination of the viewer which is `/var/lib/diffkemp/view`. For now, I think we could remove the installation/building of the viewer when `nix build` is executed and add that later.

flake.nix
viktormalik: Good catch!

Agreed - let's disable the viewer in build for now to unblock this and come back to it later.

PLukas2018:
I tried again running `pytest tests` on my 'localhost' (after downloading it with `nix develop .#test-kernel-buildenv`, all tests passed (I probably did something wrong the first time I tried this).

> > Also, I could/will try to run the EqBench benchmark using `result/bin/diffkemp ...` to check if I get the same (or similar) results as running it normally.
> 
> Thanks, that would be useful!

I tested this for the same llvm version which I have on my 'localhost' and I got the same results for `result/bin/diffkemp` as running it normally.

viktormalik:
> I tried again running `pytest tests` on my 'localhost' (after downloading it with `nix develop .#test-kernel-buildenv`, all tests passed (I probably did something wrong the first time I tried this).
> 
> > > Also, I could/will try to run the EqBench benchmark using `result/bin/diffkemp ...` to check if I get the same (or similar) results as running it normally.
> > 
> > 
> > Thanks, that would be useful!
> 
> I tested this for the same llvm version which I have on my 'localhost' and I got the same results for `result/bin/diffkemp` as running it normally.

Many thanks! Feel free to approve the PR once you see it fit.

zacikpa:
I went through the commits once more, and I haven't found any more issues, except the missing `result/` in `.gitignore`. Can you please add that, @viktormalik? I will approve then.

viktormalik:
> I went through the commits once more, and I haven't found any more issues, except the missing `result/` in `.gitignore`. Can you please add that, @viktormalik? I will approve then.

Done.

zacikpa (approved):

DanielKriz:
> > Here is the output of `rhel-kernel-get` command. (I did check whether the download haven't failed, but it did not, the kernel sources were indeed located in `/tmp/nix-shell.3pfEsS/tmp4vfzrt6p/linux-3.10.0-514.el7`).
> > [rhel_kernel_get_out.log](https://github.com/viktormalik/diffkemp/files/13079637/rhel_kernel_get_out.log)
> 
> Interesting, this looks like the source tree is corrupted somehow. Could you try to go into the source directory and run `make olddefconfig` directly from there?

That seems to work without an issue. This might be more an issue with `nixos` than with `nix`. I am going to try running different linux distro in a container and installing only nix there.

But otherwise I had no problem with nix-based build.

PLukas2018 (approved):
It looks good to me too, I did not find any other issue.

flake.nix
```diff
@@ -0,0 +1,189 @@
...
+              diffutils
+              gtest
+              gnumake
+            ];
+
+            propagatedBuildInputs = with python3Packages; [
+              cffi
+              pyyaml
+              setuptools
+            ];
```
PLukas2018, line 44: One more thing that occurred to me is that we will have to update this list when we add something to `requierments.txt`. I do not know if it is possible that the nix would take the list of packages from the `requirements.txt` file?

DanielKriz: https://nix-community.github.io/dream2nix/options/pip.html#opt-pip.requirementsFiles

There is a project that tries to automate it.

viktormalik: This is a good point, thanks @DanielKriz for the reference. I wouldn't complicate things at this point, though, as CI should capture this in most cases. But we can certainly revisit this later, there's also a widely-used [poetry2nix](https://github.com/nix-community/poetry2nix) for converting Python packages to Nix but that doesn't work for setuptools-based projects.

PLukas2018: Thank  @DanielKriz. @viktormalik you are right that CI should capture it, I did not think of that.

viktormalik:
> That seems to work without an issue. This might be more an issue with `nixos` than with `nix`. I am going to try running different linux distro in a container and installing only nix there.

@DanielKriz kindly ping, do you still want to try this or can we merge?

DanielKriz:
Sorry, I totally forgot, that I did write that comment. I through that I am still waiting for the reaction about the `rhel-kernel-get` problem.

I did test it with an `ubuntu` container and everything works. There is just some problem with NixOS, which probably is not as important at the moment.

I have only one small suggestion wouldn't `kernel-test-buildenv` be a bit better name? Until now I was thinking that `test-kernel-buildenv` is some nix testing environment.

Other than that it is working and it is great.

DanielKriz (approved):

PR #301, opened 2023-10-21
Built-in pattern matching reordered commutative operations

zacikpa:
This PR adds a pattern that can match reordered commutative binary instructions, such as:
```llvm
;old
%tmp = add i8 0,    i8 1
%res = add i8 %tmp, i8 2
;new
%tmp = add i8 1,    i8 2
%res = add i8 %tmp, i8 0
```
I also fixed two minor issues that would sometimes lead to problems for this pattern and for the upcoming structure refactoring pattern; see commit messages for details.

I assume that this pattern can be safely removed once such refactorings are covered with an SMT solver.

Tests should cover the most basic use cases.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -502,18 +502,18 @@ int DifferentialFunctionComparator::cmpAllocs(const CallInst *CL,
 bool DifferentialFunctionComparator::maySkipInstruction(
         const Instruction *Inst) const {
     if (isa<AllocaInst>(Inst)) {
-        // Ignore AllocaInsts with no specific replacement.
+        ignoredInstructions.insert(Inst);
         return true;
     }
     if (isCast(Inst)) {
         if (config.Patterns.TypeCasts) {
-            ignoredInstructions.insert({Inst, Inst->getOperand(0)});
+            replacedInstructions.insert({Inst, Inst->getOperand(0)});
             return true;
         }
         return maySkipCast(Inst);
     }
     if (isZeroGEP(Inst)) {
-        ignoredInstructions.insert({Inst, Inst->getOperand(0)});
+        replacedInstructions.insert({Inst, Inst->getOperand(0)});
         return true;
     }
     if (auto Load = dyn_cast<LoadInst>(Inst)) {
@@ -536,7 +536,7 @@ bool DifferentialFunctionComparator::maySkipCast(const User *Cast) const {
             // replacement of a simple type with a union, the
             // comparator would detect a difference when the value is
             // used, since it would have a different type.
-            ignoredInstructions.insert({Cast, Cast->getOperand(0)});
+            replacedInstructions.insert({Cast, Cast->getOperand(0)});
             return true;
         }
     }
@@ -545,7 +545,7 @@ bool DifferentialFunctionComparator::maySkipCast(const User *Cast) const {
         // an instruction affected by this (i.e. a store, load or GEP),
         // there would be an additional difference that will be detected
         // later.
-        ignoredInstructions.insert({Cast, Cast->getOperand(0)});
+        replacedInstructions.insert({Cast, Cast->getOperand(0)});
         return true;
     }
     if (SrcTy->isIntegerTy() && DestTy->isIntegerTy()) {
@@ -576,7 +576,7 @@ bool DifferentialFunctionComparator::maySkipCast(const User *Cast) const {
                         UserStack.push_back(UU);
                 }
             }
-            ignoredInstructions.insert({Cast, Cast->getOperand(0)});
+            replacedInstructions.insert({Cast, Cast->getOperand(0)});
             return true;
         }
     }
@@ -654,7 +654,7 @@ bool DifferentialFunctionComparator::maySkipLoad(const LoadInst *Load) const {
     // If the load is repeated without stores in between, skip it because
     // the load is redundant and its removal will cause no semantic difference.
     if (PreviousLoad) {
-        ignoredInstructions.insert({Load, PreviousLoad});
+        replacedInstructions.insert({Load, PreviousLoad});
         return true;
     }
     return false;
@@ -669,8 +669,8 @@ bool mayIgnoreMacro(std::string macro) {
 const Value *DifferentialFunctionComparator::getReplacementValue(
         const Value *Replaced, DenseMap<const Value *, int> &sn_map) const {
     // Find the replacement value.
-    auto replacementIt = ignoredInstructions.find(Replaced);
-    if (replacementIt == ignoredInstructions.end()) {
+    auto replacementIt = replacedInstructions.find(Replaced);
+    if (replacementIt == replacedInstructions.end()) {
         // Before failing, check whether the replaced value is an ignorable
         // bitcast or zero GEP operator. If so, return its operand. Note that if
         // the replacing operand is an instruction, it must already have been
@@ -908,12 +908,12 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
                 if (isLogicalNot(&*InstL)) {
                     matchingPair = {&*InstL,
                                     getMappedValue(InstL->getOperand(0), true)};
-                    ignoredInstructions.emplace(&*InstL, InstL->getOperand(0));
+                    replacedInstructions.emplace(&*InstL, InstL->getOperand(0));
                     InstL++;
                 } else {
                     matchingPair = {getMappedValue(InstR->getOperand(0), false),
                                     &*InstR};
-                    ignoredInstructions.emplace(&*InstR, InstR->getOperand(0));
+                    replacedInstructions.emplace(&*InstR, InstR->getOperand(0));
                     InstR++;
                 }
 
@@ -1352,7 +1352,15 @@ int DifferentialFunctionComparator::cmpValues(const Value *L,
                                               const Value *R) const {
     PREP_LOG("value", L, R);
 
-    // Use replacement references for ignored values.
+    // Instructions that were ignored without replacement must never be compared
+    // as values. Two such instructions would always synchronize (even if not
+    // equal). Comparing with an already synchronized value would pollute the
+    // SN maps and complicate eventual undo.
+    if (ignoredInstructions.count(L) || ignoredInstructions.count(R)) {
+        RETURN_WITH_LOG_NEQ(1);
+    }
+
+    // Use replacements for other ignored values, bitcasts, and zero GEPs.
     auto replaceL = getReplacementValue(L, sn_mapL);
     auto replaceR = getReplacementValue(R, sn_mapR);
     if (replaceL || replaceR) {
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -505,6 +505,11 @@ bool DifferentialFunctionComparator::maySkipInstruction(
         ignoredInstructions.insert(Inst);
         return true;
     }
+    if (config.Patterns.ReorderedBinOps && isReorderableBinaryOp(Inst)
+        && maySkipReorderableBinaryOp(Inst)) {
+        ignoredInstructions.insert(Inst);
+        return true;
+    }
     if (isCast(Inst)) {
         if (config.Patterns.TypeCasts) {
             replacedInstructions.insert({Inst, Inst->getOperand(0)});
@@ -658,6 +663,17 @@ bool DifferentialFunctionComparator::maySkipLoad(const LoadInst *Load) const {
         return true;
     }
     return false;
+}
+
+/// Check whether the given reorderable binary operator can be skipped.
+/// It can only be skipped if all its users are binary operations
+/// of the same kind.
+bool DifferentialFunctionComparator::maySkipReorderableBinaryOp(
+        const Instruction *Op) const {
+    return std::all_of(Op->user_begin(), Op->user_end(), [Op](auto user) {
+        auto userOp = dyn_cast<BinaryOperator>(user);
+        return userOp && userOp->getOpcode() == Op->getOpcode();
+    });
 }
 
 bool mayIgnoreMacro(std::string macro) {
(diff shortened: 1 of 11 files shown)


PLukas2018:
I tried this on a program from the EqBench benchmark which was not evaluated correctly and this pattern should help in the correct evaluation. Unfortunately, this pattern did not manage it. The problematic basic block:

I think that maybe this case is a little more complex,

PLukas2018:
I tried this on a program from the EqBench benchmark which was not evaluated correctly before and this pattern should help in the correct evaluation. Unfortunately, this pattern did not manage it. The problematic basic block is:

```llvm
# old:
17:                                               ; preds = %14
  %18 = fmul double %0, %0, !dbg !53
  %19 = tail call double @llvm.fmuladd.f64(double %18, double 0xBFD2492492492492, double 1.000000e+00), !dbg !55
  %20 = fmul double %18, -4.000000e-01, !dbg !56
  %21 = tail call double @llvm.fmuladd.f64(double %20, double %19, double 1.000000e+00), !dbg !56
  %22 = fmul double %18, 0xBFE5555555555555, !dbg !57
  %23 = tail call double @llvm.fmuladd.f64(double %22, double %21, double 1.000000e+00), !dbg !57
  %24 = fmul double %23, %0, !dbg !58
  br label %67, !dbg !59
```

```llvm
# new:
17:                                               ; preds = %14
  %18 = fmul double %0, %0, !dbg !54
  %19 = tail call double @llvm.fmuladd.f64(double %18, double 0xBFD2492492492492, double 1.000000e+00), !dbg !56
  %20 = fmul double %18, -4.000000e-01, !dbg !57
  %21 = tail call double @llvm.fmuladd.f64(double %20, double %19, double 1.000000e+00), !dbg !57
  %22 = fmul double %0, 0xBFE5555555555555, !dbg !58
  %23 = fmul double %22, %0, !dbg !58
  %24 = tail call double @llvm.fmuladd.f64(double %23, double %21, double 1.000000e+00), !dbg !58
  %25 = fmul double %24, %0, !dbg !59
  br label %68, !dbg !60
```

I think that maybe this case is a little more complex, because `%18` is used in both versions, but in the old version it is also used as a part of the commutative operation (not sure if this is the reason -- I did not study implementation into depth).

Otherwise, thanks to this pattern DiffKemp managed to correctly evaluate 2 more programs, which it did not manage it before when compiled with `-O1` and llvm passes which are used in `build-kernel` command, but it managed it before when compiled with `-O2`  and default `-O2` passes.

zacikpa:
Thank you, Lukáš, for the analysis. 

This is a very interesting example which I did not think about. You are right that the problem lies in `%18`, but there actually may be an easy way of solving it.

zacikpa:
@PLukas2018, can you try evaluating the EqBench program again (or point me to the source)?

The new commit should cover this case (although it makes the pattern less efficient).

PLukas2018:
> @PLukas2018, can you try evaluating the EqBench program again (or point me to the source)?
> 
> The new commit should cover this case (although it makes the pattern less efficient).

Unfortunately, it looks like it isn't still able to manage the program, I may be missing something. 

If you want to try it yourself, here are the [old](https://github.com/shrBadihi/EqBench/blob/main/benchmarks/bess/dawson/Eq/oldV.c) and [new](https://github.com/shrBadihi/EqBench/blob/main/benchmarks/bess/dawson/Eq/newV.c) versions of the program. I created a snapshot with `bin/diffkemp build --clang-append=-O2 ...`  (default built-in patterns with `-O2` optimisation), but it looks like it is the same for `-O1` optimisation. Then I compared the `snippet` function. The problematic change is 32-33/35 (old/new), it actually contains a replacement of a variable with an expression (its value), but when compiled the problem is the one which I mentioned above.

zacikpa:
Thanks, I will take another look at it.

zacikpa:
Oh. So the problem seems to be that, regardless of the optimization level, Clang compiles the following code
```c
float x = a * b + c;
```
into
```llvm
%x = call float @llvm.fmuladd.f32(%a, %b, %c)
```
which is unfortunately an intrinsic instruction that I cannot currently deal with (and probably won't).

lenticularis39:
Can't we just do an intrinsic expansion pass for that?

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1845,3 +1895,49 @@ bool DifferentialFunctionComparator::isDependingOnReloc(
...
+        std::multiset<std::string> &Constants,
+        DenseMap<const Value *, int> &sn_map) const {
+    // Substitute an ignored value with its replacement when necessary
+    auto replacementIt = ignoredWithReplacement.find(Val);
+    const Value *newVal = replacementIt == ignoredWithReplacement.end()
+                                  ? Val
+                                  : replacementIt->second;
+    // Constant leaf
+    if (auto Const = dyn_cast<ConstantInt>(newVal)) {
+        Constants.insert(valueAsString(Const));
```
viktormalik, line 1915: Why not use the value directly?

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1845,3 +1895,49 @@ bool DifferentialFunctionComparator::isDependingOnReloc(
...
+    if (BinOp && BinOp->getOpcode() == Opcode) {
+        std::multiset<int> tmpSNs;
+        std::multiset<std::string> tmpConstants;
+        if (collectBinaryOperands(
+                    BinOp->getOperand(0), Opcode, tmpSNs, tmpConstants, sn_map)
+            && collectBinaryOperands(BinOp->getOperand(1),
+                                     Opcode,
+                                     tmpSNs,
+                                     tmpConstants,
+                                     sn_map)) {
```
viktormalik, line 1929: Not very nice formatting, can we do something about it?

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1845,3 +1895,49 @@ bool DifferentialFunctionComparator::isDependingOnReloc(
...
+            SNs.merge(tmpSNs);
+            Constants.merge(tmpConstants);
+            return true;
+        }
+    }
+    // Already synchronized leaf
+    auto snMapIt = sn_map.find(newVal);
+    if (snMapIt != sn_map.end()
+        && (unsigned)snMapIt->second != sn_map.size() - 1) {
```
viktormalik, line 1938: IIUC, this is b/c we don't want to match the last-inserted value which comes from the currently processed non-equal pair. Maybe mention that in the comment?

viktormalik (changes requested):
Mathematically speaking, the pattern covers operations which are **commutative and associative** (is there a name for that?). Probably not worth changing the pattern name (unless we come up with something compact) but we should at least mention it in the comments.

lenticularis39:
> Mathematically speaking, the pattern covers operations which are **commutative and associative** (is there a name for that?)

I don't think there is a specific name for it, such operations together with their domain are just called commutative semigroups.

The fact that it also matches associative reordering could be a problem though for floating-point operations, these are not strictly associative because of rounding. E.g. in GCC optimizations they are not treated as such unless you enable `-ffast-math`.

viktormalik:
> The fact that it also matches associative reordering could be a problem though for floating-point operations, these are not strictly associative because of rounding. E.g. in GCC optimizations they are not treated as such unless you enable `-ffast-math`.

A very good point. It shouldn't be a problem for this PR b/c it doesn't apply the pattern on floating-point instructions (`fadd`, `fmul`) but certainly something that we should bear in mind should we try to do so in future.

lenticularis39:
> A very good point. It shouldn't be a problem for this PR b/c it doesn't apply the pattern on floating-point instructions (`fadd`, `fmul`) but certainly something that we should bear in mind should we try to do so in future.

I see `Instruction:FAdd` and `Instruction:FMul` in `isCommutativeBinaryOp` though, is it really not applied on these? (I haven't gone through the code, maybe I'm misunderstanding something.)

zacikpa:
As @lenticularis39 says, the pattern does apply to floating-point instructions too. Thanks for spotting the issue; I did not think about that.

viktormalik:
> I see `Instruction:FAdd` and `Instruction:FMul` in `isCommutativeBinaryOp` though, is it really not applied on these? (I haven't gone through the code, maybe I'm misunderstanding something.)

Sorry, my bad, I thought it wasn't there but should have double-checked.

@zacikpa can we remove `fadd` and `fmul` completely for now?

zacikpa:
Yes, I will do so.

zacikpa:
> Mathematically speaking, the pattern covers operations which are **commutative and associative** (is there a name for that?). Probably not worth changing the pattern name (unless we come up with something compact) but we should at least mention it in the comments.

Good point. I could rename it to something like `reordered-bin-ops`, which would be more explicit about what the pattern does (similarly to existing built-in pattern names, i.e., `function-splits` or `unused-returns`).

viktormalik:
> Good point. I could rename it to something like `reordered-bin-ops`, which would be more explicit about what the pattern does (similarly to existing built-in pattern names, i.e., `function-splits` or `unused-returns`).

Sounds quite nice and self-explanatory to me.

@lenticularis39 wdyt?

lenticularis39:
> @lenticularis39 wdyt?

I agree.

diffkemp/simpll/DifferentialFunctionComparator.cpp
zacikpa: Ah, good point. I think I wanted to avoid storing raw integer values because I thought a similar issue might occur as the one solved in https://github.com/viktormalik/diffkemp/pull/301/commits/dfccede263c2892f5507557373120374cf5a6eba (i.e., the constant being too large to fit into 64 bits). I did not realize that I could use the intermediate representation `llvm::APInt`.

I ran my experiments again, and the mentioned issue does not occur there (but I still think it could theoretically occur).

So there are three options:
1. Use `int64_t` and pray that such large integers do not get used in arithmetic operations.

2. Use `llvm::APInt`. However, `llvm:APInt`  does not implement operator `<` (because it has both signed and unsigned comparison functions) and thus cannot be used in `std::set` without creating a derived class or more obscure C++ magic (e.g., functors). 

3. Leave it as it is.

diffkemp/simpll/DifferentialFunctionComparator.h
```diff
@@ -138,11 +138,14 @@ class DifferentialFunctionComparator : public FunctionComparator {
     mutable std::vector<std::pair<const PHINode *, const PHINode *>>
             phisToCompare;
     mutable std::unordered_map<const Value *, const Value *>
-            ignoredInstructions;
+            ignoredWithReplacement;
+    mutable std::set<const Value *> ignoredNoReplacement;
```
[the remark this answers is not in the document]
zacikpa: Good idea, fixed.

diffkemp/simpll/Utils.cpp
```diff
@@ -257,10 +258,29 @@ bool isLogicalNot(const Instruction *Inst) {
     return false;
 }
 
+/// Returns true if the given instruction is a commutative binary operation
+bool isCommutativeBinaryOp(const Instruction *Inst) {
+    if (auto *BinOp = dyn_cast<BinaryOperator>(Inst)) {
+        std::set<Instruction::BinaryOps> CommutativeOps = {
```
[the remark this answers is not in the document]
zacikpa: Fixed.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1845,3 +1895,49 @@ bool DifferentialFunctionComparator::isDependingOnReloc(
...
+        SNs.insert(snMapIt->second);
+        return true;
+    }
+    return false;
+}
```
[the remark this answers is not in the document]
zacikpa: Fixed.

diffkemp/simpll/DifferentialFunctionComparator.cpp
zacikpa: Updated to use two variables to indicate success on each side of the operator.

diffkemp/simpll/DifferentialFunctionComparator.cpp
zacikpa: Yes, that's exactly the reason. Added a comment.

diffkemp/simpll/DifferentialFunctionComparator.cpp
viktormalik: Could we use 2. with `std::unordered_multiset`? I don't expect the sets to become too large for this to be a problem. Otherwise let's just stick with 1. for simplicity as the pattern should be replaced by a more complex solution anyways.

diffkemp/simpll/DifferentialFunctionComparator.h
```diff
@@ -138,11 +138,14 @@ class DifferentialFunctionComparator : public FunctionComparator {
     mutable std::vector<std::pair<const PHINode *, const PHINode *>>
             phisToCompare;
     mutable std::unordered_map<const Value *, const Value *>
-            ignoredInstructions;
+            replacedInstructions;
+    mutable std::set<const Value *> ignoredInstructions;
```
viktormalik, line 142:
```suggestion
    mutable std::unordered_set<const Value *> ignoredInstructions;
```
to unify with `replacedInstructions`

diffkemp/simpll/Utils.cpp
```diff
@@ -257,10 +258,28 @@ bool isLogicalNot(const Instruction *Inst) {
     return false;
 }
 
+/// Returns true if the given instruction is a reorderable binary operation,
+/// i.e., it is commutative and associative.
```
viktormalik, line 262: Let's add a note about floating-point operations not being associative to make sure we don't forget about this.

diffkemp/simpll/DifferentialFunctionComparator.h
```diff
@@ -138,11 +138,14 @@ class DifferentialFunctionComparator : public FunctionComparator {
...
+
     /// Contains pairs of values mapped by synchronisation maps. Enables
     /// retrieval of mapped values based on assigned numbers.
     mutable std::unordered_map<int, std::pair<const Value *, const Value *>>
             mappedValuesBySn;
+    mutable std::set<const Value *> binaryOpsWithSkippedOperands;
```
viktormalik, line 148: Unused?

zacikpa: Yep, thanks, removed.

diffkemp/simpll/DifferentialFunctionComparator.h
zacikpa: Fixed.

diffkemp/simpll/DifferentialFunctionComparator.cpp
zacikpa: Sticking with`int64_t` since `llvm::APInt` cannot be hashed by default.

zacikpa:
@viktormalik, all comments should be resolved. When you approve, I will squash the commits as agreed.

viktormalik (approved):
Nice work!

@lenticularis39 do you want to review this?

zacikpa:
Thanks for the review. I rebased onto master and squashed commits such that all functionality of the new pattern is in a single commit and can be easily reverted.

It should be easier to review like this for Tomáš (since he hasn't seen the code yet) if he decides to do so.

I moved the separation of `ignoredInstructions` into `ignoredInstructions` and `replacedInstructions` out of the main commit since I will use it in the structure pattern as well (and that one we do expect to keep).

lenticularis39:
> Nice work!
> 
> @lenticularis39 do you want to review this?

Nope, I won't have time to review this one in a reasonable time frame :D

Issue #302, opened 2023-10-30
Uninitialised value of `config.ExtendedStat`


zacikpa:
Valgrind reports the following issue when running unit tests:

```valgrind
==15614== Conditional jump or move depends on uninitialised value(s)
==15614==    at 0x4B6830: DifferentialFunctionComparator::cmpBasicBlocks(llvm::BasicBlock const*, llvm::BasicBlock const*) const (DifferentialFunctionComparator.cpp:859)
==15614==    by 0x42A2F3: TestComparator::testCmpBasicBlocks(llvm::BasicBlock*, llvm::BasicBlock*, bool) (DifferentialFunctionComparatorTest.cpp:90)
==15614==    by 0x416332: DifferentialFunctionComparatorTest_CmpBasicBlocksInlining_Test::TestBody() (DifferentialFunctionComparatorTest.cpp:922)
==15614==    by 0x48C7EB2: ??? (in /usr/lib64/libgtest.so.1.13.0)
==15614==    by 0x48B417D: testing::Test::Run() (in /usr/lib64/libgtest.so.1.13.0)
==15614==    by 0x48B4384: testing::TestInfo::Run() (in /usr/lib64/libgtest.so.1.13.0)
==15614==    by 0x48B44D6: testing::TestSuite::Run() (in /usr/lib64/libgtest.so.1.13.0)
==15614==    by 0x48C108F: testing::internal::UnitTestImpl::RunAllTests() (in /usr/lib64/libgtest.so.1.13.0)
==15614==    by 0x48BFE97: testing::UnitTest::Run() (in /usr/lib64/libgtest.so.1.13.0)
==15614==    by 0x40ECC0: RUN_ALL_TESTS() (gtest.h:2314)
==15614==    by 0x40ECAC: main (SimpLLTest.cpp:5)
```

Reproduce by running the following command from SimpLL build directory:
```bash
valgrind tests/unit_tests/simpll/runTests
```

PR #304, opened 2023-10-30
Initialize ExtendedStat in tests

FrNecas:
Fixes: #302


zacikpa (approved):
Thanks for the quick fix!

PR #305, opened 2023-11-01
Do not skip current instruction when matching custom patterns

zacikpa:
When matching custom patterns, we should not skip the currently compared instruction, i.e., the one that caused the semantic difference.

Imagine the following situation with two custom patterns:
```llvm
; Pattern 1
!0 = !{ !"pattern-start" }
declare void @diffkemp.new.skippable_1(ptr, i32)

define void @diffkemp.old.pattern_1(i8) {
  ret void
}

define void @diffkemp.new.pattern_1(i8) {
  %3 = alloca i32
  call void @diffkemp.new.skippable_1(ptr %3, i32 4), !diffkemp.pattern !0
  ret void
}
```
```llvm
; Pattern 2
!0 = !{ !"pattern-start" }
declare void @diffkemp.new.skippable_2(ptr)

define void @diffkemp.old.pattern_2(i8) {
  ret void
}

define void @diffkemp.new.pattern_2(i8) {
  %3 = alloca i32
  call void @diffkemp.new.skippable_2(ptr %3), !diffkemp.pattern !0
  ret void
}
```
The old version of the code does not contain the skippable function calls, whereas the new version contains both:
```llvm
; New version of the code
...
%1 = alloca i32
%2 = alloca i32
...
call void skippable_1(ptr %1, i32 4) ; currently compared instruction
...
call void skippable_2(ptr %2)
...
```
When the call to `skippable_1` is encountered and pattern matching starts, the behavior differs depending on which pattern we start matching first. If we start matching the second pattern first and skipping is enabled for the current module instruction, the pattern gets matched successfully against the call to `skippable_2`. However, the comparison itself fails immediately afterwards because we haven't matched the call to `skippable_1`.

Unfortunately, instruction skipping during pattern matching is not covered by any tests; when I disabled it completely, current regression tests still succeeded. @PSilling, do you remember any interesting cases which could be converted to regression tests, or even unit tests?

### DIFF
--- a/diffkemp/simpll/CustomPatternComparator.cpp
+++ b/diffkemp/simpll/CustomPatternComparator.cpp
@@ -236,3 +236,43 @@ void CustomPatternComparator::processPatternMatch(
         AllInstMatches.insert(InstPair.second);
     }
 }
+
+/// Populate the comparator maps with a given set of patterns.
+void CustomPatternComparator::addPatternSet(const CustomPatternSet *PatternSet,
+                                            const llvm::Function *FnL,
+                                            const llvm::Function *FnR) {
+    for (auto &InstPattern : PatternSet->InstPatterns) {
+        addInstPattern(InstPattern, FnL, FnR);
+    }
+    for (auto &ValuePattern : PatternSet->ValuePatterns) {
+        addValuePattern(ValuePattern, FnL, FnR);
+    }
+}
+
+/// Add an instruction pattern to the instruction pattern comparator map.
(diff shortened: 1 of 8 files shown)


viktormalik (approved):
Seems like a reasonable fix to me.

@zacikpa if you can come up with a simple unit test, that would be nice, otherwise I'm inclined to merge this as it is now.

zacikpa:
@viktormalik I added the test a couple days ago and updated it today, this should now be ready to merge after you look at the test.

viktormalik:
Looks good, thanks!

PR #306, opened 2023-11-06
SimpLL: Inverse condition pattern fixes

PLukas2018:
This PR fixes multiple problems with the inverse condition pattern:
1. The pattern could cause a false negative in a code with a combined condition because it always "skiped" a compare instruction with an inverse condition regardless of what the instruction users were. Example which caused false negative:
    ```c
    // old
    void f() {                                                               
        if (5 == 5 || 10 == 10) {                                            
            return true;                                                     
        }                                                                    
        else return false
    }
    // new
    void f() {                                                               
        if (5 != 5 || 10 != 10) {                                            
            return true;                                                     
        }                                                                    
        else return false
    }
    ```
2. It did not correctly handle inverse conditions with negations, because it searched for the current matching pair in the inverse condition list instead of the previous one when deciding if to add the pair to the list or remove it (more in [bc6abbe](https://github.com/viktormalik/diffkemp/commit/bc6abbe0f702c3b37687d3da8c5a76615dbeaa4e)). Eg it did not handle a case described in [212993d](https://github.com/viktormalik/diffkemp/commit/212993d):
    ```
    %2 = icmp eq %0, %1  | %2 = icmp ne %0, %
                         | %3 = xor %2, true
    br %2 %T %F          | br %3 %T %F
    ```
3. It did not check if an inverse condition pattern is enabled before checking if `not` instruction could not be a case of an inverse condition pattern.

* This PR also adds logging for the inverse condition pattern.

* Note: There is a problem with multiple `not` instructions which it does not handle properly because it would be necessary to find the original cmp instruction in the version of the function which does not have `not` instructions. It could be possible to fix it, but when I tried to compile programs to include the `not` (respectively `xor`) instruction -- I was unable to get such a program because the compiler usually just inverse the condition in the compare instruction `!(a!=b) -> a==b` or just swapped the branches. 

* While experimenting with it, I found out that it also has not been able to handle (with `-O1` optimisation and passes which are used for build-kernel) cases when the condition is saved to a variable, eg something like this:
    ```c
    // old
    bool a(int b, int c, int d) {
	int e=b<0;	
	if (e) {
		return false;
	}
	else {
		return true;
	}
    }
    // new
    bool a(int b, int c, int d) {
	int e=!(b<0);	
	if (e) {
		return true;
	}
	else {
		return false;
	}
    }
    ```
    The problem are `store` and `load` instructions.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -377,8 +377,10 @@ int DifferentialFunctionComparator::cmpOperations(
             // It is sufficient to compare the predicates here since the
             // operands are compared in cmpBasicBlocks.
             if (CmpL->getPredicate() == CmpR->getInversePredicate()) {
-                inverseConditions.emplace(L, R);
-                return 0;
+                if (checkInverseCondUsers(L) && checkInverseCondUsers(R)) {
+                    inverseConditions.emplace(L, R);
+                    return 0;
+                }
             }
         }
     }
@@ -940,7 +942,10 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
             // it will be used in an inverse condition. Hence, we skip it here
             // and mark that it may be inverse-matching the condition that
             // has been originally mapped to the operand of the not operation.
-            if (isLogicalNot(&*InstL) || isLogicalNot(&*InstR)) {
+            if (config.Patterns.InverseConditions
+                && ((isLogicalNot(&*InstL) && checkInverseCondUsers(&*InstL))
+                    || (isLogicalNot(&*InstR)
+                        && checkInverseCondUsers(&*InstR)))) {
                 sn_mapL.erase(&*InstL);
                 sn_mapR.erase(&*InstR);
 
@@ -1958,4 +1963,24 @@ bool DifferentialFunctionComparator::collectBinaryOperands(
         return true;
     }
     return false;
+}
+
+/// Recursively check if users of the instruction with an inverse condition
+/// are only branch or not instructions.
+/// This is done to prevent false negatives.
+/// Return true if the users are acceptable for inverse conditions pattern,
+/// false otherwise.
+bool DifferentialFunctionComparator::checkInverseCondUsers(
+        const Instruction *inst) const {
+    for (auto user : inst->users()) {
+        auto userInstruction = dyn_cast<Instruction>(user);
+        bool isNotInstruction = isLogicalNot(userInstruction);
+        if (!isNotInstruction && !isa<BranchInst>(userInstruction)) {
+            return false;
+        } else if (isNotInstruction
+                   && !checkInverseCondUsers(userInstruction)) {
+            return false;
+        }
+    }
+    return true;
 }
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -331,6 +331,8 @@ int DifferentialFunctionComparator::cmpOperations(
                                         BranchR->getCondition());
             if (inverseConditions.find(conds) != inverseConditions.end()) {
                 // Swap successors of one of the branches
+                PREP_LOG("swap-branches", L, R);
+                LOG_KEEP_FORCE(0);
                 BranchInst *BranchNew = const_cast<BranchInst *>(BranchR);
                 auto *tmpSucc = BranchNew->getSuccessor(0);
                 BranchNew->setSuccessor(0, BranchNew->getSuccessor(1));
@@ -378,6 +380,9 @@ int DifferentialFunctionComparator::cmpOperations(
             // operands are compared in cmpBasicBlocks.
             if (CmpL->getPredicate() == CmpR->getInversePredicate()) {
                 if (checkInverseCondUsers(L) && checkInverseCondUsers(R)) {
+                    PREP_LOG("inv-cond-add", L, R);
+                    LOG_KEEP_FORCE(0);
+
                     inverseConditions.emplace(L, R);
                     return 0;
                 }
@@ -986,6 +991,15 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
                 size_t erased = inverseConditions.erase(prevCondPair);
                 if (!erased)
                     inverseConditions.insert(condPair);
+
+                LOG_VERBOSE_EXTRA((erased ? "Removed" : " Added")
+                                  << " inverse condition\n");
+                auto &pairToLog = (erased) ? prevCondPair : condPair;
+                LOG_INDENT();
+                LOG_VERBOSE_EXTRA("L: " << (*(pairToLog.first)) << "\n");
+                LOG_VERBOSE_EXTRA("R: " << (*(pairToLog.second)) << "\n");
+                LOG_UNINDENT();
+
                 continue;
             }
 
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -941,32 +941,51 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
             // If one of the instructions is a logical not, it is possible that
             // it will be used in an inverse condition. Hence, we skip it here
             // and mark that it may be inverse-matching the condition that
-            // has been originally mapped to the operand of the not operation.
+            // has been previously mapped to the operand of the not operation.
             if (config.Patterns.InverseConditions
                 && ((isLogicalNot(&*InstL) && checkInverseCondUsers(&*InstL))
                     || (isLogicalNot(&*InstR)
                         && checkInverseCondUsers(&*InstR)))) {
-                sn_mapL.erase(&*InstL);
-                sn_mapR.erase(&*InstR);
+                // Note: This does not handle cases with a lot of `not`:
+                // %1 = cmp true, false | %1 = cmp true, false
+                //                      | %2 = not %1 (%2 = xor %1, true)
+                //                      | %3 = not %2 (%3 = xor %2, true)
+                //                      | %4 = not %3 (%4 = xor %3, true)
+                // br %1, l %T, l %F    | br %4, l %F, l %T
 
-                std::pair<const Value *, const Value *> matchingPair;
+                undoLastInstCompare(InstL, InstR);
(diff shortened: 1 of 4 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -940,26 +946,46 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
...
+            if (config.Patterns.InverseConditions
+                && (isLogicalNot(&*InstL) || isLogicalNot(&*InstR))) {
+                // Note: This does not handle cases with a lot of `not`:
+                // %1 = cmp true, false | %1 = cmp true, false
+                //                      | %2 = not %1 (%2 = xor %1, true)
+                //                      | %3 = not %2 (%3 = xor %2, true)
+                //                      | %4 = not %3 (%4 = xor %3, true)
+                // br %1, l %T, l %F    | br %4, l %F, l %T
                 sn_mapL.erase(&*InstL);
                 sn_mapR.erase(&*InstR);
```
PLukas2018, line 958: Also I am not sure, if this should not erase only the `not` instruction or maybe even the original `cmp` instruction?

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1959,3 +1985,21 @@ bool DifferentialFunctionComparator::collectBinaryOperands(
...
+/// false otherwise.
+bool DifferentialFunctionComparator::checkInverseCondUsers(
+        const Instruction *inst) const {
+    for (auto user : inst->users()) {
+        auto userInstruction = dyn_cast<Instruction>(user);
+        if (isa<BranchInst>(userInstruction) || isLogicalNot(userInstruction)) {
+            checkInverseCondUsers(userInstruction);
+        } else {
+            return false;
+        }
```
viktormalik, line 2002: This doesn't seem correct, you're not using the result of the recursive call. Also no need for the recursive call in case of `BranchInst`.
Perhaps should be like this?
```suggestion
        if (isLogicalNot(userInstruction) && !checkInverseCondUsers(userInstruction)) {
            return false;
        } else if (!isa<BranchInst>(userInstruction)) {
            return false;
        }
```
We could possibly join the conditions together, but it may be less readable.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -940,26 +946,46 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
...
                 // If the conditions are already inverse, remove them from the
                 // list. Otherwise, add them.
-                size_t erased = inverseConditions.erase(matchingPair);
+                size_t erased = inverseConditions.erase(oldMatchingPair);
+                LOG_VERBOSE_EXTRA((erased ? "Removed" : " Added")
+                                  << " inverse condition\n");
+                LOG_INDENT();
+                LOG_VERBOSE_EXTRA("L: " << (*(matchingPair.first)) << "\n");
+                LOG_VERBOSE_EXTRA("R: " << (*(matchingPair.second)) << "\n");
+                LOG_UNINDENT();
```
viktormalik, line 988: This visually decouples the two related operations of updating `inverseConditions`. Better move it below the `insert` line?

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -940,26 +946,46 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
...
+                // %1 = cmp true, false | %1 = cmp true, false
+                //                      | %2 = not %1 (%2 = xor %1, true)
+                //                      | %3 = not %2 (%3 = xor %2, true)
+                //                      | %4 = not %3 (%4 = xor %3, true)
+                // br %1, l %T, l %F    | br %4, l %F, l %T
                 sn_mapL.erase(&*InstL);
                 sn_mapR.erase(&*InstR);
 
                 std::pair<const Value *, const Value *> matchingPair;
+                std::pair<const Value *, const Value *> oldMatchingPair;
```
viktormalik, line 961: Nice catch! I completely missed this one.

I wonder if we could come up with better names.
Suggestion:
```suggestion
                std::pair<const Value *, const Value *> condPair;
                std::pair<const Value *, const Value *> origCondPair;
```
Maybe even a comment explaining the purpose of each of these would be useful.

viktormalik (changes requested):
These are great bug discoveries and fixes, many thanks!

One note on the commit order: the golden rule should be that all tests always pass after each commit (helps when hunting bugs down the history). IIUC, that's not the case here as the first commit adds a test which is only fixed by one of the later commits. If I'm right, could you please reorder the commits? Thanks!

Other than that, I've just some minor comments.

diffkemp/simpll/DifferentialFunctionComparator.cpp
PLukas2018: Thanks for the explanation, it makes sense. Now I am thinking if it should not also be erased from `mappedValuesBySn`, but I think it is not necessary, because the next instructions will overwrite it .

diffkemp/simpll/DifferentialFunctionComparator.cpp
PLukas2018: Thanks, I somehow did not realise or overlooked that...
I used your suggestion but slightly modified it (there was a problem when it was `not` instruction).

diffkemp/simpll/DifferentialFunctionComparator.cpp
PLukas2018: Thanks, I edited it and I think it is better.

diffkemp/simpll/DifferentialFunctionComparator.cpp
PLukas2018: Thanks, I added the comment and used your suggestion, but I used `prevCondPair` instead of `origCondPair`, because if there are 2 or more `not` instructions the old pair will not be the original condition but it would look like something like this `{not ..., nullptr}` because it is not currently able to find the original instruction. 

I am thinking of fixing it but when I tried to compile some programs with negation (`!`), after compilation there were none `not` (`xor`) instructions. Also currently it would not be able to handle if `not` instructions would be on both sides. But I am not sure if it makes sense to fix this.

PLukas2018:
> One note on the commit order: the golden rule should be that all tests always pass after each commit (helps when hunting bugs down the history). IIUC, that's not the case here as the first commit adds a test which is only fixed by one of the later commits. If I'm right, could you please reorder the commits? Thanks!

Thanks, did not know that.

I addressed the comments. Fixed few other things I overlooked:
- forgot to check users in `not` instruction,
- wrong logging in `not` instruction (did log current pair even when the old was removed),
- add an update of number of compared instructions.

diffkemp/simpll/DifferentialFunctionComparator.cpp
viktormalik: That's correct but I think that we should erase it anyways - just to be sure that there is nothing left from the (incorrect) comparison of the last 2 instructions. Ideally, we'd use `undoLastInstCompare` but we can't because we don't want to control the `InstL` and `InstR` iterators in a slightly different way.

Good catch btw! Can you please add the `mappedValuesBySn` erase?

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -939,29 +946,64 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
...
                     replacedInstructions.emplace(&*InstR, InstR->getOperand(0));
                     InstR++;
+                    // left instruction will be compared again
+                    ComparedInstL--;
                 }
 
-                // If the conditions are already inverse, remove them from the
-                // list. Otherwise, add them.
-                size_t erased = inverseConditions.erase(matchingPair);
+                // If the previous pair was are already inverse,
```
viktormalik, line 992:
```suggestion
                // If the previous pair already had inverse instructions,
```

diffkemp/simpll/DifferentialFunctionComparator.cpp
viktormalik: Yeah, LLVM does a good job in removing successive negations. You could try to use `-O0` to see if it helps but we still apply some transformations in SimpLL which may remove those.

Anyways, since this is so hard to produce in the IR, it practically shouldn't occur in real code and therefore I don't think it's worth fixing.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -939,29 +946,64 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
...
+                // remove it from the set. Otherwise, add the current pair.
+                size_t erased = inverseConditions.erase(prevCondPair);
                 if (!erased)
-                    inverseConditions.insert(matchingPair);
+                    inverseConditions.insert(condPair);
+
+                LOG_VERBOSE_EXTRA((erased ? "Removed" : " Added")
+                                  << " inverse condition\n");
+                std::pair<const Value *, const Value *> &pairToLog =
+                        (erased) ? prevCondPair : condPair;
```
viktormalik, line 1001: Would this work?
```suggestion
                auto &pairToLog = (erased) ? prevCondPair : condPair;
```

diffkemp/simpll/DifferentialFunctionComparator.cpp
viktormalik: Good point, I see it now.

viktormalik (commented):
LGTM, just a couple of small nits. Also please add the `undoLastInstCompare` call.

diffkemp/simpll/DifferentialFunctionComparator.cpp
PLukas2018: Thanks, good idea.

diffkemp/simpll/DifferentialFunctionComparator.cpp
PLukas2018: I did not manage to get LLVM IR containing `not` (xor) instruction even with `-O0`, so I will leave it like this.

diffkemp/simpll/DifferentialFunctionComparator.cpp
PLukas2018: Thanks.

PLukas2018:
Thanks, I fixed the problems.

viktormalik (approved):

PR #307, opened 2023-11-11
Add a pattern which detects local variable refactorings

zacikpa:
The pattern works as follows:
- Differing allocas of local memory and stores to such memory are ignored.
- When there is a load from such memory, identifying the replacement works the same as with repetitive loads, i.e., a suitable store has to be found in all backward CFG paths from the load.

Similar to the repetitive loads, this approach is limited by possibly conflicting stores and function calls. Further improvements may be possible using pointer alias analysis (maybe using LLVM's `MemoryDependenceAnalysis` pass?).

In the process of implementing this, I found multiple bugs in the repetitive load skipping implementation; see the first commit.

I'm not really sure about the pattern name; maybe something like `local-memory` would be more suitable?

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -504,13 +504,16 @@ int DifferentialFunctionComparator::cmpAllocs(const CallInst *CL,
                     || cmpIntWithConstant(TypeInfoR.Size, CR->getOperand(0)));
 }
 
-/// Check if the given instruction can be ignored (it does not affect
-/// semantics). Replacements of ignorable instructions are stored
-/// inside the ignored instructions map.
+/// Check if the given instruction can be skipped because it does not
+/// affect semantics or can be handled later. If yes, the instruction
+/// is stored within the set/map of ignored, skipped or replaced
+/// instructions (depending on the context).
 bool DifferentialFunctionComparator::maySkipInstruction(
         const Instruction *Inst) const {
     if (isa<AllocaInst>(Inst)) {
-        ignoredInstructions.insert(Inst);
+        // An alloca can always be skipped without any replacement.
+        // It can later be synchronized if used as an operand.
+        skippedInstructions.insert(Inst);
         return true;
     }
     if (config.Patterns.ReorderedBinOps && isReorderableBinaryOp(Inst)
@@ -537,7 +540,7 @@ bool DifferentialFunctionComparator::maySkipInstruction(
 
 /// Check whether the given cast can be ignored (it does not affect
 /// semantics. First operands of ignorable casts are stored as their
-/// replacements inside the ignored instructions map.
+/// replacements inside the replaced instructions map.
 bool DifferentialFunctionComparator::maySkipCast(const User *Cast) const {
     auto SrcTy = Cast->getOperand(0)->getType();
     auto DestTy = Cast->getType();
@@ -596,9 +599,9 @@ bool DifferentialFunctionComparator::maySkipCast(const User *Cast) const {
     return false;
 }
 
-/// Check whether the given instruction is a repetitive variant of a previous
-/// load with no store instructions in between. Replacements of ignorable loads
-/// are stored inside the ignored instructions map.
+/// Check whether the given instruction is a repetitive variant of
+/// a previous load with no store instructions in between.
+/// Load replacements are stored inside the replaced instructions map.
 bool DifferentialFunctionComparator::maySkipLoad(const LoadInst *Load) const {
     auto BB = Load->getParent();
     if (!BB) {
@@ -689,7 +692,7 @@ bool mayIgnoreMacro(std::string macro) {
     return ignoredMacroList.find(macro) != ignoredMacroList.end();
 }
 
-/// Retrive the replacement for the given value from the ignored instructions
+/// Retrieve the replacement for the given value from the replaced instructions
 /// map. Try to generate the replacement if a bitcast is given.
 const Value *DifferentialFunctionComparator::getReplacementValue(
         const Value *Replaced, DenseMap<const Value *, int> &sn_map) const {
@@ -1440,11 +1443,35 @@ int DifferentialFunctionComparator::cmpValues(const Value *L,
                                               const Value *R) const {
     PREP_LOG("value", L, R);
 
-    // Instructions that were ignored without replacement must never be compared
-    // as values. Two such instructions would always synchronize (even if not
-    // equal). Comparing with an already synchronized value would pollute the
-    // SN maps and complicate eventual undo.
-    if (ignoredInstructions.count(L) || ignoredInstructions.count(R)) {
+    // If both values are reversibly skipped, we can try to compare them
+    // fully - as instructions. This can help match reordered allocas.
+    auto SkippedEnd = skippedInstructions.end();
+    auto SkippedItL = skippedInstructions.find(L);
+    auto SkippedItR = skippedInstructions.find(R);
+    if (SkippedItL != SkippedEnd && SkippedItR != SkippedEnd) {
+        auto InstL = dyn_cast<Instruction>(*SkippedItL);
+        auto InstR = dyn_cast<Instruction>(*SkippedItR);
+        // The instructions must not be marked as skipped now, because
+        // comparison of operations calls cmpValues immediately again.
+        skippedInstructions.erase(SkippedItL);
+        skippedInstructions.erase(SkippedItR);
+        if (!cmpOperationsWithOperands(InstL, InstR)) {
+            RETURN_WITH_LOG_NEQ(0);
+        }
+        // If the comparison failed, restore the ignored instructions map
+        // and synchronization maps, which are polluted.
+        skippedInstructions.insert(InstL);
+        skippedInstructions.insert(InstR);
+        sn_mapL.erase(InstL);
+        sn_mapR.erase(InstR);
+        mappedValuesBySn.erase(sn_mapL.size());
+        RETURN_WITH_LOG_NEQ(1);
+    }
+
+    // If either of the values is irreversibly ignored, or if only one of
+    // the instructions is skipped, fail.
+    if (ignoredInstructions.count(L) || ignoredInstructions.count(R)
+        || SkippedItL != SkippedEnd || SkippedItR != SkippedEnd) {
         RETURN_WITH_LOG_NEQ(1);
     }
 
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -14,6 +14,7 @@
 
 #include "DifferentialFunctionComparator.h"
 #include "Config.h"
+#include "DebugInfo.h"
 #include "FieldAccessUtils.h"
 #include "Logger.h"
 #include "SourceCodeUtils.h"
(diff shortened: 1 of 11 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -595,99 +616,114 @@ bool DifferentialFunctionComparator::maySkipCast(const User *Cast) const {
...
 bool DifferentialFunctionComparator::maySkipReorderableBinaryOp(
         const Instruction *Op) const {
-    return std::all_of(Op->user_begin(), Op->user_end(), [Op](auto user) {
-        auto userOp = dyn_cast<BinaryOperator>(user);
-        return userOp && userOp->getOpcode() == Op->getOpcode();
+    return std::all_of(Op->user_begin(), Op->user_end(), [&](auto user) {
+        if (auto userBinOp = dyn_cast<BinaryOperator>(user))
+            return userBinOp->getOpcode() == Op->getOpcode();
+        if (auto userStore = dyn_cast<StoreInst>(user))
+            return hasIgnoredAllocaAsPtrOperand(userStore);
```
viktormalik, line 717: Why do we need to have this check here? This pattern doesn't seem specific to binary operations, could we come up with some general approach that would allow us to avoid this?

Also this is missing the `config.Patterns.StructChanges` check.

viktormalik (commented):
I'm a bit confused about what situations this pattern tries to cover. The unit test features a replacement of multiple local constants (possibly local variables?) by a single local variable of a struct type. In literature, this is called *group a set of variables into a new structure*, in which case we could call it `group-vars` in DiffKemp.

On the other hand, the pattern seems to be more general than that because it doesn't store a mapping between the original values (constants) and the new memory locations (fields of a local struct) - or I've failed to find that. As I noted in one of the comments, this approach seems too general and prone to false positives.

Am I missing something?

Besides that, I've a couple of minor comments.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -504,10 +505,22 @@ int DifferentialFunctionComparator::cmpAllocs(const CallInst *CL,
...
-/// inside the ignored instructions map.
+/// semantics). Ignorable instructions are stored inside the replaced
+/// instructions map or the ignored instructions set, depending on whether
+/// they need a replacement.
 bool DifferentialFunctionComparator::maySkipInstruction(
         const Instruction *Inst) const {
+    if (Inst->isSafeToRemove()
+        && (Inst->getNumUses() == 0
+            || (config.Patterns.StructChanges
+                && isOnlyUsedByIgnorableMemInstructions(Inst)))) {
```
viktormalik, line 516:
```suggestion
    if (config.Patterns.StructChanges && Inst->isSafeToRemove() && isOnlyUsed...) {
```

diffkemp/simpll/DifferentialFunctionComparator.cpp
viktormalik: Idea: add a method which will follow the stored value and return uses of all loads from the same pointer. Then, check that all uses of these are binary operations.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1379,6 +1415,12 @@ int DifferentialFunctionComparator::cmpFieldAccess(
     if (!(GEPL && GEPR))
         RETURN_WITH_LOG(1);
 
+    // If we previously ignored allocation of memory that any of the two GEPs
+    // point to, fail (cmpValues would compare their pointer operands as equal).
+    if (hasIgnoredAllocaAsPtrOperand(GEPL)
+        || hasIgnoredAllocaAsPtrOperand(GEPR))
+        RETURN_WITH_LOG(1);
```
viktormalik, line 1422: What is the purpose of this? It can lead to false positives so we double-check why it is necessary.

viktormalik (changes requested):

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -504,10 +505,22 @@ int DifferentialFunctionComparator::cmpAllocs(const CallInst *CL,
 }
 
 /// Check if the given instruction can be ignored (it does not affect
-/// semantics). Replacements of ignorable instructions are stored
-/// inside the ignored instructions map.
+/// semantics). Ignorable instructions are stored inside the replaced
+/// instructions map or the ignored instructions set, depending on whether
+/// they need a replacement.
```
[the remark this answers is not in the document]
zacikpa: Updated.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -595,99 +616,114 @@ bool DifferentialFunctionComparator::maySkipCast(const User *Cast) const {
...
+                    && Store->getPointerOperand()
+                               == Load->getPointerOperand()) {
+                    // Found a previously ignored store, save the candidate.
+                    ReplacementCandidate = Store->getValueOperand();
+                }
+            }
+            if (ReplacementCandidate) {
+                if (Replacement) {
+                    // If another suitable replacement has been found previously
+                    // in a different CFG path, fail. We need a single one.
```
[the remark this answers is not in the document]
zacikpa: Yes, exactly.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -595,99 +616,114 @@ bool DifferentialFunctionComparator::maySkipCast(const User *Cast) const {
...
-        if (searchPredecessors) {
+        if (checkPredecessors) {
             auto BBPredecessors = predecessors(BB);
             if (BBPredecessors.begin() == BBPredecessors.end()) {
-                // If there are no more predecessors available,
-                // end the analysis with a failure.
-                PreviousLoad = nullptr;
-                break;
+                // Fail if no more predecessors are available.
+                return false;
```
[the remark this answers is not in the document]
zacikpa: Updated the comment to be more explicit about this, thanks.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -528,6 +541,14 @@ bool DifferentialFunctionComparator::maySkipInstruction(
         replacedInstructions.insert({Inst, Inst->getOperand(0)});
         return true;
     }
+    if (auto Store = dyn_cast<StoreInst>(Inst)) {
+        if (config.Patterns.StructChanges
+            && hasIgnoredAllocaAsPtrOperand(Store)) {
+            // Skip a store if it points to a previously ignored alloca.
```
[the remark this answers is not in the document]
zacikpa: As we discussed in person, this could not lead to false negatives due to the check at the beginning of `cmpValues` which disabled value comparison of skipped allocas.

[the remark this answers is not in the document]
zacikpa: In the updated version, where we _can_ "un-ignore" allocas, this is a problem which I fixed by converting the `ignoredInstruction` set to a map. The value in the map is a bool which specifies whether the ignored instruction can be "un-ignored" in the future. If we ever skip a store to the alloca before un-ignoring it, the value is set to false for the alloca, and from that point onward, it cannot be un-ignored.

diffkemp/simpll/Utils.cpp
```diff
@@ -874,3 +875,66 @@ StructType *getTypeByName(const Module &Mod, StringRef Name) {
     return Mod.getTypeByName(Name);
 #endif
 }
+
+/// Given an instruction and a pointer value, try to determine whether the
+/// instruction may store to the memory pointed to by the pointer. This is
+/// currently implemented only as a set of heuristics - if they do not suffice,
+/// we return true.
+bool mayStoreToMemoryAtPtr(const Instruction *Inst, const Value *Ptr) {
```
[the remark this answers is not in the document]
zacikpa: Fixed, together with other naming issues that we talked about.

diffkemp/simpll/DifferentialFunctionComparator.cpp
zacikpa: I didn't figure it out. It might have been relevant before I added some other check elsewhere, but everything seems to work fine without this now.

diffkemp/simpll/DifferentialFunctionComparator.cpp
zacikpa: I decided not to follow this approach because I think it's unnecessary. If the result of the skipped binary operation is stored to an ignored variable, we know that the store will always be skipped, which should never lead to false negatives.

I did, however, refactor `maySkipInstruction` to make this more general. See my comment summarizing the changes.

zacikpa:
Thanks for the thorough review. I updated the PR with the following changes:

- The pattern is now named `group-vars`.
- I refactored a couple of utility functions so as not to duplicate code.
- `ignoredInstructions` is now a map that stores the information on whether the ignored instruction can be "un-ignored" if it is successfully compared (fully) in `cmpValues`. This is currently only used for ignored `alloca` instructions.
- New logic added to the beginning of `cmpValues` related to the above: if we encounter two ignored instructions that can both be un-ignored, we attempt to compare them properly.
- When skipping a `store` to an ignored `alloca`, the `alloca` can't be un-ignored from that point onward.
- Generalized skipping instructions based on their users; in some cases, an instruction may be skipped because it only has certain kinds of users. The `maySkipInstruction` function now collects alternative conditions that the users may fulfill in order to allow instruction skipping.

viktormalik:
One high-level comment (I'm still trying to figure out the code):

> Thanks for the thorough review. I updated the PR with the following changes:
> 
> 
> 
> - The pattern is now named `group-vars`.
> 
> - I refactored a couple of utility functions so as not to duplicate code.
> 
> - `ignoredInstructions` is now a map that stores the information on whether the ignored instruction can be "un-ignored" if it is successfully compared (fully) in `cmpValues`. This is currently only used for ignored `alloca` instructions.
> - New logic added to the beginning of `cmpValues` related to the above: if we encounter two ignored instructions that can both be un-ignored, we attempt to compare them properly.

Idea: could we separate "ignored instructions" which are never compared again and "skipped instructions" which are temporarily skipped but compared again once encountered as operands?

> - When skipping a `store` to an ignored `alloca`, the `alloca` can't be un-ignored from that point onward.
> 
> - Generalized skipping instructions based on their users; in some cases, an instruction may be skipped because it only has certain kinds of users. The `maySkipInstruction` function now collects alternative conditions that the users may fulfill in order to allow instruction skipping.

zacikpa:
I updated the PR again.

> Idea: could we separate "ignored instructions" which are never compared again and "skipped instructions" which are temporarily skipped but compared again once encountered as operands?

Good idea. It requires updating a couple of checks (newly checking both `ignored` and `skipped` instructions), but I believe it is still worth it. Done.

Let's not support obscure combinations of patterns for now. I reverted all changes to the `reordered-bin-ops` pattern, which left three major changes in this PR. I force-pushed them as new commits (sorry for that;  the changes were starting to get really messy).

@viktormalik 
All the comments from the first review should be resolved.

zacikpa:
New edit: Swapped the meaning of skipped and ignored instructions (skipped instructions can be compared again as operands; ignoring is irreversible).

diffkemp/simpll/Utils.cpp
```diff
@@ -874,3 +875,66 @@ StructType *getTypeByName(const Module &Mod, StringRef Name) {
     return Mod.getTypeByName(Name);
 #endif
 }
+
+/// Given an instruction and a pointer value, try to determine whether the
+/// instruction may store to the memory pointed to by the pointer. This is
+/// currently implemented only as a set of heuristics - if they do not suffice,
+/// we return true.
```
viktormalik, line 882: Yet there's `return false` at the end of the function :-)

viktormalik (commented):
Overall looks good to me, just some remarks and details for comments and naming.

diffkemp/simpll/Utils.cpp
zacikpa: :) Reworded.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -2025,3 +2111,28 @@ bool DifferentialFunctionComparator::checkInverseCondUsers(
     }
     return true;
 }
+
+/// Check whether the pointer operand of the given instruction (e.g., load)
+/// is a previously ignored alloca or a GEP pointing to one.
```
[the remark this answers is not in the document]
zacikpa: Reworded to "local memory allocated by an ignored alloca". The `getAllocaOp` retrieves the alloca through a GEP if necessary.

viktormalik (approved):
Thanks for all the work put into this pattern!

viktormalik:
Can you squash the last commit?

PR #309, opened 2023-11-24
Various bug fixes

zacikpa:
This PR fixes 3 bugs I encountered when analyzing cryptographic libraries. For details, see commit messages.

### DIFF
--- a/diffkemp/output.py
+++ b/diffkemp/output.py
@@ -79,6 +79,8 @@ class YamlOutput:
         self._create_type_defs()
 
     def _create_function_defs(self):
+        if self.result.graph is None:
+            return
         definitions = {}
         for name in self.function_names:
             if name not in self.result.graph.vertices:
@@ -102,11 +104,12 @@ class YamlOutput:
         self.output["definitions"].update(definitions)
 
     def _create_type_defs(self):
-        definitions = {}
         """Returns definitions of types.
         :param type_names: Set of tuples (type name, parent name),
                            parent name - name of function in which type is used
         """
+        if self.result.graph is None:
+            return
         definitions = {}
         for type_name, parent_name in self.type_names:
             if parent_name not in self.result.graph.vertices:
--- a/diffkemp/simpll/passes/CalledFunctionsAnalysis.h
+++ b/diffkemp/simpll/passes/CalledFunctionsAnalysis.h
@@ -43,6 +43,9 @@ class CalledFunctionsAnalysis
   private:
     friend AnalysisInfoMixin<CalledFunctionsAnalysis>;
     static AnalysisKey Key;
+    /// The set of values that were already processed in the current run.
+    /// Prevents infinite recursion when processing instruction operands.
+    std::set<const Value *> ProcessedValues;
 };
 
 #endif // DIFFKEMP_SIMPLL_CALLEDFUNCTIONSANALYSIS_H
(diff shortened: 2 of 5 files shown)


diffkemp/output.py
```diff
@@ -81,7 +81,7 @@ def _create_definitions(self):
     def _create_function_defs(self):
         definitions = {}
         for name in self.function_names:
-            if name not in self.result.graph.vertices:
+            if not self.result.graph or name not in self.result.graph.vertices:
```
viktormalik, line 84: Move the new check out of the loop.

diffkemp/output.py
```diff
@@ -109,7 +109,8 @@ def _create_type_defs(self):
         """
         definitions = {}
         for type_name, parent_name in self.type_names:
-            if parent_name not in self.result.graph.vertices:
+            if not self.result.graph or \
```
viktormalik, line 112: Same as above.

diffkemp/simpll/passes/CalledFunctionsAnalysis.h
```diff
@@ -38,7 +38,9 @@ class CalledFunctionsAnalysis
     void collectCalled(const Function *Fun, Result &Called);
     /// Looks for functions in a value (either a function itself, or a composite
     /// type constant).
-    void processValue(const Value *Val, Result &Called);
+    void processValue(const Value *Val,
+                      Result &Called,
+                      std::set<const Value *> &Processed);
```
viktormalik, line 43: Could we make `Processed` a private class member?

viktormalik (changes requested):

PLukas2018:
Thanks for the check if `graph` is not `None`, I did not realise it could happen. I think the better place for the check would be before creating the definitions,
https://github.com/viktormalik/diffkemp/blob/f1e801dabbe9c9c35052c7900f1b9725bb230519/diffkemp/output.py#L28-L34
because both of the methods uses the `graph`.

viktormalik:
> Thanks for the check if `graph` is not `None`, I did not realise it could happen. I think the better place for the check would be before creating the definitions,
> 
> https://github.com/viktormalik/diffkemp/blob/f1e801dabbe9c9c35052c7900f1b9725bb230519/diffkemp/output.py#L28-L34
> 
> 
> because both of the methods uses the `graph`.

I'd argue that it's up to the methods to check that everything they need exists. If we added the check to the caller method, we'd have to add checks for graph existence for all future calls of `_create_results` and `_create_definitions`. I'm aware that these are class-internal methods which are very unlikely to be called from another place, but it's safer (although slightly more verbose) this way.

PLukas2018:
> I'd argue that it's up to the methods to check that everything they need exists. If we added the check to the caller method, we'd have to add checks for graph existence for all future calls of `_create_results` and `_create_definitions`. I'm aware that these are class-internal methods which are very unlikely to be called from another place, but it's safer (although slightly more verbose) this way.

Thanks, that makes sense.

diffkemp/output.py
zacikpa: Done.

diffkemp/output.py
zacikpa: Done.

diffkemp/simpll/passes/CalledFunctionsAnalysis.h
zacikpa: We certainly can. That's a much cleaner solution, thank you.

viktormalik (approved):
LGTM, thanks!

The CI failure doesn't seem to come from this PR, I deleted the CI caches and restarted it.

Can you squash the commits, please? 2 should be sufficient (1 for YAML fix and 1 for recursion fix).

zacikpa:
Done, thanks for the review.

PR #310, opened 2023-11-25
Enable linking missing source modules for wrapper-built snapshots

zacikpa:
This was possible only for kernel snapshots until now.

It is quite a dirty fix, however. It should probably be updated in the future
in such a way that we copy all modules to the snapshot right away.

### DIFF
--- a/diffkemp/snapshot.py
+++ b/diffkemp/snapshot.py
@@ -8,6 +8,7 @@ from diffkemp.llvm_ir.llvm_module import LlvmModule
 from diffkemp.llvm_ir.kernel_source_tree import KernelSourceTree
 from diffkemp.llvm_ir.source_tree import SourceTree
 from diffkemp.llvm_ir.kernel_llvm_source_builder import KernelLlvmSourceBuilder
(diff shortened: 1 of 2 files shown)


diffkemp/snapshot.py
```diff
@@ -198,6 +199,11 @@ def _from_yaml(self, yaml_file):
             source_dir = self.snapshot_tree.source_dir
             source_finder = KernelLlvmSourceBuilder(source_dir)
             self.snapshot_tree = KernelSourceTree(source_dir, source_finder)
+        elif yaml_dict["llvm_source_finder"]["kind"] == "wrapper_build_tree":
```
viktormalik, line 202: Update the comment above, please.

viktormalik (approved):
> It is quite a dirty fix, however. It should probably be updated in the future
in such a way that we copy all modules to the snapshot right away.

Yeah, agreed. It was originally created for `build-kernel` because the necessary module may not have been built at the time of requesting it. That shouldn't happen with `build` where we always build all sources so copying everything sounds like a better option (although it'd make the snapshot unnecessarily large).

Still LGTM as a temporary fix, just a nit.

Issue #287, opened 2023-09-29
Cleanup LLVM passes for snapshot building


viktormalik:
We now have two commands for creating snapshots that use Clang to build the analysed project: `build` and `build-kernel`. Each of these uses a different approach to run LLVM passes:
- `build` respects whatever is the default for the analysed project. This may result in insufficient passes being applied (e.g. if `-O0` is used).
- `build-kernel` uses `-O1 -Xclang -disable-llvm-passes` and then applies custom passes by running opt (see [kernel_llvm_source_builder.py:412](https://github.com/viktormalik/diffkemp/blob/master/diffkemp/llvm_ir/kernel_llvm_source_builder.py#L412)). It originally used `-O0` but 91f72dd changed that b/c of LLVM 5.0. We don't support LLVM 5.0 anymore so that can be probably dropped.

We should unify both (and any future) commands to use a common approach. The `build-kernel` one allows more control over what's being run, the question is whether it gives any advantage over simple `-O2`. Some experiments on that would be useful.

PLukas2018:
It looks like it will probably not be possible to use `-O2`. 

I [changed `build-kernel`](https://github.com/PLukas2018/diffkemp/commit/b0faed3915533572c4b992505085adcba56d639b) to use `-O2` and to not use custom passes (run `opt`). For the regression (pytest) tests, 28 tests failed (29 for LLVM 12 and below). It would probably be necessary to go through the tests more thoroughly to be certain that the reason is not different.

In any case, by using `-O2` it would probably be harder to identify exactly which function caused the semantic difference because the compiler can inline the functions (as I saw on EqBench benchmarks analysis). This could maybe cause better results but for the cost of not exactly knowing which function caused semantic difference -- I am not sure if the inlining is the reason for better results it would be necessary to analyse what is the cause of better results with `-O2`  (which is the case in EqBench benchmarks).

viktormalik:
Thanks for the analysis! It's possible that it'd just require to add one or two specific passes for the tests to pass. It may also be the case that `-O2` is too aggressive.

Also, you're probably right with the inlining. Since we intend to identify the function with differences as precisely as possible, we want to have inlining under control (which we probably wouldn't have with `-O2`).

[...]

PR #311, opened 2023-12-11
Snapshot building unification

PLukas2018:
DiffKemp contains multiple commands for creating snapshots (`build`, `build-kernel`, `llvm-to-snapshot`), this PR tries to unify  `build` command with `build-kernel` command (as discussed in #287 ). The clang options and LLVM passes are taken from `build-kernel` command, which is taken as the expected default behaviour for `build`.

The `build` command will now by default:
- compile file/project using `-O1 -Xclang -disable-llvm-passes`,
- run selected simplification passes using `opt`.

Using `--clang-drop` does not change the behaviour mentioned above only drops certain flags which are in the Makefile.

Same as before this PR, `--clang_append` can be used to specify other flags which should be added. You can also use it to change the optimization level, but it will still use `-Xclang -disable-llvm-passes` and run the default simplification passes.

By using `--no-opt-override` you can turn off the default behaviour, then the `-O1 -Xclang -disable-llvm-passes` flags are not used and the default simplification passes are not run. For the optimisation level is then used the one which is specified in Makefile or you can change it by `--clang-append`.

### DIFF
--- a/diffkemp/building/CMakeLists.txt
+++ b/diffkemp/building/CMakeLists.txt
@@ -20,9 +20,13 @@ if (NOT DEFINED ENV{WITHOUT_RPYTHON})
   else ()
     set(CC_WRAPPER "${CMAKE_BINARY_DIR}/cc_wrapper-c")
     add_custom_target(cc-wrapper ALL DEPENDS ${CC_WRAPPER})
+    # Note: Setting PYTHONPATH to be able to use diffkemp modules
+    # and also external modules (in case we decide to use them) from the wrapper.
+    set(PYTHONPATH "$ENV{PYTHONPATH}")
     add_custom_command(OUTPUT ${CC_WRAPPER}
                        DEPENDS ${CC_WRAPPER_SOURCE}
-                       COMMAND ${RPYTHON} ${CC_WRAPPER_SOURCE} "--output=${CC_WRAPPER}"
+                       COMMAND "PYTHONPATH=${CMAKE_SOURCE_DIR}:${PYTHONPATH}"
+                         ${RPYTHON} ${CC_WRAPPER_SOURCE} "--output=${CC_WRAPPER}"
                        WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
     install(PROGRAMS ${CC_WRAPPER} RENAME diffkemp-cc-wrapper DESTINATION ${CMAKE_INSTALL_BINDIR})
   endif ()
--- a/diffkemp/diffkemp.py
+++ b/diffkemp/diffkemp.py
@@ -1,6 +1,7 @@
 from diffkemp.building.cc_wrapper import get_cc_wrapper_path, wrapper_env_vars
 from diffkemp.config import Config
 from diffkemp.snapshot import Snapshot
+from diffkemp.llvm_ir.optimiser import opt_llvm, BuildException
 from diffkemp.llvm_ir.kernel_source_tree import KernelSourceTree
 from diffkemp.llvm_ir.kernel_llvm_source_builder import KernelLlvmSourceBuilder
 from diffkemp.llvm_ir.source_tree import SourceTree, SourceNotFoundException
@@ -58,6 +59,8 @@ def build_c_project(args):
         if args.clang_drop is not None
         else "",
         wrapper_env_vars["llvm_link"]: args.llvm_link,
+        wrapper_env_vars["no_opt_override"]: ("1" if args.no_opt_override
+                                              else "0")
     }
     environment.update(os.environ)
 
@@ -113,6 +116,19 @@ def build_c_project(args):
     # Build the project using generated wrapper
     check_call(make_target_args, env=environment)
 
+    # Run LLVM IR simplification passes if the user did not request
+    # to use the default project's optimization.
+    if not args.no_opt_override:
+        # Run llvm passes on created LLVM IR files.
+        with open(db_filename, "r") as db_file:
+            for line in [r for r in db_file if r.startswith("o:")]:
+                llvm_file = line.split(":")[1].rstrip()
+                try:
+                    opt_llvm(llvm_file)
+                except BuildException:
+                    # Unsuccessful optimization, leaving as it is.
+                    pass
+
     # Create a new snapshot from the source directory.
     source_finder = WrapperBuildFinder(args.source_dir, db_filename)
     source = SourceTree(args.source_dir, source_finder)
@@ -145,7 +161,8 @@ def build_c_file(args):
     # Create a new snapshot and generate its content.
     source_finder = SingleCBuilder(source_dir, source_file_name,
                                    clang=args.clang,
-                                   clang_append=clang_append)
+                                   clang_append=clang_append,
+                                   default_optim=not args.no_opt_override)
     source = SourceTree(source_dir, source_finder)
     snapshot = Snapshot.create_from_source(source, args.output_dir, "function")
     if args.symbol_list is None:
--- a/diffkemp/llvm_ir/compiler.py
+++ b/diffkemp/llvm_ir/compiler.py
@@ -1,10 +1,14 @@
 """
 Functions for compilation of c files to LLVM IR.
+
+This file must be RPython compatible because it is used from cc_wrapper.
 """
 
 
-def get_clang_default_options():
-    """Returns clang options for compiling c files to LLVM IR."""
-    return ["-S", "-emit-llvm", "-O1", "-Xclang",
-            "-disable-llvm-passes", "-g", "-fdebug-macro",
-            "-Wno-format-security"]
+def get_clang_default_options(default_optim=True):
+    """Returns clang options for compiling c files to LLVM IR.
+    :param default_optim: By default adds also optimization flags."""
+    opts = ["-S", "-emit-llvm", "-g", "-fdebug-macro", "-Wno-format-security"]
+    if default_optim:
+        opts.extend(["-O1", "-Xclang", "-disable-llvm-passes"])
+    return opts
--- a/diffkemp/llvm_ir/compiler.py
+++ b/diffkemp/llvm_ir/compiler.py
@@ -1,0 +1,10 @@
+"""
+Functions for compilation of c files to LLVM IR.
+"""
+
+
+def get_clang_default_options():
+    """Returns clang options for compiling c files to LLVM IR."""
+    return ["-S", "-emit-llvm", "-O1", "-Xclang",
+            "-disable-llvm-passes", "-g", "-fdebug-macro",
+            "-Wno-format-security"]
--- a/diffkemp/llvm_ir/single_c_builder.py
+++ b/diffkemp/llvm_ir/single_c_builder.py
@@ -1,6 +1,8 @@
 """
 LLVM source builder for single C file.
 """
+from diffkemp.llvm_ir.compiler import get_clang_default_options
+from diffkemp.llvm_ir.optimiser import opt_llvm
 from diffkemp.llvm_ir.single_llvm_finder import SingleLlvmFinder
 from diffkemp.utils import get_functions_from_llvm
 import os
@@ -13,10 +15,12 @@ class SingleCBuilder(SingleLlvmFinder):
     Extends the SingleLlvmFinder class by compiling C file to LLVM IR file.
     """
     def __init__(self, source_dir, c_file_name, clang="clang",
-                 clang_append=[]):
+                 clang_append=[], default_optim=True):
         """
         :param clang: clang compiler to be used
         :param clang_append: list of args to add when compiling
+        :param default_optim: use default optimalisations flags
+            and run LLVM IR simplification passes
         """
         llvm_file_name = os.path.splitext(c_file_name)[0] + ".ll"
         SingleLlvmFinder.__init__(self, source_dir, llvm_file_name)
@@ -24,17 +28,26 @@ class SingleCBuilder(SingleLlvmFinder):
         self.c_file_name = c_file_name
         self.clang = clang
         self.clang_append = clang_append
+        self.default_optim = default_optim
         self.initialize()
 
     def str(self):
         return "single_c_file"
 
     def initialize(self):
+        """Compiles the C file to LLVM IR file and runs passes"""
         # Using clang to compile c file to llvm file.
-        command = [self.clang, "-S", "-emit-llvm", "-g", self.c_file_name,
-                   "-o", self.llvm_file_name]
+        command = [self.clang, self.c_file_name, "-o", self.llvm_file_name]
+        # Note: clang uses the last specified optimization level so
+        # extending with the default options must be done before
+        # extending with the clang_append option.
+        command.extend(get_clang_default_options(self.default_optim))
         command.extend(self.clang_append)
         check_call(command, cwd=self.source_dir)
+
+        # Running llvm passes
+        if self.default_optim:
+            opt_llvm(self.llvm_file_path)
 
     def get_function_list(self):
         """Returns a list of functions that are found in the source file."""
--- a/flake.nix
+++ b/flake.nix
@@ -120,6 +120,13 @@
             # the function (and all functions it uses) as commands.
             shellHook = ''
               export -f setuptoolsShellHook runHook _eval _callImplicitHook
+
+              # Adding current (diffkemp) directory to PYTHONPATH,
+              # the `diffkemp build` subcommand does not work without it
+              # - `cc_wrapper.py` ends with `ModuleNotFoundError` because
+              # `setuptoolsShellHook` does not make diffkemp package
+              # importable for subprocesses called from python.
+              export PYTHONPATH="$(pwd):$PYTHONPATH"
             '';
           };
     in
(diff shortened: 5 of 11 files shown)


diffkemp/llvm_ir/compiler.py
```diff
@@ -0,0 +1,46 @@
+"""
+Functions for compilation of c files to LLVM IR, optimizations of LLVM IR.
+
+Syntax of this module must be python2 compatible because it is checked
+when building diffkemp.bulding.cc_wrapper with RPython.
+"""
+from diffkemp.utils import get_opt_command
```
viktormalik, line 7: Could we move `get_opt_command` to `compiler.py`? I'd like to minimize the amount of modules which must be Python2-compatible.

diffkemp/diffkemp.py
```diff
@@ -58,6 +59,8 @@ def build_c_project(args):
         if args.clang_drop is not None
         else "",
         wrapper_env_vars["llvm_link"]: args.llvm_link,
+        wrapper_env_vars["no_opt_override"]: ("1" if args.no_opt_override
+                                              else "0")
```
viktormalik, line 63: Isn't it possible to pass a boolean to the env variable? So that we could do
```suggestion
        wrapper_env_vars["no_opt_override"]: not args.no_opt_override
```

diffkemp/diffkemp.py
```diff
@@ -113,6 +116,16 @@ def build_c_project(args):
     # Build the project using generated wrapper
     check_call(make_target_args, env=environment)
 
+    # Run LLVM IR simplification passes if the user did not specified
+    # to not override project optimization.
+    if not args.no_opt_override:
+        # Run llvm passes on created LLVM IR files.
+        with open(db_filename, "r") as db_file:
+            for line in db_file:
+                if line.startswith("o:"):
```
viktormalik, line 125: Nit: this saves one level of indentation:
```suggestion
            for line in [l for l in db_file if l.startswith("o:")]:
```
You could even meld the next line inside but that would be probably less readable due to line breaks:
```suggestion
            for line in [l.split(":")[1].rstrip() for l in db_file if l.startswith("o:")]:
```

viktormalik (commented):
Nice! Just a couple of comments.

I also didn't run the code so I'd like to do that before merging.

Note: I don't think that we have tests that try to run `diffkemp build`, perhaps something worth adding to be able to test PRs like this one.

diffkemp/llvm_ir/compiler.py
PLukas2018: - I could move `get_opt_command` to `compiler.py`, but it would be necessary to also move `get_llvm_version` to `compiler.py` because it is called from `get_opt_command`, then the `utils.py` syntax would not have to be python2 compatible.
- The second option which comes to my mind, is to move the `opt_llvm` function from `compiler.py` to a different file (eg. `diffkemp/llvm_ir/optimiser.py`), because it is not used from `cc_wrapper` (by `rpython`), only the `get_clang_default_options` is used from `cc_wrapper`, but rpython requires all functions in the file and called from them to have python2 syntax. This would minimise the functions which would have to have python2 syntax, but there would be two new modules/files (`compiler.py` and eg. `optimiser.py`) which would both contain only one function for now.

Let me know, which option would you prefer.

diffkemp/diffkemp.py
PLukas2018: It is possible to pass a boolean, but I would still need to do something like this in `cc_wrapper` when loading the value:
```python3
if os.environ.get(wrapper_env_vars["no_opt_override"]) == "True":
        no_opt_override = True
    else:
        no_opt_override = False
```

because passing `False` will pass "False" and `bool("False") == True`

PLukas2018:
> Note: I don't think that we have tests that try to run `diffkemp build`, perhaps something worth adding to be able to test PRs like this one.

Yes, I think that we do not have any tests for `build` command. I think that in one PR, I tried to write some tests for the command, I tested creating a snapshot from of a file/parts of kernel, but that did not get merged because I think we want to get rid of depending on kernel sources when testing...
So, I do not really know what the tests should look like, if the repo should contain a small make-based project (eg. Makefile + one c file) which would be used for testing of the `build` command?

diffkemp/diffkemp.py
viktormalik: I see. Ok, let's stick with `1` and `0`.

FrNecas:
> I also didn't run the code so I'd like to do that before merging.

I've been using this PR for building snapshots for experimenting with SMT solving and so far it works nicely and I haven't encountered any issues as far as the functionality goes.

diffkemp/llvm_ir/compiler.py
viktormalik: Option 1 doesn't make much sense to me as `get_llvm_version` is used from other modules, too.

So, I'd either go with option 2 or with moving `get_opt_command` to `compiler.py` and moving `get_llvm_version` to a separate file (`llvm-version.py`?). I don't really have a strong preference here.

viktormalik:
> > Note: I don't think that we have tests that try to run `diffkemp build`, perhaps something worth adding to be able to test PRs like this one.
> 
> Yes, I think that we do not have any tests for `build` command. I think that in one PR, I tried to write some tests for the command, I tested creating a snapshot from of a file/parts of kernel, but that did not get merged because I think we want to get rid of depending on kernel sources when testing... So, I do not really know what the tests should look like, if the repo should contain a small make-based project (eg. Makefile + one c file) which would be used for testing of the `build` command?

Yeah, I imagined something of that sort - our own toy Makefile-based project for testing (in future, we could add CMake, Meson, etc.).

viktormalik:
> > I also didn't run the code so I'd like to do that before merging.
> 
> I've been using this PR for building snapshots for experimenting with SMT solving and so far it works nicely and I haven't encountered any issues as far as the functionality goes.

Thanks, this is very useful!

diffkemp/llvm_ir/compiler.py
PLukas2018: I used option 2 (moved the `opt_llvm` to `optimiser.py` ), problem with the third option was that the `get_opt_command` is also used from multiple modules.

diffkemp/diffkemp.py
PLukas2018: I used the first suggestion, the second was too long for one line.

PLukas2018:
> > > Note: I don't think that we have tests that try to run `diffkemp build`, perhaps something worth adding to be able to test PRs like this one.
> > 
> > 
> > Yes, I think that we do not have any tests for `build` command. I think that in one PR, I tried to write some tests for the command, I tested creating a snapshot from of a file/parts of kernel, but that did not get merged because I think we want to get rid of depending on kernel sources when testing... So, I do not really know what the tests should look like, if the repo should contain a small make-based project (eg. Makefile + one c file) which would be used for testing of the `build` command?
> 
> Yeah, I imagined something of that sort - our own toy Makefile-based project for testing (in future, we could add CMake, Meson, etc.).

I added the our toy Makefile-based project and tests, also I added testing of compilation of the `cc-wrapper` with RPython to CI.

viktormalik:
> I added the our toy Makefile-based project and tests, also I added testing of compilation of the `cc-wrapper` with RPython to CI.

Nice! IIUC, these two could be posted as separate PRs. Let's do that so that we can merge them straight away and use them to verify this PR.

Thanks!

PLukas2018:
> Nice! IIUC, these two could be posted as separate PRs. Let's do that so that we can merge them straight away and use them to verify this PR.

I find a problem with running it with Nix, I will try to fix it and open another PR with just the tests.
Btw I included in the tests some parts specific to this PR, so I will add the specific things in this PR after the PR with the tests will be merged.

PLukas2018:
> I find a problem with running it with Nix.

Just an update on the nix problem: I found out that in the nix development environment (`nix develop`) when running from one Python script second one as a subprocess (using eg. `check_call`), if the second Python script is importing diffkemp it fails on  `ModuleNotFoundError`.

This is the problem with `build` command right now, because `bin/diffkemp build` calls `make` which calls python script `cc_wrapper` in which is imported diffkemp, but the module is not found.

As far as I was able to find, the thing is that `setuptoolsShellHook` sets an environment variable `NIX_PYTHONPATH` which contains a path to a directory which contains `*.pth` file which contains the path to the directory with diffkemp. When running python, nix uses the `NIX_PYTHONPATH` variable and sets python to search for modules in all directories which are written in the `.pth` file (so the diffkemp can be imported), but then it unsets the `NIX_PYTHONPATH` which is the reason why subprocesses from python script cannot find diffkemp anymore. 

So far, I did not find  good way how to handle this,  one possible solution would be to add the path to diffkemp directory to the environment variable `PYTHONPATH` then it works fine, I think that the best place to update the variable would be in the `flake.nix` file, but did not find yet if there is way how to get the path to a directory in which is the `flake.nix` file located.

So if somebody understands this better, I would greatly appreciate any insights or suggestions on how to resolve this issue.

PLukas2018:
I solved the nix problem by adding `export PYTHONPATH="$(pwd):$PYTHONPATH"` to `flake.nix` which is not best solution but so far I did not find better.
I opened new PR #324 which contains the tests for build command, after the PR is merged this PR should be then rebased on it.

FrNecas:
I've been playing around with the changes a bit more and trying to build `dietlibc-0.34` (https://www.fefe.de/dietlibc/ ) using this PR results in an error thrown by opt when optimizing `start.ll` file:

```
opt: bin-x86_64/start.ll:1:1: error: expected top-level entity
# 1 "x86_64/start.S"
^
```
The content of `start.ll` is:

```
# 1 "x86_64/start.S"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 381 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "x86_64/start.S" 2
# 1 "./dietfeatures.h" 1
# 2 "x86_64/start.S" 2

.text
.global _start
_start:



 popq %rdi
 movq %rsp,%rsi
 pushq %rdi

 leaq 8(%rsi,%rdi,8),%rdx






 movq %rdx, environ(%rip)
 .hidden environ
# 48 "x86_64/start.S"
 call stackgap
 .hidden stackgap
# 58 "x86_64/start.S"
 movq %rax, %rdi
# 70 "x86_64/start.S"
 call exit

 hlt
.Lstart:
 .size _start,.Lstart-_start
 .section .note.GNU-stack,"",@progbits
```

This works correctly on master because opt is not run there. It seems like that opt requires e.g. a function on the highest level which is not the case in this example (I am not even sure where this LLVM IR is coming from tbh). Maybe in such cases, diffkemp should catch the opt error and leave the file unoptimized?

lenticularis39:
> This works correctly on master because opt is not run there. It seems like that opt requires e.g. a function on the highest level which is not the case in this example (I am not even sure where this LLVM IR is coming from tbh). Maybe in such cases, diffkemp should catch the opt error and leave the file unoptimized?

This seems to be x86 assembly rather than LLVM IR, not sure why that ended up there. Can you post the original command that generated the file?

PLukas2018:
> This works correctly on master because opt is not run there. It seems like that opt requires e.g. a function on the highest level which is not the case in this example (I am not even sure where this LLVM IR is coming from tbh). Maybe in such cases, diffkemp should catch the opt error and leave the file unoptimized?

Thank you for the report. I will add the catching of error, I think it is good idea to do that if the optimization for some reason was unsuccessful rather than to end the build command.

PLukas2018:
> This seems to be x86 assembly rather than LLVM IR, not sure why that ended up there. Can you post the original command that generated the file?

The `cc_wrapper` was called with 
```bash
-I. -isystem include -Os -fstrict-aliasing -momit-leaf-frame-pointer -mfancy-math-387  -W -Wall -Wextra -Wchar-subscripts -Wmissing-prototypes -Wmissing-declarations -Wno-switch -Wno-unused -Wredundant-decls -Wshadow -fpie  -c x86_64/start.S -Wa,--noexecstack -fno-pie -o bin-x86_64/start.o
```
cc_wrapper compiles it like this:
```bash
clang -I. -isystem include -Os -fstrict-aliasing -momit-leaf-frame-pointer -mfancy-math-387 -W -Wall -Wextra -Wchar-subscripts -Wmissing-prototypes -Wmissing-declarations -Wno-switch -Wno-unused -Wredundant-decls -Wshadow -fpie -c x86_64/start.S -Wa,--noexecstack -fno-pie -o bin-x86_64/start.ll -S -emit-llvm -g -fdebug-macro -Wno-format-security -O1 -Xclang -disable-llvm-passes
```
The compilation is successful even tho the output is not LLVM IR (but `-emit-llvm` is used in flags).
In this PR I added running opt with certain passes on files which `cc_wrapper` adds to `diffkemp-wdb` (compiles to LLVM IR), and because this is added to `diffkemp-wdb` I run on it `opt` which results in the error. 

I am not sure how `*.S` files should be handled, should they even be compiled by `cc_wrapper`?

lenticularis39:
> I am not sure how `*.S` files should be handled, should they even be compiled by `cc_wrapper`?
>

Assembly files should ideally be omitted by the wrapper, since it is not possible to get any LLVM IR from them. IIRC the current version of cc_wrapper does not account for assembly source, only for object files and source files.

PLukas2018:
> Assembly files should ideally be omitted by the wrapper, since it is not possible to get any LLVM IR from them. IIRC the current version of cc_wrapper does not account for assembly source, only for object files and source files.

Currently, there is not check what the source file is (C, assembly), I opened PR #326 which checks if the source file is C source file. This should solve the error, but I would still rather leave the catch for running `opt` tool if in other library/project would be some other problems.

diffkemp/llvm_ir/single_c_builder.py
```diff
@@ -13,29 +15,41 @@ class SingleCBuilder(SingleLlvmFinder):
...
     def initialize(self):
+        """Compiles the C file to LLVM IR file and runs passes"""
         # Using clang to compile c file to llvm file.
-        command = [self.clang, "-S", "-emit-llvm", "-g", self.c_file_name,
-                   "-o", self.llvm_file_name]
+        command = [self.clang, self.c_file_name, "-o", self.llvm_file_name]
+        # Note: extending with the default options must be before
+        # extending with the clang_append option
+        # so the optimisations flags can be redefined
+        # (it looks like clang uses last specified optimisaton flag).
```
viktormalik, line 44: What about this:
```suggestion
        # Note: clang uses the last specified optimization level so
        # extending with the default options must be done before
        # extending with the clang_append option.
```
(probably better wrapped)

viktormalik (commented):
Looks good to me, just a couple of minor comments. Thanks!

diffkemp/building/CMakeLists.txt
```diff
@@ -20,9 +20,10 @@ if (NOT DEFINED ENV{WITHOUT_RPYTHON})
   else ()
     set(CC_WRAPPER "${CMAKE_BINARY_DIR}/cc_wrapper-c")
     add_custom_target(cc-wrapper ALL DEPENDS ${CC_WRAPPER})
+    # Note: Setting PYTHONPATH to be able to use diffkemp modules from wrapper.
     add_custom_command(OUTPUT ${CC_WRAPPER}
                        DEPENDS ${CC_WRAPPER_SOURCE}
-                       COMMAND ${RPYTHON} ${CC_WRAPPER_SOURCE} "--output=${CC_WRAPPER}"
+                       COMMAND PYTHONPATH=${CMAKE_SOURCE_DIR} ${RPYTHON} ${CC_WRAPPER_SOURCE} "--output=${CC_WRAPPER}"
```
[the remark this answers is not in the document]
PLukas2018: I added the `$PYTHONPATH` to the end, so we do not have to think about this if we decide to use external modules in the future.

flake.nix
```diff
@@ -120,6 +120,13 @@
             # the function (and all functions it uses) as commands.
             shellHook = ''
               export -f setuptoolsShellHook runHook _eval _callImplicitHook
+
+              # Adding current (diffkemp) directory to PYTHONPATH,
+              # the `diffkemp build` subcommand does not work without it
+              # - `cc_wrapper.py` ends with `ModuleNotFoundError` because
+              # `setuptoolsShellHook` does not make diffkemp package
+              # importable for subprocesses called from python.
+              export PYTHONPATH="$(pwd):$PYTHONPATH"
```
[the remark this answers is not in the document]
PLukas2018: Yes, it probably won't work when run from a different directory (not sure if it si possible), but so far I did not think up a better approach.

PLukas2018:
I addressed the comments.

Btw the [CI failed](https://github.com/diffkemp/diffkemp/actions/runs/8648498687/job/23712463316), which happened to me already in the past, I am not sure if it is something connected with this PR, I think it should not be, because when I reran the CI last time it was ok, I will rerun it again this time too.
It fails on missing kernels in regression tests, but on closer look, there is an error when the cached kernels are being restored. I will probably look closer at this problem (Could it be because we have not yet updated actions to a newer one which uses Node.js 20, probably not, ...).

viktormalik (approved):
Nice, thanks for this work!

PR #314, opened 2023-12-26
Viewer - add visualisation of macros and macro-function diffs

PLukas2018:
This PR adds visualisations of macros to the result viewer.

It was necessary to get information about definitions of macros. I did not use `cscope` because there can be multiple macros with the same name, I took advantage of the fact that the call stack for macros does not contain a line from where it is called at the macro but the line where the macro starts  (btw. I am not sure that this is expected behaviour, I think that it probably should be also a line from which is called the macro as it is for functions).
For the last macro in the call stack, the information about the definition is extracted from Simpll and passed down to DiffKemp and the result viewer.

Also, this PR extends the function, for getting a line where function/type ends, for getting it also for macros.

Because the macros can appear only on one side/version of the call stack, it was necessary to edit components of the viewer and copy also new source files to viewer dir (subdir `new-src`), the dir for the old source files is renamed to `old-src`.

This PR also extends the viewer for visualisation of macro-function/function-macro differences. To make this possible information about the SyntaxDifference kind (macro, macro-function, function-macro, assembly) is added and necessary information about definitions are extracted from LLVM IR (there is a case when the information is not possible to be extracted if the LLVM IR contains information only about declaration of the function and not about its definition).

I compared 8.0, 8.1, 8.2, 8.3 and 8.4 versions of RHEL/CentOS to check I did not break the comparison.
And I went over some differences in the viewer.

Screenshots of the visualisation:

Macro code
image
Macro which is only located in the old call stack
image
Macro which is only located in the new call stack
image
Macro code
image
Macro difference
image
Macro-function difference
image

### DIFF
--- a/view/src/components/DiffViewWrapper.jsx
+++ b/view/src/components/DiffViewWrapper.jsx
@@ -55,17 +55,73 @@ function createFileStructure(oldCode, oldStart, newStart) {
  * @returns Returns array of lines to be highlighted acceptable by Diff.
  */
 function getLinesToHighlight(linesToShow, hunks) {
-  if (linesToShow) {
+  const changesToHighlight = [];
+  // Do not highlight when linesToShow[] is set to -1.
+  if (!linesToShow) return [];
+  if (linesToShow[0] !== -1) {
     const oldChange = findChangeByOldLineNumber(hunks, linesToShow[0]);
-    const newChange = findChangeByNewLineNumber(hunks, linesToShow[1]);
-    if (oldChange && newChange) {
+    if (oldChange) {
       const oldChangeKey = getChangeKey(oldChange);
-      const newChangeKey = getChangeKey(newChange);
-      if (oldChangeKey === newChangeKey) return [oldChangeKey];
-      return [oldChangeKey, newChangeKey];
+      changesToHighlight.push(oldChangeKey);
     }
   }
-  return [];
+  if (linesToShow[1] !== -1) {
+    const newChange = findChangeByNewLineNumber(hunks, linesToShow[1]);
+    if (newChange) {
+      const newChangeKey = getChangeKey(newChange);
+      changesToHighlight.push(newChangeKey);
+    }
+  }
+  return changesToHighlight;
+}
+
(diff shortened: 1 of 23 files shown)


view/src/components/DiffViewWrapper.jsx
```diff
@@ -89,37 +145,60 @@ function getSyntaxHighlightedTokens(hunks) {
...
   const file = useMemo(() => {
-    if (diff !== '') {
-      const [fileObject] = parseDiff(diff);
-      return fileObject;
+    let returnValue;
+    if (onlyOneSide) {
+      const onlyOld = (newCode === '');
+      const code = onlyOld ? oldCode : newCode;
+      const start = onlyOld ? oldStart : newStart;
+      const end = onlyOld ? oldEnd : newEnd;
```
viktormalik, line 183: Nit: can JS do something like this?
```suggestion
      const [code, start, end] = onlyOld ? [oldCode, oldStart, oldEnd] : [newCode, newStart, newEnd];
```

view/src/components/Difference.jsx
```diff
@@ -9,7 +9,7 @@ import Col from 'react-bootstrap/Col';
 import Row from 'react-bootstrap/Row';
 import Alert from 'react-bootstrap/Alert';
 
-import Callstack from './Callstack';
+import Callstack, { CallType } from './Callstack';
```
viktormalik, line 12: Shouldn't be?
```suggestion
import { Callstack, CallType } from './Callstack';
```

viktormalik (commented):
Great amount of work! The screenshots look really nice.

Unfortunately, it's too much code for me to review at once. I reviewed the first 7 commits (i.e. up to function-macro differences). There's a number of comments but it will certainly need another review.

Do you think it would be possible to split the PR into several smaller ones? That would simplify reviewing and allow us to merge the work piece by piece. For example, it looks like we can at least separate visualization of macro differences from macro-function differences. On top of that, if you could identify some smaller changes that can be posted separately, that would be great.

PR #319, opened 2024-02-22
Add pass that removes tail tag from function calls

xkucma00:
Adds pass to remove ```tail``` qualifier from function calls, fixing the issue where equal call instructions, one with ```tail``` and one without, were evaluated as unequal.

I could also move the functionality into an existing pass, eg ```RemoveFunctionMetadataPass```, if that made sense.


viktormalik:
> I could also move the functionality into an existing pass, eg `RemoveFunctionMetadataPass`, if that made sense.

Yes, that pass looks like a good target for this.

viktormalik (approved):

Issue #260, opened 2023-05-24
Logging: do not dump function values


viktormalik:
If a difference is found in a function call, the logger (with `-d` set) prints the entire function bodies of the called functions. This pollutes the output (the functions can be quite long) and doesn't provide any relevant information.

Reproducer: `deferred_free_range` function between `linux-4.18.0-80.el8` and `linux-4.18.0-147.el8`.

PLukas2018:
Resolved by #320 .

PR #320, opened 2024-03-04
Logger: logging function name instead of its body

PLukas2018:
This PR solves issue #260 .
For methods like cmpValues in Simpll DifferentialFunctionComparator, if the value was llvm::Function, the logger dumped the entire body of the function (which polluted the output), this PR changes it and logs only the function name.

DiffKemp compare output with logger enabled (`-d` set)
- [before the change](https://github.com/viktormalik/diffkemp/files/14478290/old-output.log)
- [after the change](https://github.com/viktormalik/diffkemp/files/14478332/new-output.log)

### DIFF
--- a/diffkemp/simpll/Logger.cpp
+++ b/diffkemp/simpll/Logger.cpp
@@ -90,7 +90,12 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &out,
         out << *(value.type);
         break;
     case value.llvmValue:
-        out << *(value.value);
+        // In case the value is a function, log its name instead of its body.
+        if (auto *fun = dyn_cast<Function>(value.value)) {
+            out << fun->getName();
+        } else {
+            out << *(value.value);
+        }
         break;
     }
     return out;


diffkemp/simpll/Logger.cpp
```diff
@@ -90,7 +90,12 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &out,
         out << *(value.type);
         break;
     case value.llvmValue:
-        out << *(value.value);
+        // In case the value is a function, log its name instead of its body.
+        if (isa<Function>(value.value)) {
+            out << dyn_cast<Function>(value.value)->getName();
```
viktormalik, line 95: LLVM code style:
```suggestion
        if (auto *fun = dyn_cast<Function>(value.value)) {
            out << fun->getName();
```

PLukas2018: Thanks, fixed.

viktormalik (approved):
Thanks!

PR #322, opened 2024-03-18
Integrate formal SMT-based methods

FrNecas:
This PR introduces an SMT-based comparison of short sequential snippets using Z3 solver. The general approach can be described as follows:
 - when we find a difference and no pattern is applicable, try to check equivalence using an SMT solver
 - the differing code snippets to be analyzed are determined based on searching for a synchronization point, i.e. a pair of instructions after which the code can be synchronized until the end of the basic block
 - we construct a formula expressing the equivalence of the identified snippets -- if the inputs are the same and the blocks are executed, the outputs of the blocks must be the same. Since checking satisfiability is a hard problem, we apply a timeout.
 - furthermore, the extension facilitates inverse-branching-condition pattern, e.g. it can detect a more complex change such as `icmp slt %1, 101` being an inverse of `icmp sgt %1, 100` if `%1` is an integer.

Something to discuss:
 - The new component is off by default and must be enabled using `--use-smt`. Do we always want to build this extension? The problem is that it introduces `z3` as a dependency which is quite significant. However, since `SMTBlockComparator` is tightly coupled with `DifferentialFunctionComparator`, trying to exclude it from the build unless explicitly turned on would require quite a lot of preprocessor if-s. Furthermore, even if it was conditionally removed from simpll, it would still be part of the python frontend CLI which could be quite confusing.

Depends on: #323 #325 (needed for correct functionality in some more complicated cases, e.g. the KABI experiment)


FrNecas:
Hm, trying to run this on the KABI experiments results in a segfault... I guess more debugging time :/

lenticularis39:
Since LLVM is built for performance, it usually just segfaults when something goes wrong, one needs to extensively use GDB :D

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -2450,3 +2450,341 @@ TEST_F(DifferentialFunctionComparatorTest, ReorganizedLocalVariables) {
...
+    auto AddL = BinaryOperator::Create(
+            BinaryOperator::Add, ConstL1, ConstL2, "", BBL);
+    auto CondL = ICmpInst::Create(llvm::Instruction::ICmp,
+                                  llvm::CmpInst::ICMP_SLT,
+                                  AddL,
+                                  ConstantInt::get(Type::getInt32Ty(CtxL), 101),
+                                  "",
+                                  BBL);
+    auto ConstR1 = ConstantInt::get(Type::getInt32Ty(CtxR), 2);
+    auto ConstR2 = ConstantInt::get(Type::getInt32Ty(CtxR), 2);
```
PLukas2018, line 2773: Should not it be `0` instead of `2`?

diffkemp/simpll/SMTBlockComparator.cpp
```diff
@@ -0,0 +1,842 @@
...
+        return c.fpa_val(
+                dyn_cast<ConstantFP>(constant)->getValueAPF().convertToFloat());
+#else
+        return c.fpa_val(
+                dyn_cast<ConstantFP>(constant)->getValue().convertToFloat());
+#endif
+    } else if (constant->getType()->isDoubleTy()) {
+#if LLVM_VERSION_MAJOR < 11
+        return c.fpa_val(
+                dyn_cast<ConstantFP>(constant)->getValueAPF().convertToFloat());
```
PLukas2018, line 170: Shouldn't it be
```suggestion
                dyn_cast<ConstantFP>(constant)->getValueAPF().convertToDouble());
```
?

diffkemp/simpll/SMTBlockComparator.cpp
```diff
@@ -0,0 +1,842 @@
...
+        try {
+            int res = compareSnippets(StartL, InstL, StartR, InstR, false);
+
+            if (res == 0)
+                return res;
+
+            // Try to invert a cmp operation whose result is an output value.
+            // This aims to facilitate more complex refactorings than
+            // inverse-branch-condition pattern, e.g. replacing x < 101 with
+            // x > 101 (where x is an integer) and swapping branches.
```
PLukas2018, line 788: Shouldn't it be
```suggestion
            // inverse-branch-condition pattern, e.g. replacing x < 101 with
            // x > 100 (where x is an integer) and swapping branches.
```
?

viktormalik:
> * The new component is off by default and must be enabled using `--use-smt`. Do we always want to build this extension? The problem is that it introduces `z3` as a dependency which is quite significant. However, since `SMTBlockComparator` is tightly coupled with `DifferentialFunctionComparator`, trying to exclude it from the build unless explicitly turned on would require quite a lot of preprocessor if-s. Furthermore, even if it was conditionally removed from simpll, it would still be part of the python frontend CLI which could be quite confusing.

I would say, let's not complicate things now and always build the extension in. If we wanted to make the z3 dependency optional, probably the easiest way is to have an alternative (ifdef-ed) version of `SMTBlockComparator` which has empty implementations of the exported functions.

FWIW, I would also love to make it on by default but IIRC, the overhead is still quite high, especially due to finding the snippets.

diffkemp/cli.py
```diff
@@ -119,6 +119,9 @@ def make_argument_parser():
                             help="compare only selected function")
     compare_ap.add_argument("--custom-patterns", "-p",
                             help="custom pattern file or configuration")
+    compare_ap.add_argument("--use-smt",
+                            help="Use SMT-based checking of short snippets",
+                            action="store_true")
```
viktormalik, line 133: Point for discussion: should we make the contents of this PR a new "builtin pattern" instead of a new CLI option? I can imagine that we'll eventually have more patterns which require SMT and we'd want to allow selecting just some. Using any such pattern would automatically turn on the SMT support.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1112,15 +1124,15 @@ int DifferentialFunctionComparator::cmpBasicBlocks(
 
             // Try to find a match by moving one of the instruction iterators
             // forward (find a code relocation).
-            if (config.Patterns.Relocations
+            if (config.Patterns.Relocations && !suppressRelocationsAndSMT
```
viktormalik, line 1127: The `suppressRelocationsAndSMT` parameter feels a bit clumsy, TBH. What if we find out that we need to turn off more builtin patterns in future?

Two alternatives come to mind:
1. Turn off `config.Patterns.Relocations` and `config.UseSMT` for the particular call of `cmpBasicBlocksFromInstructions`.
2. Make `SMTComparator` use a separate instance of `DifferentialFunctionComparator` having a different config.

diffkemp/simpll/SMTBlockComparator.cpp
```diff
@@ -16,7 +16,67 @@
 
 using namespace llvm;
 
+void SMTBlockComparator::findSnippetEnd(BasicBlock::const_iterator &InstL,
+                                        BasicBlock::const_iterator &InstR) {
```
viktormalik, line 28: Note: this is something we will need to other extensions, too (e.g. equivalence slicing introduced by #182) so we should at least think about a better place for this method b/c we'll want to have it unified.

viktormalik (commented):
Reviewed the first 2 commits so far. There's a couple of points for discussion so I'm posting them straight away.

Also, could you please expand the commit messages? I prefer a commit to explain what it does (even if it is obvious) so that its intentions are clear when we get back to it in future.

diffkemp/cli.py
FrNecas: Interesting idea. I'd suggest keeping it as an option for now -- all patterns are enabled by default, right? Making this a pattern that's disabled by default feels a bit inconsistent. Furthermore, I perceive the builtin patterns as something rather lightweight and quite fast which definitely is not the case here. We can obviously change it later if we have more SMT-based patterns.

diffkemp/simpll/DifferentialFunctionComparator.cpp
FrNecas: Yeah, I wasn't a big fan of this either but couldn't think of a better solution. The first alternative feels quite dirty as well, `Config` represents the user configuration and should therefore IMO be constant throughout the whole run. Furthermore, I am not sure if it is even possible to use it, `cmpBasicBlocks` is declared as const inside LLVM API, i.e., it cannot change the values of data fields. So even if we changed `config` to non-const, it probably wouldn't be possible to modify its fields inside `SMTBlockComparator` since it is called from `cmpBasicBlocks`.

I can try the second alternative but there may be other drawbacks. One thing that comes to mind -- `DifferentialFunctionComparator` has some state, e.g. some patterns keep track of various information. We would somehow have to transfer this information (the current constructor is rather basic and accepts only a few arguments). Maintaing the list of fields that need to be transferred for the new `DifferentialFunctionComparator` instance to be fully functional seems quite tedious. WDYT?

diffkemp/simpll/SMTBlockComparator.cpp
FrNecas: Idea: we could have an abstract class (naming TBD, e.g. `SearchSynchronizationComparator` comes to mind) that has a single field (of type `DifferentialFunctionComparator`) and implements this method. The slicer and SMTBlockComparator could then inherit from this class

viktormalik: Yes, that sounds good. An alternative would be to have a helper class which does pretty much the same but the slicer and `SMTBlockComparator` use it as a field. I'm not sure which approach would be better so I'd stick with the current implementation for now and introduce the abstraction only when necessary.

diffkemp/cli.py
viktormalik: > Interesting idea. I'd suggest keeping it as an option for now -- all patterns are enabled by default, right? Making this a pattern that's disabled by default feels a bit inconsistent. 

No, there are patterns which are off by default, namely `control-flow-only` and `type-casts`.

> Furthermore, I perceive the builtin patterns as something rather lightweight and quite fast which definitely is not the case here. We can obviously change it later if we have more SMT-based patterns.

Yes, but it feels like changing it later would remove the new `--use-smt` option (or at least change) which is not a nice breakage of the CLI. I'm a bit indecisive here, too, but I don't see any fundamental difference between this and built-in patterns. We'd just have to come up with a name for it.

diffkemp/simpll/SMTBlockComparator.cpp
```diff
@@ -0,0 +1,842 @@
...
+            }
+            if (fComp->mappedValuesBySn.find(snIt->second)
+                != fComp->mappedValuesBySn.end()) {
+                auto values = fComp->mappedValuesBySn[snIt->second];
+                outL.erase(values.first);
+                outR.erase(values.second);
+                auto left = createExprFromValue(c, LPrefix, values.first);
+                auto right = createExprFromValue(c, RPrefix, values.second);
+                postcond = postcond && (left == right);
+            }
```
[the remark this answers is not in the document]
FrNecas: While it's similar, there is still the added aspect of tracking the remaining output variables. It probably could be factored out but it feels like more hassle than it's worth for such a short piece of duplication.

tests/unit_tests/simpll/CMakeLists.txt
```diff
@@ -34,7 +34,16 @@ exec_program(llvm-config ARGS --libs irreader passes support OUTPUT_VARIABLE llv
...
+# Try to find system-wide Z3 (e.g. local build)
+find_package(Z3 CONFIG)
+if (NOT ${Z3_FOUND})
+    # Use our FindZ3.cmake (e.g. nix build)
+    find_package(Z3 REQUIRED MODULE)
+endif()
+
+target_include_directories(runTests PRIVATE ${Z3_CXX_INCLUDE_DIRS})
+target_link_libraries(runTests gtest simpll-lib ${llvm_libs} ${Z3_LIBRARIES})
+target_compile_options(runTests PRIVATE ${Z3_COMPONENT_CXX_FLAGS})
```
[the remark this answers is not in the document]
FrNecas: I would expect this as well but compilation of tests fails without this, which is weird... Maybe we are doing something wrong in our test setup?

FrNecas:
I've rebased the PR and addressed the comments (hopefully all of them, other than the more high-level comment if this should be a pattern or an option -- which we probably need to discuss more closely). This is probably ready to be reviewed again.

FrNecas:
Option converted to a pattern (named `smt-sequential-blocks` -- if you find a better name, happy to rename). I decided to keep the smt-timeout option not specific to the pattern for now, as I believe that even if we add new patterns based on SMT solving (like the more complex branching stuff), it doesn't hurt if they share the same timeout -- if a user increases the timeout, they are fine with diffkemp running longer and so it doesn't really matter if the longer runtime comes from our current sequential-blocks pattern or from some other more complex patterns.

diffkemp/cli.py
```diff
@@ -163,7 +178,8 @@ def make_argument_parser():
                         "control-flow-only",
                         "inverse-conditions",
                         "reordered-bin-ops",
-                        "group-vars"]
+                        "group-vars",
+                        "smt-sequential-blocks"]
```
viktormalik, line 182: We don't support all kinds of operations within sequential blocks, do we? For instance memory reads/writes must exactly match. What about `sequential-alu-ops` for the name? I'm not sure if we need the `smt-` prefix either.

viktormalik (commented):
Looks great! Just some naming stuff to resolve, otherwise it's pretty much ready.

viktormalik (approved):
Looks good!

We have CI failure, though :cry:

viktormalik:
And we're done here :partying_face: 

Many thanks!

PR #323, opened 2024-03-20
Check types when applying inverse cond pattern

FrNecas:
In certain conditions, not checking the type could lead to an assertion failure in cmpOperationsWithOperands. FunctionComparator::cmpOperations checks the types of operands, however if the instructions are an inverse icmp instructions, Diffkemp discards this result. We need to differentiate, whether the non-zero return code from cmpOperations was caused by a difference in types or difference in values before applying the pattern.

The attached test resulted in an assertion failure with current diffkemp version on this line:

https://github.com/diffkemp/diffkemp/blob/789ac18c2c6864b003eac2d77d155cdf87da0a72/diffkemp/simpll/DifferentialFunctionComparator.cpp#L1849-L1850

I thought of two possible solutions to this problem:

1. the one submitted in this PR
2. checking type equality in `DifferentialFunctionComparator::cmpValues`

The second option sounds a bit cleaner but I feel like it could have some side-effects that I may not be aware of (since the method is called quite often from various contexts) and it may be intentional that the comparison is left out. Therefore, I decided to use the "safer" alternative.

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -377,9 +377,13 @@ int DifferentialFunctionComparator::cmpOperations(
             auto CmpL = dyn_cast<CmpInst>(L);
             auto CmpR = dyn_cast<CmpInst>(R);
 
+            // Only apply inverse condition pattern if the operands are of the
+            // same type, i.e. cmpOperations returned non-0 because of the
+            // values rather than the types.
             // It is sufficient to compare the predicates here since the
-            // operands are compared in cmpBasicBlocks.
-            if (CmpL->getPredicate() == CmpR->getInversePredicate()) {
+            // values of operands are compared in cmpBasicBlocks.
+            if (cmpOperandTypes(CmpL, CmpR) == 0
+                && CmpL->getPredicate() == CmpR->getInversePredicate()) {
                 if (checkInverseCondUsers(L) && checkInverseCondUsers(R)) {
                     PREP_LOG("inv-cond-add", L, R);
                     LOG_KEEP_FORCE(0);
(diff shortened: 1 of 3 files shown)


diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -377,9 +377,20 @@ int DifferentialFunctionComparator::cmpOperations(
...
 
+            // Only apply inverse condition pattern if the operands are of the
+            // same type, i.e. cmpOperations returned non-0 because of the
+            // values rather than the types.
+            bool typesMatch = true;
+            for (unsigned i = 0, e = CmpL->getNumOperands(); i != e; ++i) {
+                if (cmpTypes(L->getOperand(i)->getType(),
+                             R->getOperand(i)->getType()))
+                    typesMatch = false;
+            }
```
viktormalik, line 388: This looks like something that could be useful from other places. Maybe extract into a function (e.g. `cmpOperandTypes`)?

FrNecas: Good idea, done.

diffkemp/simpll/DifferentialFunctionComparator.cpp
```diff
@@ -1853,6 +1857,17 @@ int DifferentialFunctionComparator::cmpOperationsWithOperands(
...
 }
 
+int DifferentialFunctionComparator::cmpOperandTypes(
+        const llvm::Instruction *L, const llvm::Instruction *R) const {
+    if (int Res = cmpNumbers(L->getNumOperands(), R->getNumOperands()))
+        return Res;
+    for (unsigned i = 0, e = L->getNumOperands(); i != e; ++i)
+        if (int Res = cmpTypes(L->getOperand(i)->getType(),
+                               R->getOperand(i)->getType()))
+            return Res;
```
viktormalik, line 1867: We prefer braces around multi-line loop bodies (even if it's a just single command):
```suggestion
    for (unsigned i = 0, e = L->getNumOperands(); i != e; ++i) {
        if (int Res = cmpTypes(L->getOperand(i)->getType(),
                               R->getOperand(i)->getType()))
            return Res;
    }
```

viktormalik (commented):
Looks great, just one nit. Thanks!

viktormalik (approved):

PR #326, opened 2024-04-02
cc wrapper: skipping not C files

PLukas2018:
The cc wrapper used to try to compile also assembly, this PR changes it by checking if arguments contain a C source file (`*.c`) if not then it does not tries to compile it. For commands containing linking changes nothing.

To check if I did not break anything I build `musl-1.2.1` -- `musl-1.2.5` and compared snapshots between master and this branch, the snapshots between the branches contain the same files. In musl there were some cases of compilation `*.s` files, before this PR cc_wrapper did try to compile it but because the compilation did not created output file, the file was never added to db file https://github.com/diffkemp/diffkemp/blob/54deef1fbdaabbc6014223a2cf1b4418641b76bb/diffkemp/building/cc_wrapper.py#L210. Now it does not even try to compile it.

I also build `dietlibc-0.34`, for this library there were before this PR multiple cases of `*.S` files compilation, because the compilation did not failed, the output file was created and added to db file. For this I also compared snapshots created on master branch and this branch, the snapshots contains same files (the `*.S` files are compiled on master branch but because they do not contain LLVM IR functions they were/are not copied to snapshot dir, therefore the snapshots still contains the same files). Now the `*.S` files are not compiled and therefore not added to db file.

### DIFF
--- a/tests/testing_projects/make_based/Makefile
+++ b/tests/testing_projects/make_based/Makefile
@@ -2,9 +2,18 @@
 CC = gcc
 CFLAGS = -O2 -std=c99
 
-.PHONY: all clean
+.PHONY: default with-assembly clean
 
-all: file.o
+default: file.o
+
+# Assembly files should not be compiled to `.ll`
+with-assembly: default mod.o sub.o
+
+# Note: mod.o (mod.S) is implicitly compiled using CC, but sub.o (sub.s)
+#   is implicilty compiled using AS therefore for sub.o is neccessary to
+#   add explicit target to use CC (for testing cc_wrapper behaviour).
+sub.o: sub.s
+	$(CC) -c $@ $^
 
 clean:
 	rm -f *.o *.ll
--- a/tests/testing_projects/make_based/Makefile
+++ b/tests/testing_projects/make_based/Makefile
@@ -2,12 +2,15 @@
 CC = gcc
 CFLAGS = -O2 -std=c99
 
-.PHONY: default with-assembly clean
+.PHONY: default with-assembly with-linking clean
 
 default: file.o
 
 # Assembly files should not be compiled to `.ll`
 with-assembly: default mod.o sub.o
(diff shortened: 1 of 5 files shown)


viktormalik (commented):
Nice! I'm wondering if we could add a `.S` file (or any other non-source file) to our testing `Makefile` project and cover this in the tests somehow.

PLukas2018:
Sure, I can try to add a test/s.

PLukas2018:
I added the tests, which test that for `.S` and `.s` file is not created LLVM IR and that it is not added to db file.
I also added a test for compilation with linking (in this case creating shared library from `.o` file).

tests/testing_projects/make_based/Makefile
```diff
@@ -2,9 +2,23 @@
...
 CFLAGS = -O2 -std=c99
 
-.PHONY: all clean
+.PHONY: default with-assembly with-linking clean
 
-all: file.o
+default: file.o
+
+# Assembly files should not be compiled to `.ll`
+with-assembly: default mod.o sub.o
```
viktormalik, line 10: Aren't we missing a rule for the `mod.o` target? Ot is it implicit?

viktormalik (approved):
Looks great, thanks! Just one question.

tests/testing_projects/make_based/Makefile
PLukas2018: It is implicit. For `sub.o` (`sub.s`) I wrote it explicitly because implicitly would be used `as` (assembler) which would mean that `cc_wrapper`would not be called.  For `mod.o` (`mod.S`) is implicitly used `$(CC)`.

viktormalik: Thanks for the explanation. Let's put it to a comment as I'm sure that I'll have the same question when I look at the code in future.

PR #329, opened 2024-04-04
Tests in llvm ir

PLukas2018:
Introduction to unit testing of `DifferentialFunctionComparator` using LLVM IR.
Tests can be written like this:
```cpp
TEST_F(DFCLlvmIrTest, ReorderedBinaryOperationDifferentOperands) {
    auto left = R"(define i8 @f() {
        %1 = add i8 0, 0
        ret i8 %1
    })";
    auto right = R"(define i8 @f() {
        %1 = add i8 1, 0
        ret i8 %1
    })";
    CREATE_FROM_LLVM(left, right);
    ASSERT_EQ(DiffComp->compare(), 1);
}
```

PR contains 12 rewritten unit tests from 13 which were using only `compare` method, (the last one is working with a custom pattern, which should also be rewritable but needs additional logic). 

There are more than 28 other tests which test a specific `DifferentialFunctionComparator` method, some of these tests could be also rewritable:
a) by rewriting it and testing it as whole - we will lose the level of detail,
b) by testing specific instructions for eg. like this:
```cpp
/// Tests a comparison of two GEPs of a structure type with indices compared by
/// value.
TEST_F(DFCLlvmIrTest, CmpGepsSimple) {
    auto left = R"(
        %struct = type { i8, i16 }
        define void @f() {
            %var = alloca %struct
            %gep1 = getelementptr %struct, %struct* %var, i32 0, i32 0
            %gep2 = getelementptr %struct, %struct* %var, i32 0, i32 0
            ret void
        }
    )";

    auto right = R"(
        %struct = type { i8, i16 }
        define void @f() {
            %var = alloca %struct
            %gep1 = getelementptr %struct, %struct* %var, i32 0, i32 0
            %gep2 = getelementptr %struct, %struct* %var, i32 0, i32 1
            ret void
        }
    )";
    CREATE_FROM_LLVM(left, right);
    ASSERT_EQ(testCmpGEPs("gep1"), 0);
    ASSERT_EQ(testCmpGEPs("gep2"), 1);
}
```


PLukas2018:
Btw for now I used typed pointer in LLVM IR, therefore the tests are run on all LLVM versions which we test in CI. 
I tried to parse typed pointer in LLVM 18 and it worked, so probably the not supporting typed pointer is meant only in API
> Typed pointers are no longer supported and the -opaque-pointers option has been removed. See the [opaque pointers](https://releases.llvm.org/17.0.1/docs/OpaquePointers.html) documentation for migration instructions.

<https://releases.llvm.org/17.0.1/docs/ReleaseNotes.html#id5>

viktormalik:
> b) by testing specific instructions for eg. like this:
> 
> ```c++
>     [...]
>     CREATE_FROM_LLVM(left, right);
>     ASSERT_EQ(testCmpGEPs("gep1"), 0);
>     ASSERT_EQ(testCmpGEPs("gep2"), 1);
> }
> ```

This looks great, that would be a nice functionality! But let's leave it to a follow-up PR.

I'll wait with reviewing this one until you move it out of the Draft state.

viktormalik:
> Btw for now I used typed pointer in LLVM IR, therefore the tests are run on all LLVM versions which we test in CI. I tried to parse typed pointer in LLVM 18 and it worked, so probably the not supporting typed pointer is meant only in API

I'm a bit surprised that LLVM 18 can parse IR files with explicit pointers but if that works, let's use it for now. If some future LLVM version stops supporting this, it should be easy to rewrite it to opaque pointers.

PLukas2018:
> I'll wait with reviewing this one until you move it out of the Draft state.

I moved it out of draft, but the LLVM IR tests will maybe make more sense to move to a new file, but for now I am leaving it as it is so it easier to compare the rewritten tests.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -203,20 +215,24 @@ class DifferentialFunctionComparatorTest : public ::testing::Test {
                 FunctionType::get(Type::getVoidTy(CtxL), {}, false),
                 GlobalValue::ExternalLinkage,
                 "F",
-                &ModL);
+                &*ModL);
```
viktormalik, line 218:
```suggestion
                ModL.get());
```
same below (multiple occurrences)

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -314,6 +330,75 @@ class DifferentialFunctionComparatorTest : public ::testing::Test {
...
+        if (ModL == nullptr) {
+            errorMsg = "Error while parsing left LLVM IR " + errorMsg;
+            return false;
+        }
+        ModR = stringToModule(right, CtxR);
+        if (ModR == nullptr) {
+            errorMsg = "Error while parsing right LLVM IR " + errorMsg;
+            return false;
+        }
+        FL = getFunByName(*ModL, "f");
```
viktormalik, line 354: There's `llvm::Module::getFunction(StringRef Name)`, could we use it?

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -314,6 +330,75 @@ class DifferentialFunctionComparatorTest : public ::testing::Test {
...
+    /// if unsuccessful sets error message.
+    std::unique_ptr<Module> stringToModule(const char *llvm, LLVMContext &ctx) {
+        SMDiagnostic err;
+
+        StringRef strRef(llvm);
+        std::unique_ptr<MemoryBuffer> memBuf =
+                MemoryBuffer::getMemBuffer(strRef);
+        MemoryBufferRef memBufRef(*memBuf);
+
+        std::unique_ptr<Module> mod(parseIR(memBufRef, err, ctx));
```
viktormalik, line 378: I think that it should be possible to directly use `*memBuf` here:
```suggestion
        std::unique_ptr<Module> mod(parseIR(*memBuf, err, ctx));
```
and you can spare one variable.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -314,6 +330,75 @@ class DifferentialFunctionComparatorTest : public ::testing::Test {
...
+        if (mod == nullptr) {
+            // Parsing was unsuccessful sets error message.
+            raw_string_ostream stream(errorMsg);
+            std::string info = "(" + std::to_string(err.getLineNo()) + ":"
+                               + std::to_string(err.getColumnNo()) + ")\n";
```
viktormalik, line 383:
```suggestion
            const std::string info = "(" + std::to_string(err.getLineNo()) + ":"
                               + std::to_string(err.getColumnNo()) + ")\n";
```

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -314,6 +330,75 @@ class DifferentialFunctionComparatorTest : public ::testing::Test {
...
+            err.print("", stream);
+            errorMsg = info + errorMsg;
```
viktormalik, line 385: Nit: just for the purpose of efficiency, could we do
```suggestion
            errorMsg << info;
```
and have the line:column info at the end?

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -1496,469 +1581,185 @@ TEST_F(DifferentialFunctionComparatorTest, CmpSkippedBitcast) {
...
+            %var = alloca %struct.s
+            %gep1 = getelementptr %struct.s, %struct.s* %var, i32 0, i32 0
+            %load1 = load i32, i32* %gep1
+            %icmp = icmp ne i32 %load1, 0
+            br i1 %icmp, label %1, label %2
+        1:
+            ; instructions which will be relocated
+            %gep2 = getelementptr %struct.s, %struct.s* %var, i32 0, i32 1
+            %load2 = load i32, i32* %gep2
+
```
viktormalik, line 1695: Extra empty line?

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -1496,469 +1581,185 @@ TEST_F(DifferentialFunctionComparatorTest, CmpSkippedBitcast) {
...
+        }
+    )";
+    auto right = R"(
+        %struct.s = type { i32, i32 }
+        define i32 @f() {
+            %var = alloca %struct.s
+            %gep1 = getelementptr %struct.s, %struct.s* %var, i32 0, i32 0
+            %load1 = load i32, i32* %gep1
+
+            ; following two instructions were safely relocated
```
viktormalik, line 1708: nit
```suggestion
            ; the following two instructions were safely relocated
```

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -2314,139 +2090,54 @@ TEST_F(DifferentialFunctionComparatorTest, CustomPatternSkippingInstruction) {
...
+TEST_F(DFCLlvmIrTest, SkipRepetitiveLoad) {
+    auto left = R"(define i32 @f() {
+            %1 = alloca i32
+            %2 = load i32, i32* %1
+            %3 = icmp ne i32 %2, 0
+            br i1 %3, label %4, label %5
+        4:
+            br label %5
+        5:
+            ret i32 %2
```
viktormalik, line 2102: Just want to mention it, the original test seems to use:
```suggestion
            br i1 %3, label %4, label %5
        4:
            br label %6
        5:
            br label %6
        6:
            ret i32 %2
```
But your version seems to be sufficient.

viktormalik (commented):
Beautiful! Just a couple of comments but overall it's a great simplification.

I agree that we should move the migrated tests to a separate test file. Good idea for keeping them in the original file for simpler review, I think that you can move them now.

Thanks for this work!

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
PLukas2018: Thanks, fixed.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
PLukas2018: Yes, thanks.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
PLukas2018: I changed it -- I am printing the message right away to stderr.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
PLukas2018: I changed it -- I am using stream instead.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
PLukas2018: Removed.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
PLukas2018: Thanks, fixed.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
PLukas2018: Yes, it looks like it is sufficient.

viktormalik (approved):
Nice work and a beautiful cleanup of the testing code!

It just needs a rebase b/c there's a conflict with some recently merged changes.

PLukas2018:
Thanks, I cleaned up the commits and rebased it on master.

PLukas2018:
> > b) by testing specific instructions for eg. like this:
> > ```c++
> >     [...]
> >     CREATE_FROM_LLVM(left, right);
> >     ASSERT_EQ(testCmpGEPs("gep1"), 0);
> >     ASSERT_EQ(testCmpGEPs("gep2"), 1);
> > }
> > ```
> 
> This looks great, that would be a nice functionality! But let's leave it to a follow-up PR.

Implemented in #358 PR.

PR #330, opened 2024-04-12
Fix handling of union types

FrNecas:
Union types from C are represented by a structure type in LLVM IR generated by clang. When we are applying the StructAlignment pattern, there is, therefore, some ambiguity.

The previous implementation did not work correctly if C union and C structure types were being compared in cmpTypes -- the result was dependant on what was the Left and Right type due to a bug in the implementation. In this case, Left would be considered as the potential union type, even if it was the other way around.

Reproducer of the bug -- swapping the order of snapshots influences the analysis verdict:
```
$ bin/diffkemp -v compare ../diffkemp-experiments/kabi/snapshots/linux-4.18.0-240.el8 ../diffkemp-experiments/kabi/snapshots/linux-4.18.0-305.el8 --report-stat --extended-stat -f rcu_barrier

$ bin/diffkemp -v compare ../diffkemp-experiments/kabi/snapshots/linux-4.18.0-305.el8 ../diffkemp-experiments/kabi/snapshots/linux-4.18.0-240.el8 --report-stat --extended-stat -f rcu_barrier
```

### DIFF
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -1662,6 +1662,13 @@ int DifferentialFunctionComparator::cmpCallsWithExtraArg(
     return 1;
 }
 
+bool DifferentialFunctionComparator::isUnionTy(const Type *Ty) {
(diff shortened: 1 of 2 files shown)


viktormalik (approved):
Wow, nice catch!

PR #333, opened 2024-04-20
Caching information about definitions of differing macros

PLukas2018:
First part of commits from #314 .
This PR adds:
- caching of Syntax difference kind (because there can be multiple kinds of syntax differences - `macro`, `assembly`, `macro-function`, `function-macro`) --  I think it could be useful to have this saved in Python, personally I used this information for visualisation of `function-macro`/`macro-function` differences in the result viewer, but right now I am thinking it is not absolutely necessary because I should be able to deduce this based on names of the differing 'objects', but I am still thinking it could be useful to have it.
- caching of the definition of differing/last macro in Syntax difference -- this is useful for showing code/differences of the macro in the result viewer.

I will reference comments/questions from #314 which are relevant to this PR.

### DIFF
--- a/diffkemp/semdiff/caching.py
+++ b/diffkemp/semdiff/caching.py
@@ -225,23 +225,34 @@ class ComparisonGraph:
             but the line where the definition of the caller starts.
         :param body: A tuple containing the body of the 'object' for both
             modules.
+        :param diff_def: Tuple (first, second), for macro difference contains
+            information about definition (name, file, line) of the differing
+            macros, otherwise contains None.
         """
-        def __init__(self, kind, name, parent_fun, callstack, body):
+        def __init__(self, kind, name, parent_fun, callstack, body,
+                     diff_def=None):
             self.kind = kind
             self.name = name
             self.parent_fun = parent_fun
             self.callstack = callstack
             self.body = body
+            self.diff_def = diff_def
 
         @classmethod
         def from_yaml(cls, nonfun_diff):
             """Generates a SyntaxDiff object from YAML returned by SimpLL."""
+            diff_def = None
+            if ("diff-def-first" in nonfun_diff and
+                    "diff-def-second" in nonfun_diff):
+                diff_def = (nonfun_diff["diff-def-first"],
+                            nonfun_diff["diff-def-second"])
             return cls(
                 nonfun_diff["kind"],
                 nonfun_diff["name"],
                 nonfun_diff["function"],
                 (nonfun_diff["stack-first"], nonfun_diff["stack-second"]),
-                (nonfun_diff["body-first"], nonfun_diff["body-second"])
+                (nonfun_diff["body-first"], nonfun_diff["body-second"]),
+                diff_def
             )
 
     class TypeDiff(NonFunDiff):
--- a/diffkemp/semdiff/caching.py
+++ b/diffkemp/semdiff/caching.py
@@ -209,10 +209,25 @@ class ComparisonGraph:
     class SyntaxDiff(NonFunDiff):
         """
         A syntax difference.
-
-        Note: body is a tuple containing the values for both modules.
+        :param kind: The kind of difference: 'macro', 'assembly',
+            'function-macro', 'macro-function', 'unknown'.
+        :param name: Name of an 'object' (eg. macro) in the first module in
+            which was found difference. In case of inline assembly difference
+            the name starts with "assembly code " followed by a number.
+        :param parent_fun: Name of vertex in which was the syntax diff found.
+        :param callstack: Tuple (call stack in first module, second one),
+            both call stacks are list containing calls, where each
+            call is dict with keys `file`, `line`, `weak`, `function`.
+            The `function` contains the name of called 'object',
+            in the case of macro it has extension " (macro)".
+            If the call is called from another macro then `line` does not
+            represent the line at which is the call called
+            but the line where the definition of the caller starts.
+        :param body: A tuple containing the body of the 'object' for both
+            modules.
         """
-        def __init__(self, name, parent_fun, callstack, body):
+        def __init__(self, kind, name, parent_fun, callstack, body):
+            self.kind = kind
             self.name = name
             self.parent_fun = parent_fun
             self.callstack = callstack
@@ -222,6 +237,7 @@ class ComparisonGraph:
         def from_yaml(cls, nonfun_diff):
             """Generates a SyntaxDiff object from YAML returned by SimpLL."""
             return cls(
+                nonfun_diff["kind"],
                 nonfun_diff["name"],
                 nonfun_diff["function"],
                 (nonfun_diff["stack-first"], nonfun_diff["stack-second"]),
--- a/diffkemp/simpll/Result.h
+++ b/diffkemp/simpll/Result.h
@@ -127,6 +127,10 @@ struct SyntaxDifference : public NonFunctionDifference {
     Kind syntaxKind;
     /// The difference.
     std::string BodyL, BodyR;
+    // Information about definitions of differing 'object'.
+    // For now it is only used in macro differences, for other differences it is
+    // null.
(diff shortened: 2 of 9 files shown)


diffkemp/simpll/Result.h
```diff
@@ -106,19 +120,37 @@ struct NonFunctionDifference {
...
+        FUNCTION_MACRO,
+        MACRO_FUNCTION,
+        UNKNOWN
+    };
+    SyntaxKind syntaxKind;
     /// The difference.
     std::string BodyL, BodyR;
-    SyntaxDifference() : NonFunctionDifference(SynDiff){};
-    SyntaxDifference(std::string name,
+    // Information about definitions of differing 'object'.
```
PLukas2018, line 136: For context instead of `differing` I used previously `last`
> I was a bit confused by the word "last" here. Maybe "differing" would be better (if it's always the case that the last object is the differing one)?

Yes, I think that the last macro should be always the differing one, but if I understand the implementation correctly the macro does not have to contain semantic difference but just syntactic one https://github.com/diffkemp/diffkemp/blob/e6cded9aefd2e51ada26c972f499fe08d442d789/diffkemp/simpll/SourceCodeUtils.cpp#L262-L263 , but we are still reporting it.

diffkemp/simpll/Result.h
```diff
@@ -106,19 +120,37 @@ struct NonFunctionDifference {
...
+    SyntaxDifference(SyntaxKind syntaxKind,
+                     std::string name,
                      std::string BodyL,
                      std::string BodyR,
                      CallStack StackL,
                      CallStack StackR,
-                     std::string function)
+                     std::string function,
+                     std::unique_ptr<Definition> diffDefL,
+                     std::unique_ptr<Definition> diffDefR)
```
PLukas2018, line 150: > Any reason we need std::unique_ptr here? Why not just plain `Definition`

I used `unique_ptr`, because I wanted it to be nullptr in case where there is no saved definition eg. for assembly difference or right now also for `macro-function`/`function-macro` differences.

tests/unit_tests/conftest.py
```diff
@@ -39,12 +39,13 @@ def graph():
     )
     # Non-function differences
     g["do_check"].nonfun_diffs.append(ComparisonGraph.SyntaxDiff(
-        "MACRO", "do_check",
+        "macro", "___MACRO", "do_check",
```
viktormalik, line 42: I'm not very familiar with this code but why did the `name` change from `MACRO` to `___MACRO`? Was it incorrect before?

diffkemp/simpll/Result.h
```diff
@@ -79,6 +79,20 @@ struct FunctionInfo {
     }
 };
 
+/// Type for storing information about definition of an analysed
+// 'object' (eg. macro).
+struct Definition {
+    // Name of the 'object'.
+    std::string name;
+    // Line in the sourceFile on which is located definition of the 'object'.
+    int line;
```
viktormalik, line 88: Could be `unsigned`, right?

diffkemp/simpll/Result.h
```diff
@@ -79,6 +79,20 @@ struct FunctionInfo {
     }
 };
 
+/// Type for storing information about definition of an analysed
+// 'object' (eg. macro).
+struct Definition {
```
viktormalik, line 84: This class is very similar to `CallInfo` (which also contains a name and a location) so I'm wondering if it makes sense to eventually merge them (something like `CodeLocation`). I'm not saying that we should do it now but it's something to think about.

viktormalik (commented):
LGTM! Just a couple of nits and questions to clear up some things.

diffkemp/semdiff/caching.py
```diff
@@ -209,10 +209,25 @@ def __init__(self, name, parent_fun, callstack):
     class SyntaxDiff(NonFunDiff):
         """
         A syntax difference.
-
-        Note: body is a tuple containing the values for both modules.
+        :param kind: The kind of difference: 'macro', 'assembly',
+            'function-macro', 'macro-function', ('unknown').
```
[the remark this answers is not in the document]
PLukas2018: I removed the parentheses (I think I put it there to indicate that it is not a kind of difference).

diffkemp/simpll/Result.h
```diff
@@ -106,19 +106,30 @@ struct NonFunctionDifference {
...
-/// Note: this can be either a macro difference or inline assembly difference.
+/// There are multiple kinds of the differences: a macro diff,
+/// inline assembly diff, function-macro diff or macro-function diff.
 struct SyntaxDifference : public NonFunctionDifference {
+    enum SyntaxKind {
+        MACRO,
+        ASSEMBLY,
+        FUNCTION_MACRO,
+        MACRO_FUNCTION,
+        UNKNOWN
```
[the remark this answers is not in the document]
PLukas2018: Thanks, that's clever.

diffkemp/simpll/Result.h
```diff
@@ -106,19 +120,37 @@ struct NonFunctionDifference {
 
 /// Syntactic difference between objects that cannot be found in the original
 /// source files.
-/// Note: this can be either a macro difference or inline assembly difference.
+/// There are multiple kinds of the differences: a macro diff,
+/// inline assembly diff, function-macro diff or macro-function diff.
 struct SyntaxDifference : public NonFunctionDifference {
+    enum SyntaxKind {
```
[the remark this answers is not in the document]
PLukas2018: Thanks, I renamed it, but for the `syntaxKind` attribute I leave the name same, so it is not confused with the `kind` of parent structure https://github.com/diffkemp/diffkemp/blob/e6cded9aefd2e51ada26c972f499fe08d442d789/diffkemp/simpll/Result.h#L142

tests/unit_tests/conftest.py
PLukas2018: The tests should probably work, but I think it was not correct - as `name` of SyntaxDifference is in Simpll used name of the last/differing macro - in this test, it was not the case.

diffkemp/simpll/Result.h
PLukas2018: Thanks, makes more sense for line to be `unsigned`, I changed it (originally in `MacroDef` it was `int` - but the number that is saved in it is `unsigned` https://github.com/diffkemp/diffkemp/blob/e6cded9aefd2e51ada26c972f499fe08d442d789/diffkemp/simpll/SourceCodeUtils.cpp#L370).

diffkemp/simpll/Result.h
PLukas2018: There is also [`FunctionInfo`](https://github.com/diffkemp/diffkemp/blob/e6cded9aefd2e51ada26c972f499fe08d442d789/diffkemp/simpll/Result.h#L58) which also contains name and location and other things. So I am thinking that I could rename in this PR the `Definition` struct to `CodeLocation`  and make another PR where I could refactor the `CallInfo` and `FunctionInfo` structs so they extends the `CodeLocation`, if that makes sense to you?

viktormalik: Yes, that makes sense. `CodeLocation` sounds like something that could be reused in multiple places (definitions, call sites, etc.).

diffkemp/simpll/Result.h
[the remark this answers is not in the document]
viktormalik: Makes sense, thanks.

diffkemp/simpll/Result.h
PLukas2018: It is renamed.

viktormalik (approved):
LGTM!

diffkemp/simpll/Result.h
PLukas2018: > ... make another PR where I could refactor the `CallInfo` and `FunctionInfo` structs so they extends the `CodeLocation` ...

Implemented in #348

PR #336, opened 2024-05-15
Release 0.6.0

viktormalik:
Changelog:
- Allow creating snapshots from a single C source file
- Add extended analysis statistics
- New pattern for matching reordered binary operations
- New pattern for variable grouping
- Nix flakes support for building and development
- Unification of applied optimizations for various build commands
- Fix comparison of PHI nodes and union types

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_policy(SET CMP0048 NEW)
 cmake_minimum_required(VERSION 3.5)
 project(diffkemp
-    VERSION 0.5.0)
+    VERSION 0.6.0)
 
 find_package(LLVM REQUIRED CONFIG)
 message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
--- a/flake.nix
+++ b/flake.nix
@@ -17,7 +17,7 @@
           with pkgs;
           python3Packages.buildPythonPackage {
             pname = "diffkemp";
-            version = "0.5.0";
+            version = "0.6.0";
 
             src = self;
 
--- a/rpm/diffkemp.spec
+++ b/rpm/diffkemp.spec
@@ -1,5 +1,5 @@
 Name:           diffkemp
-Version:        0.5.0
+Version:        0.6.0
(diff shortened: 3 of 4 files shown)


viktormalik:
Unfortunately, we still don't support LLVM 17 so we won't be able to build this release to newer Fedora versions. Still, I'd do the release since there's a lot of new changes coming. We may still do a feature release once we add the LLVM support.

viktormalik:
> The version number should probably be changed also here
> 
> https://github.com/diffkemp/diffkemp/blob/e6cded9aefd2e51ada26c972f499fe08d442d789/flake.nix#L20
> 

Nice catch! Fixed.

> , otherwise I think we can do the release.
> Originally, I encountered two problems when I tried the current (respectively previous) RPM package on Fedora, but after further investigation it should be fine:
> [...]

If you wouldn't mind, I'll ping you when the new RPMs are ready and you could test if the above issues are resolved.

PLukas2018 (approved):

PLukas2018:
> If you wouldn't mind, I'll ping you when the new RPMs are ready and you could test if the above issues are resolved.

Sure, I can test it.

viktormalik:
@PLukas2018 here's the list of artifacts from the newest RPM scratch build (for Fedora 39, i.e. LLVM 17):

https://download.copr.fedorainfracloud.org/results/viktormalik/diffkemp/fedora-39-x86_64/07475809-diffkemp/

If you could try to download the RPM, install it, and check if the errors that you previously encountered disappeared, that would be great. I had to do some changes to the specfile, see #339 (these are already included in the above scratch build).

Thanks!

PLukas2018:
> If you could try to download the RPM, install it, and check if the errors that you previously encountered disappeared, that would be great. I had to do some changes to the specfile, see #339 (these are already included in the above scratch build).

So I tried it (I tried `build`, `build-kernel`, `view` commands) and did not find any problem, everything worked well.

viktormalik:
Thanks! v0.6.0 is out:

https://github.com/diffkemp/diffkemp/releases/tag/v0.6.0

PR #339, opened 2024-05-22
RPM fixes for version 0.6.0

### LINKED ISSUE #334: Missing pip requirement in RPM package
I tried to install DiffKemp on Fedora (in docker) from the RPM package and it looks like that the specification is missing `pip` as a requirement.

I used the following commands for installation:

```bash
dnf install -y dnf-plugins-core
dnf copr enable -y viktormalik/diffkemp
dnf install -y diffkemp
```

The installation is successful but when I run `diffkemp -h` error is thrown:
```bash
Traceback (most recent call last):
  File "/usr/bin/diffkemp", line 3, in <module>
    from diffkemp import diffkemp
  File "/usr/lib64/python3.11/site-packages/diffkemp/diffkemp.py", line 3, in <module>
    from diffkemp.config import Config
  File "/usr/lib64/python3.11/site-packages/diffkemp/config.py", line 3, in <module>
    from diffkemp.snapshot import Snapshot
  File "/usr/lib64/python3.11/site-packages/diffkemp/snapshot.py", line 14, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

After installing pip with `dnf install -y pip` it works.

viktormalik:
Several RPM fixes were necessary when building version 0.6.0:
- Replace `pathfix.py ... FILE` by `%py3_shebang_fix FILE` as recommended by a [Fedora change](https://bugzilla.redhat.com/show_bug.cgi?id=1868347).
- Add `BuildRequires: gtest-devel` as we're not vendoring GoogleTest by default.
- Add `Requires: python3-setuptools` to fix #334.

### DIFF
--- a/rpm/diffkemp.spec
+++ b/rpm/diffkemp.spec
@@ -13,7 +13,6 @@ BuildRequires:  gcc gcc-c++ cmake ninja-build
 BuildRequires:  llvm-devel
 BuildRequires:  python3-devel python3-pip python3-setuptools python2
 BuildRequires:  git
-BuildRequires:  /usr/bin/pathfix.py
 BuildRequires:  npm
 Requires:       cscope
 Requires:       clang llvm-devel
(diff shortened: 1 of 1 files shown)


FrNecas (approved):
Looks good to me

PR #340, opened 2024-05-22
Extraction of macro definitions from callstacks

PLukas2018:
Second part of commits from https://github.com/diffkemp/diffkemp/pull/314 .
This PR adds:
- method for extracting macro definitions from the call stacks (it is possible to extract information about definitions for all macros except the differing macros for which is the definition extracted in [#333](https://github.com/diffkemp/diffkemp/pull/333) PR)
- update of `get_end_line` function to be possible to use it also for finding line where macro ends.

### DIFF
--- a/diffkemp/semdiff/result.py
+++ b/diffkemp/semdiff/result.py
@@ -118,6 +118,47 @@ class Result:
                     type_names.add((name, parent_name))
             return function_names, macro_names, type_names
 
+        def get_macro_defs(self, compared_fun):
+            """
+            Returns a tuple:
+            - list of macro definitions (except last macro in the callstack)
+              = list of dicts (`file`, `name`, `line`),
+            - (vertex name, last macro name) for retrieving info
+              about last macro defintion from the ComparisonGraph.
+            In case the call stack does not includes macros, returns ([], None)
+            :param compared_fun: Name of compared function.
+            """
+            macro_defs = []
+            # For getting last macro deffinition
+            last_macro_name = None
+            last_function_name = compared_fun
+            if not self.calls:
+                return macro_defs, None
+            for index, call in enumerate(self.calls):
+                name, kind = self.split_name_and_kind(call)
+                if kind != "macro":
+                    # Note: In case the kind is type then there shouldn't
+                    # be macros so it doesn't matter if the var is set.
+                    last_function_name = name
+                    continue
+                # Last call in the callstack
+                if index + 1 == len(self.calls):
+                    last_macro_name = name
+                else:
+                    # The call contains a macro name, a file from which
+                    # the macro was used and a line where starts the macro
+                    # body/def from which it was used.
+                    # To get information about a current macro file and
+                    # line, we have to look on a call below.
+                    macro_defs.append({
+                        "name": call["name"],
+                        "file": self.calls[index + 1]["file"],
+                        "line": self.calls[index + 1]["line"]
+                    })
+            last_macro_info = (last_function_name, last_macro_name) \
+                if last_macro_name is not None else None
+            return macro_defs, last_macro_info
+
         def __str__(self):
             """Converts a callstack to a string representation."""
             if self.calls is None:
--- a/diffkemp/semdiff/result.py
+++ b/diffkemp/semdiff/result.py
@@ -81,6 +81,15 @@ class Result:
                      "line": call["line"]}
                     for call in self.calls]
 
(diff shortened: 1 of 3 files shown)


diffkemp/semdiff/result.py
```diff
@@ -81,6 +81,17 @@ def to_output_yaml_with_rel_path(self, prefix):
...
+        @staticmethod
+        def get_name_and_kind(call):
+            """Returns tuple (name, kind) for a call."""
+            name = call["name"]
+            kind = "function"
+            if " " in name:
+                name, kind = name.split(" ")
+                # kind looks like this "(kind)", extracting it from brackets
+                kind = kind[1:-1]
+            return (name, kind)
```
viktormalik, line 93: What about this? Slightly shorter and more readable (at least for me)
```suggestion
            split = call["name"].split()
            if len(split) == 1:
                return (split[0], "function")
            # The expected form is "name (kind)"
            return (split[0], split[1][1:-1])
```
Also note that whitespace is the default separator for `split`.

diffkemp/utils.py
```diff
@@ -53,25 +53,29 @@ class EndLineNotFound(Exception):
...
-        # The end of the function is detected as a line that contains
-        # nothing but an ending curly bracket
         line_number = start
         line = lines[line_number - 1]
-        while line.rstrip() not in terminator_list:
+        lines_amount = len(lines)
+        while not is_end(line):
             line_number += 1
-            if line_number > len(lines):
+            if line_number > lines_amount:
```
viktormalik, line 78: Why not use `len(lines)` directly? It's a constant function so there's no performance lost.

diffkemp/semdiff/result.py
```diff
@@ -120,6 +120,46 @@ def get_symbol_names(self, compared_fun):
                     type_names.add((name, parent_name))
             return function_names, macro_names, type_names
 
+        def get_macro_defs(self, compared_fun):
+            """
+            Returns a tuple:
+            - list of macro definitions (except last macro in the callstack)
```
viktormalik, line 126: I'm wondering if we should call this "macro locations" rather than "macro definitions". The latter implies to me that we're extracting macro bodies while in fact, we're just extracting locations where the macros start. OTOH, if we're using the word "definition" in this context for functions, then it's probably better to stay consistent.

diffkemp/semdiff/result.py
```diff
@@ -120,6 +120,46 @@ def get_symbol_names(self, compared_fun):
...
+            - (vertex name, last macro name) for retrieving info
+              about last macro defintion from the ComparisonGraph.
+            In case the call stack does not includes macros, returns ([], None)
+            :param compared_fun: Name of compared function.
+            """
+            macro_defs = []
+            # For getting last macro deffinition
+            last_macro_name = None
+            last_function_name = compared_fun
+            if self.calls:
```
viktormalik, line 137: I'd invert this to save some indentation.

viktormalik (commented):
LGTM, just a couple of code styling remarks (feel free to ignore those that you don't like).

Also one more thing: could you try to add a short description to each commit? Even if the commit is simple (like a refactoring), it's always good to explain what's done there as it gives the author's intent and therefore makes the commit easier to understand.

For instance, for your commits no.1 and 3, there could be something like:
```
Result Callstack refactoring

Extract splitting symbol name from a symbol kind to a separate function.
Also invert a condition to decrease indentation.
```
```
Add getting end line for macros

The end line of a macro definition is the line which does not end with
an escape backslash ('\').
```

diffkemp/utils.py
PLukas2018: Thanks, I did not know that.

diffkemp/semdiff/result.py
PLukas2018: Good point, we are already using the word 'definition' eg. here https://github.com/diffkemp/diffkemp/blob/4b94c9acd38909fc5b934dabde188bd15135e05b/diffkemp/output.py#L76-L79 , so I would probably stick with the word 'definition'.

PLukas2018:
Thanks for the review, I resolved the comments.

Also, I fixed a mistake that I made in `get_end_line` function - because `readlines()` returns a newline character it is necessary always to use `rstrip()` (even for checking end line in macros). 
I also added unit tests for the `get_end_line` function.

viktormalik (approved):
Nice!

PR #342, opened 2024-05-27
build-command: Ensure that the symbol list exists

PLukas2018:
When using `build` command, user can specify file containing names of symbols/functions which he wants to compare.
The command  firstly compiles the project using the CC-wrapper and after that the list of symbols (functions) which should be compared is extracted from the specified file.
So in case the user misstypes name of the file it can takes several minutes than he founds out about the error and he has to run the command again.

A check if the file is valid is added before compiling a project to avoid potentional delays.

### DIFF
--- a/diffkemp/diffkemp.py
+++ b/diffkemp/diffkemp.py
@@ -28,6 +28,8 @@ import yaml
 VIEW_INSTALL_DIR = "/var/lib/diffkemp/view"
 # Name of YAML output file created by diffkemp compare command.
 YAML_FILE_NAME = "diffkemp-out.yaml"
+# Error message shown when no symbols were found in the symbol list file.
+EMSG_EMPTY_SYMBOL_LIST = "ERROR: symbol list is empty or could not be read\n"
 
 
 def build(args):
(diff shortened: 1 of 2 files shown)


diffkemp/diffkemp.py
```diff
@@ -29,6 +29,12 @@
 
 
 def build(args):
+    # If the symbol list is specified, check that the file exists.
+    if args.symbol_list is not None and not os.path.isfile(args.symbol_list):
+        sys.stderr.write(
+            "Error: specified symbol_list file does not exist\n"
```
viktormalik, line 35: Nice catch! Could we use the same error message that `build_kernel` uses when the symbol list file doesn't exist (or is empty)?

PLukas2018: So I found out that neither of the build commands (`build-kernel`, `build`, `llvm-to-snapshot`) were checking if the specified `symbol_list` file exist, so I added the check to all commands. Also I added a check to the `build` command if it was possible to extract some symbols from the `symbol_list` / project itself - more info in the commit messages.

viktormalik (approved):
LGTM, thanks!

PR #345, opened 2024-06-06
Viewer: preparation of macro visualisation

PLukas2018:
This PR contains the next set of commits from #314. It prepares the viewer for visualisation of macros, it enables to show code of macros that are located only on one side of a call stack.

The main changes are:
- Previously, we copied only the sources of old version of the compared program,. This PR adds copying of the new project version sources to the folder accessible by the viewer.
- Updates in the `Callstack` component to indicate whether the selected call from the call stack is located in the left, right, or both call stacks. This helps to show the correct version of the code (code for old/new/both versions).
- Enhancements of `DiffViewWrapper` component (component for rendering the code) so it to allow it to show code for only old/new version.
- Because from call stack information we are unable to deduce for macros on which line is used the next macro (lines are present in the call stack but they do not represent line where the macro is used but line where the parent macro definition starts),  the last 3 commits ensure that the entire macro definition is shown without highlighting the incorrect line.

This PR only prepares the visualisation but does not add macro definitions to `diffkemp-out.yaml` (created by `compare` command). So the viewer is able to show the macros, but currently, it does not have the necessary information for that (the macro definitions). I will open another PR which will add the macro definitions to `diffkemp-out.yaml after this PR is merged.
If this PR would be too much to review I could split it into more PRs based on commits.

Screenshots of code visualisation of macros:
- Same macro in both call stacks:
  image
- Macro located only in left/old call stack:
  image
- Macro located only in right/new call stack:
  image
- Macro in which was found difference:
  image

### DIFF
--- a/diffkemp/diffkemp.py
+++ b/diffkemp/diffkemp.py
@@ -655,7 +655,8 @@ class Viewer:
         self.compare_output_dir = compare_output_dir
         self.devel = devel
         # Relative path of src files which were already copied to viewer dir.
-        self.processed_files = set()
+        self.old_processed_files = set()
+        self.new_processed_files = set()
         # Path to snapshot directories.
         self.old_snapshot_dir = None
         self.new_snapshot_dir = None
@@ -668,7 +669,8 @@ class Viewer:
         # Path to the viewer dir containing data of the compared project.
         self.data_dir = None
         # Path to the viewer dir containing sources of the compared project.
-        self.source_dir = None
+        self.old_out_dir = None
+        self.new_out_dir = None
         # Path to the viewer dir containing diff of functions.
         self.diff_dir = None
         # Path to the yaml (describing the result) in the viewer directory.
@@ -732,10 +734,13 @@ class Viewer:
         self.data_dir = os.path.join(self.public_dir, "data")
         if not os.path.exists(self.data_dir):
             os.mkdir(self.data_dir)
-        # Preparing source directory
-        self.source_dir = os.path.join(self.data_dir, "src")
-        if not os.path.exists(self.source_dir):
-            os.mkdir(self.source_dir)
+        # Preparing source directories
+        self.old_out_dir = os.path.join(self.data_dir, "src-old")
+        self.new_out_dir = os.path.join(self.data_dir, "src-new")
+        if not os.path.exists(self.old_out_dir):
+            os.mkdir(self.old_out_dir)
+        if not os.path.exists(self.new_out_dir):
+            os.mkdir(self.new_out_dir)
         # Preparing diff directory
         self.diff_dir = os.path.join(self.data_dir, "diffs")
         if not os.path.exists(self.diff_dir):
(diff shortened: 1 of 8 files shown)


view/src/components/Difference.jsx
```diff
@@ -1,15 +1,15 @@
...
+import { useMemo, useState } from 'react';
 import { PropTypes } from 'prop-types';
 
 import Container from 'react-bootstrap/Container';
 import Col from 'react-bootstrap/Col';
 import Row from 'react-bootstrap/Row';
 import Alert from 'react-bootstrap/Alert';
 
-import Callstack from './Callstack';
+import Callstack, { CallSide } from './Callstack';
```
PLukas2018, line 12: Relevant comment from #314:
> Shouldn't be?
> ```suggestion
> import { Callstack, CallSide } from './Callstack';
>   ```

I do not think so, the component `CallStack` is exported as default export, therefore it is outside of the braces, but `CallSide` is exported as named export, therefore it is in the braces ([import documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#syntax)). I could change the way how it is exported, but I exported the `CallStack` as default because it is the main thing from the file, furthermore I think I saw somewhere this approach - eg. [here](https://create-react-app.dev/docs/importing-a-component/), where in old (with class approach) React you did something like  `import React, { Component } from 'react';`)

view/src/components/DiffViewWrapper.jsx
```diff
@@ -87,39 +135,73 @@ function getSyntaxHighlightedTokens(hunks) {
...
-      return fileObject;
+      returnValue = fileObject;
+    } else {
+      // Code does not contain syntax differences - creating the object using first line of code.
+      returnValue = createFileStructure(oldCode, oldStart, newStart);
     }
-    return createFileStructure(oldCode, oldStart, newStart);
-  }, [diff, oldCode, oldStart, newStart]);
-    /* hunks of differences/code which will be shown */
+    return returnValue;
```
DanielKriz, line 195: Perhaps it is because I use different style, but wouldn't multiple return statements be more readable then the returnValue? It is a small anonymous function, so your intent is clear.

view/src/components/Difference.jsx
DanielKriz: I think that it is all right, but I do not have much experience with clean javascript code. But your point makes sense, especially if you are using it elsewhere in the project.

view/src/components/DiffViewWrapper.jsx
PLukas2018: Thanks for the feedback. I originally wanted to write it as you mentioned, but the linter that we are using marked it as [no-else-return](https://eslint.org/docs/latest/rules/no-else-return) error, so this was my approach to overcome it. I changed it and now I am using `if` instead of `else if`, the linter is OK with this and hopefully it will be a little bit more readable.

view/src/components/DiffViewWrapper.jsx
```diff
@@ -152,27 +231,31 @@ export default function DiffViewWrapper({
...
-        const oldLineOfNewCalling = getCorrespondingOldLineNumber(
-          file.hunks,
-          linesToShow[1],
-        );
-        if (oldLineOfNewCalling !== linesToShow[0]) {
-          addLineWithContext(oldLineOfNewCalling);
-        }
-      } else {
-        addAllLinesOfFunction();
+    if (showDiff) return;
```
viktormalik, line 238: Why is this here?

view/src/tests/DiffViewWrapper.test.jsx
```diff
@@ -321,4 +323,106 @@ describe('testing a caller function visualisation', () => {
...
+    );
+
+    const callstack = within(screen.getByTestId('callstack'));
+
+    // Selecting function which is only in the old call stack.
+    userEvent.click(callstack.getByTitle('left (macro)'));
+    // Waiting for update.
+    await screen.findByText(/left_file.c/);
+    // Testing that the whole code of the function is shown
+    // and it is shown on the left side and the right side is empty.
```
viktormalik, line 393: The latter part ("right side is empty") is not tested, is it?

viktormalik (commented):
Looks good! Just a couple of comments. I'm still a bit lost in the JavaScript code but the changes seem to be fine.

diffkemp/diffkemp.py
```diff
@@ -668,7 +669,8 @@ def __init__(self, compare_output_dir, devel):
         # Path to the viewer dir containing data of the compared project.
         self.data_dir = None
         # Path to the viewer dir containing sources of the compared project.
-        self.source_dir = None
+        self.out_old_dir = None
+        self.out_new_dir = None
```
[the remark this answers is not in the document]
PLukas2018: Thanks, good point, for now I changed it to a prefix.

By a quick look at the codebase, I found out that we are mostly prefixing with `old`/`new`. The only inconsistency I discovered is with the naming of the snapshot directory (`snapshot_dir_old`/`snapshot_dir_new`) where we use suffixes instead. 

There is also inconsistency caused by using different terms for the same thing: `old`/`first`/`left`/`L` and `new`/`second`/`right`/`R` and prefixing/suffixing connected with it. Unifying this would require big refactoring as you mentioned.

diffkemp/diffkemp.py
```diff
@@ -732,10 +734,13 @@ def _prepare_dirs(self):
         self.data_dir = os.path.join(self.public_dir, "data")
         if not os.path.exists(self.data_dir):
             os.mkdir(self.data_dir)
-        # Preparing source directory
-        self.source_dir = os.path.join(self.data_dir, "src")
-        if not os.path.exists(self.source_dir):
-            os.mkdir(self.source_dir)
+        # Preparing source directories
+        self.out_old_dir = os.path.join(self.data_dir, "old-src")
+        self.out_new_dir = os.path.join(self.data_dir, "new-src")
```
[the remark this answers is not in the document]
PLukas2018: I do not have preferences. I changed it to `src-old`/`src-new` as you suggested, I feel like it can make more sense.

view/src/components/Code.jsx
```diff
@@ -31,7 +36,9 @@ const DIFF_DIRECTORY = 'diffs';
  *                                              of the function ends.
  * @param {Number} [props.specification.calling] - Tuple (old, new) of line numbers where are called
  *                                                 next functions which should be highlighted
- *                                                 or undefined if it is differing function.
+ *                                                 or undefined if no functions are called.
+ * @param {boolean} props.specification.differing - True if the symbol is the differing one
+ *   and the syntax difference should be highlighted.
```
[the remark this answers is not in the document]
PLukas2018: Thanks, I have fixed it, so the text description starts at the same column.

To be honest, I have been inconsistent with the comment indentation in the viewer. I think I started by indenting the next line so the description starts at the same column. Later, I thought it could waste a lot of space (lines) if the description is long, so for newly added parameters I continued the description at the beginning of the line and sometimes I probably indented the next line by two spaces. I'm not really sure what is the best approach.

view/src/tests/DiffViewWrapper.test.jsx
PLukas2018: Good question. It was actually tested by providing the empty string to `testCodeNumberMatch` function
https://github.com/PLukas2018/diffkemp/blob/8c561152923644fbfb537a1693bc0f59197eaf9e/view/src/tests/DiffViewWrapper.test.jsx#L393-L397
, which would fail here
https://github.com/PLukas2018/diffkemp/blob/8c561152923644fbfb537a1693bc0f59197eaf9e/view/src/tests/DiffViewWrapper.test.jsx#L54
if the side was not empty (if a line would be shown).

I have felt like the `testCodeNumberMatch` function was not clearly documented so I added a commit that adds comments to make it hopefully more clear, also I simplified one thing by using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_function_calls).

view/src/components/DiffViewWrapper.jsx
PLukas2018: Good question. This line was added in `Viewer: Show entire definition for macros` commit,  I realize now that the commit message did not fully describe the change I made.  The intention was to keep the behavior for macros containing differences unchanged -- by default, showing only the lines with syntax differences (+ context lines).
If there is a difference, the lines with syntax differences are already shown (respectively included in `hunks` to be shown later), the  lines below `    if (showDiff) return;` determine other lines of source code which should be shown. In case of macros all lines would be added. So by this line, I am ending the function early to ensure that we show only the differing lines and their context.

I updated the commit message so it includes the information that for differing macros we are by default showing only the syntax differences. Also, I added a comment above the line, so it is more clear why it is there.

view/src/components/Code.jsx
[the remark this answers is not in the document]
viktormalik: Yeah, me neither. Let's keep it as it is for now.

view/src/components/DiffViewWrapper.jsx
viktormalik: Ah, ok, makes sense now. Thanks!

viktormalik (approved):
Looks good!

PR #346, opened 2024-06-09
README split, edits, examples

PLukas2018:
The changes made in this PR can be previewed on this [link](https://github.com/PLukas2018/diffkemp/blob/docs/README.md).

In the PR the original README is split into multiple files:
- `docs/about.md` containing information about DiffKemp, implementation details and links to publication,
- `docs/development.md` containing information for developing DiffKemp (environments, build, tests, coding style, useful links, ...)
- `docs/installation.md` containing information about how to install DiffKemp from source (dependencies, building, installation),
- `docs/usage.md` containing more detailed info about individual commands and in the future could contain description of options and output for the commands.

The main README is shortened and contains links to the other files.

In some places, I also added some things (image capturing the relation between commands, examples of dependencies installation for ubuntu/fedora, commands for installing from Copr, links to tools and documentations, info about coding style, ...). The changes are more thoroughly described in individual commit messages.

I also added 3 examples of using DiffKemp:
- Example of simple program: Contains an example of using DiffKemp on a file that contains two functions. It mentions the difference between syntax and semantic difference, shows how to use `build` and `compare` command, explains the output of the `compare` command and informs that DiffKemp evaluates functions including the called functions.
- Example of using DiffKemp on `musl` library: Explains the structure of the `compare`'s command output directory and shows how to use the viewer (contains a few screenshots of it).
- Example of using DiffKemp for the kernel: Mentions the `rhel-kernel-get` script and the usage of `build-kernel` command.

In the last commit, I tried to add a motivation for using DiffKemp, I am not really sure if what I wrote is good so I am thinking of removing it. If you have some suggestions what could be placed there instead, let me know.


README.md
```diff
@@ -10,150 +10,114 @@ The main use-case of DiffKemp is to compare selected functions and configuration
...
+## How does it work?
 
-Sysctl option comparison compares semantics of the proc handler function and
-semantics of all functions using the data variable that the sysctl option sets.
+The main focus of DiffKemp is high scalability, such that it can be applied to
+large-scale projects containing a lot of code. To achieve that, the analysed
+functions are first compiled into LLVM IR, then several code transformations are
+applied, and finally the comparison itself is performed.
+
+Wherever possible, DiffKemp tries to compare instruction-by-instruction (on LLVM
```
FrNecas, line 112: I'd use `at the level of LLVM IR` instead of `on LLVM IR instructions`

README.md
```diff
@@ -10,150 +10,114 @@ The main use-case of DiffKemp is to compare selected functions and configuration
...
+IR instructions) which is typically sufficient for most of the code. When not
+sufficient, DiffKemp tries to apply one of the built-in or user-supplied
+*semantics-preserving patterns*. If a semantic difference is still discovered,
```
FrNecas, line 115:
```suggestion
*semantics-preserving patterns*. If no pattern can be applied,
```

README.md
```diff
@@ -291,22 +145,26 @@ order:
...
+- Tomáš Kučma
 - Lukáš Petr
 - Tatiana Malecová
 - Jakub Rozek
 
 ## Publications and talks
 
 There is a number of publications and talks related to DiffKemp:
+
+- NETYS'22 talk
```
FrNecas, line 158: I'd suggest keeping NETYS at the bottom, even though it is the newest. Custom patterns are quite specific and advanced and I think it's better to have a general introduction (the ICST paper and devconf talk) at the top.

docs/development.md
```diff
@@ -0,0 +1,249 @@
...
+pip install -e . # In case of Nix use setuptoolsShellHook instead
+```
+
+- `-DBUILD_VIEWER=On`: This flag will install packages and build the result
+  viewer. It will try to install the packages on every `cmake` run, so you may
+  want to use `-DBUILD_VIEWER=Off` instead.
+
+If you make changes to the SimpLL library, you will need to rebuild it by
+running `ninja -C build`. We are using [CFFI](https://cffi.readthedocs.io/en/stable/)
+for accessing the library from the Python. The CFFI is not rebuild by default,
```
FrNecas, line 134:
```suggestion
for accessing the library from the Python. The CFFI is not rebuilt by default,
```

docs/development.md
```diff
@@ -0,0 +1,249 @@
...
+  so use option 1 instead.)
+
+## Coding style
+
+We require that the code is according to a certain coding style, which can be
+checked with the following tools:
+
+- For Python code: `flake8 diffkemp tests`.
+- For C++ code: `tools/check-clang-format.sh -d` (the coding style can be
+  automatically fixed by `tools/check-clang-format.sh -di`).
```
FrNecas, line 156: I'd mention that it's necessary to have `clang-format` installed.

docs/examples/linux_kernel.md
```diff
@@ -0,0 +1,99 @@
+# Comparing Linux kernels
+
+The main focus of DiffKemp is on Red Hat Enterprise Linux (RHEL), whose kernel
+includes a list of functions (kABI symbols) whose semantics should remain the
```
FrNecas, line 4:
```suggestion
includes a list of functions (KABI -- Kernel Application Binary Interface) whose semantics should remain the
```

docs/examples/linux_kernel.md
```diff
@@ -0,0 +1,99 @@
...
+same across the lifetime of a single major release. In this example, we will
+show how to compare two versions of the kernel. If you want to try this on your
+own, be aware that it can take a few hours to get the results.
```
FrNecas, line 7: Hours, really? :D Doesn't the comparison take like 5 minutes?

docs/examples/linux_kernel.md
```diff
@@ -0,0 +1,99 @@
...
+mkdir -p snapshots
+diffkemp build-kernel kernel/linux-4.18.0-193.el8/ snapshots/linux-4.18.0-193.el8/ kernel/linux-4.18.0-193.el8/kabi_whitelist_x86_64
+diffkemp build-kernel kernel/linux-4.18.0-240.el8/ snapshots/linux-4.18.0-240.el8/ kernel/linux-4.18.0-240.el8/kabi_whitelist_x86_64
+```
+
+The `kernel/linux-4.18.0-XXX.el8/` directories are directories with
+the downloaded and prepared 8.2 and 8.3 kernels,
+the `snapshots/linux-4.18.0-XXX.el8/` are directories where the created
+snapshots will be saved and `kerne/linux-4.18.0-XXX.el8/kabi_whitelist_x86_64`
+files are files that contain the list of kABI symbols.
```
FrNecas, line 44: I think that kABI should be upper-case since it's an abbreviation.

docs/examples/linux_kernel.md
```diff
@@ -0,0 +1,99 @@
...
+
+When running the command, we are informed which symbols were already found
+and compiled into LLVM IR (LLVM IR is used for the comparison). For some
+symbols we can get `source not found` note, this is normal, so do not worry.
+
+> [!NOTE]
+> If you get `source not found` note for all symbols, there is a problem.
+> It could be caused by cancelling the `build-kernel` command prematurely
+> which could result in the `cscope` cross-referencer, used by DiffKemp, not
+> being build for all symbols. Removing the `kernel/linux-4.18.0-XXX.el8/cscope.files`
```
FrNecas, line 54:
```suggestion
> being run for all symbols. Removing the `kernel/linux-4.18.0-XXX.el8/cscope.files`
```

docs/examples/simple_program.md
```diff
@@ -0,0 +1,425 @@
...
+diffkemp compare --stdout old-snapshot new-snapshot
+```
+
+Wait, there is no output?! That's right, similar to the classic diff utility,
+DiffKemp `compare` command does not output anything if it evaluates the programs
+(individual functions) as equal. Even if there is a syntactic difference
+which we could see using the diff utility, the semantic of the functions
+remained the same.
+
+## 4. Making semantic difference
```
FrNecas, line 154:
```suggestion
## 4. Making a semantic difference
```

docs/examples/simple_program.md
```diff
@@ -0,0 +1,425 @@
...
+
+```sh
+diffkemp build src/old.c old-snapshot
+diffkemp build src/new.c new-snapshot
+```
+
+### Comparing the program and interpreting the result
+
+Now when we have created the snapshots of both programs, we can compare
+the programs (respectively the created snapshots). For this, we will use
```
FrNecas, line 138:
```suggestion
the programs (or more precisely, the created snapshots). For this, we will use
```

docs/examples/simple_program.md
```diff
@@ -0,0 +1,425 @@
...
+We have skipped one part of the output which we did not see previously.
+The part contains call stacks, which is a list of called functions from the
+compared function (`mul`) to the differing function (`add`). For each function,
+there is information about in which file was the function called and on which
+line. We can see that the function `add` was called in the old version in the
+`old.c` file on line `11` and in the new version in the file `new.c` on
+line `9`. Since there are no other functions in the call stack, it indicates
+that `add` was directly called from `mul`.
+
+To explain more the call stacks, consider a program structure like this:
```
FrNecas, line 363:
```suggestion
To explain the call stacks more, consider a program structure like this:
```

docs/examples/musl_library.md
```diff
@@ -0,0 +1,250 @@
...
+diffkemp build musl-1.2.3 snapshot-musl-1.2.3 function_list
+diffkemp build musl-1.2.4 snapshot-musl-1.2.4 function_list
+```
+
+Building of the snapshots will take some time.
+
+## 3. Comparing two versions of the library
+
+After creating the snapshots, we can compare them. Because there are many
+functions, using the `--stdout` argument (printing the result to standard
```
FrNecas, line 78: Maybe we could mention `--report-stat` here since that allows to see at least some statistics about the analysis that could be useful?

docs/examples/musl_library.md
```diff
@@ -0,0 +1,250 @@
...
+output) isn't the best idea. We can use the `-o` option followed by the
+directory name to which the result will be saved:
+
+```sh
+diffkemp compare snapshot-musl-1.2.{3,4} -o diff-snapshot-musl-1.2.3-snapshot-musl-1.2.4
```
FrNecas, line 83: I am not a big fan of this `{3, 4}`. I think that a new user should be able to copy paste every command from the example without any change and have it run just fine. I know you probably did this to save space but I'd rather drop `-o` or replace it with `-o diff` or something short :)

docs/examples/musl_library.md
```diff
@@ -0,0 +1,250 @@
...
+fork.diff
+sem_post.diff
+```
+
+If there are no differences found then the directory is not even created.
+
+## 4. Going through the results
+
+In the directory, there are multiple files, the files with extension `.diff`
+represents compared functions in which were found differences. So in our case,
```
FrNecas, line 102:
```suggestion
represents compared functions in which differences were found. So in our case,
```

docs/examples/musl_library.md
```diff
@@ -0,0 +1,250 @@
...
+if we compared `sin`, `cos`, `open`, `sem_post`, `dlopen` and `fork`
+functions, we can see that the `dlopen`, `fork` and `sem_post` functions
+contain semantic differences according to DiffKemp. The files contain more
+details about the differences, let's look eg. in the `dlopen.diff` file:
```
FrNecas, line 106:
```suggestion
details about the differences, let's look e.g., in the `dlopen.diff` file:
```

FrNecas (commented):
Very nicely done, I really like the new structure of documentation! I have a couple of suggestions, a few grammar/spelling fixes but nothing major.

docs/examples/musl_library.md
PLukas2018: So the command is functional at least when using bash ([brace expansion](https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html)), not sure if the syntax is valid in other shells, so I will rather change it, thanks.

FrNecas: Oh cool, I didn't know about this feature. It works in zsh too but maybe some more minimal shells may not have this, I don't think it is a standard feature. And also I think that a lot of people may not know about this so it could cause confusion

README.md
```diff
@@ -10,150 +10,114 @@ The main use-case of DiffKemp is to compare selected functions and configuration
...
+  dnf install -y diffkemp
+  ```
 
-### Install from RPM
+## Usage
 
-Alternatively, you can use a prepared RPM package for Fedora that can be
-installed from our Copr repository:
-[https://copr.fedorainfracloud.org/coprs/viktormalik/diffkemp/](https://copr.fedorainfracloud.org/coprs/viktormalik/diffkemp/)
+![DiffKemp commands](docs/img/commands.svg)
```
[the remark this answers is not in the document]
PLukas2018: Thanks, I added the checkmark and cross. I am thinking if I should also add an arrow to the semantic comparison indicating that it is possible to specify custom change patterns for the comparison.

docs/examples/linux_kernel.md
PLukas2018: I removed the sentence. The sentence was not very well put, I meant the whole example with everything can take even a few hours -- downloading and preparing the kernels with `rhel-kernel-get` (2x ~50s--1.5min), running `diffkemp build` (2x ~35--50min), running `diffkemp compare`.

docs/examples/linux_kernel.md
PLukas2018: Sure, I changed it. Originally I wrote it like that because I saw it written like that [here](https://access.redhat.com/solutions/444773).

docs/examples/musl_library.md
PLukas2018: Thanks, I added it to the example and added mention about the `--extended-stat` option.

PLukas2018:
@FrNecas, thanks for the review, all comments should hopefully now be resolved.

viktormalik (commented):
I really like this refactoring, especially the examples are so great!

Many thanks for the work, there's a couple of remarks below but overall I think that this is an excellent improvement!

PLukas2018:
Working on resolving the comments and I looked at this:

> I'm wondering if we could use a better image format which would be text (or source-code) based. That way, we could easily extend it, see the history of changes, and review them. I'm thinking about Graphviz or TikZ here.
> 
> If not, SVG is IMO fine (I'll have some remarks to the image itself, though).

- For **TikZ** I did not find a way to render it in the GitHub markdown, so we would probably need to commit both the source code and the rendered image.
- For [**Graphviz**](https://graphviz.org/) exists [Gravizo](https://www.gravizo.com/) which allows to include graphs in markdown. There are two ways how to write it
  - Directly writing the graph description as part of Gravizo URL, the problem is that the description must be on one line (and probably some chars has to be urlencoded), because GitHub does not support anymore multiline URLs.
  - Indirectly - the description of the graph can be in a different file and the image is rendered by providing the URL to the file as part of Gravizo URL. The advantage I see in this is that the description can be in one file but we can render the image in multiple files. A disadvantage I see is in the URL -- the URL of the file containing the graph description will be different after it is merged to the master so the PR would need to contain URL on which the file with the graph will be accessible after the merge.

  I tried the indirect way ([file with description](https://github.com/PLukas2018/diffkemp/blob/docs-graphviz/docs/img/graphviz.md?plain=1), [file with rendered graph](https://github.com/PLukas2018/diffkemp/blob/1dc77c6009da8c79b9cce0304a913b9b4a5c5d79/README.md?plain=1#L36)) and tried to recreate the original image, the result looks like this:
  DiffKemp commands
- I also tried [Mermaid](https://mermaid.js.org/) which is mentioned in [Github docs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams#creating-mermaid-diagrams) ([my attempt](https://github.com/PLukas2018/diffkemp/blob/00aa9b12830e91b66d157d9cd62d7389303f5968/README.md?plain=1#L34-L84)). The description is written in fence code block and GitHub manages the rendering by itself.  My attempt of recreate of the origin image looks like this:
  ```mermaid
  flowchart LR
    %% comment
    pv(project versions)
    sl(SYMBOL LIST)
    subgraph sgsubgraph["build\nbuild-kernel\nllvm-to-snapshot"]
        direction TB
        tmp(" ")
        sg(Snapshot\ngeneration)
    end
    subgraph scsubgraph["compare"]
        sc(Semantic\ncomparison)
    end
    neq(✗ not equal)
    eq(✓ equal)
    tmp2(" ")
    pv -- OLD --> sg
    pv -- NEW --> sg
    sl --> sg
    sg -- OLD SNAPSHOT --> sc
    sg -- NEW SNAPSHOT --> sc
    sc --> eq ~~~ tmp2
    sc --> neq ~~~ tmp2
    subgraph rvsubgraph["view"]
        rv(Result viewer)
    end
    sc -. "report for not equal symbols" .-> rv
    linkStyle 2 stroke:darkblue
    linkStyle 3 stroke:darkblue,color:darkblue;
    linkStyle 4 stroke:darkblue,color:darkblue;
    linkStyle 5 stroke:darkgreen;
    linkStyle 6 style:invis;
    linkStyle 7 stroke:darkred;
    linkStyle 8 style:invis;
    linkStyle 9 stroke:darkred,color:darkred;
    style pv fill:white,stroke:white;
    style sl fill:white,stroke:white,color:darkblue;
    style sg color:black,fill:white,stroke:black;
    style sc color:darkblue,stroke:darkblue,fill:white;
    style eq color:darkgreen,stroke:white,fill:white;
    style neq color:darkred,stroke:white,fill:white;
    style rv color:darkred,stroke:darkred,fill:white;
    classDef subgraphs fill:white,stroke:white,font-size:small,text-align:center, display: block;
    class sgsubgraph,scsubgraph,rvsubgraph subgraphs;
    style scsubgraph color:darkblue;
    style rvsubgraph color:darkred;
    style tmp display: none;
    style tmp2 display: none;
  ```
The Graphvizo probably contains more features but maybe I would prefer the Mermaid because of its easier use (support by GitHub itself). WDYT?  Also if you have remarks about the image itself, could you please write them down for me?

viktormalik:
> The Graphvizo probably contains more features but maybe I would prefer the Mermaid because of its easier use (support by GitHub itself). WDYT?

Agreed, I also really like the way Mermaid image looks. IMHO it's much nicer than the default Graphviz style.

> Also if you have remarks about the image itself, could you please write them down for me?

Sure, a couple of things, mostly styling:
- If it fits, I'd use `diffkemp build` instead of `build` (same for other commands) and typeset it in truetype (if possible) so that it highlights that those are DiffKemp commands that can be used in the given stage.
- I understand that the point of all-uppercase labels is to make them closer to the tool's command line synopsis but in this case, I think that it may be a bit nicer if we used lowercase, just as everywhere else.
- It might be clearer if the "report for not equal symbols" was on the edge from "Semantic comparison" to "not equal" and the dashed edge to "Result viewer" led from the label only.
- This may be a hard (maybe impossible) thing to do but my aesthetic eye doesn't like the asymmetry, especially between the "old snapshot" and "new snapshot" edges.
- Last and not too important, I'd make the image a bit more "compact", especially the rightmost arrows look way too long (perhaps we could wrap line in "report for non equal symbols" if that's possible to make them shorter).
- Also, why is the "symbol list" part blue? The way I understand it is that the build phase is black, the compare phase is blue, and the result phase is green/red. IMO the blue color is not needed at all and we could have most of the picture black.

README.md
```diff
@@ -10,150 +10,118 @@ The main use-case of DiffKemp is to compare selected functions and configuration
...
-once such as:
-* `kernel.*`
-* `kernel.{sysctl-1|sysctl-2}`
+## Development
 
-Currently, these sysctl option groups are supported: `kernel.*`,
-`vm.*`, `fs.*`, `net.core.*`, `net.ipv4.conf.*`.
+- [Development guide](docs/development.md)
 
 ## Important implementation details
```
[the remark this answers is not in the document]
PLukas2018: I deleted the section and moved the debug info sentence to brackets in snapshot generation paragraph.

README.md
```diff
@@ -291,22 +149,26 @@ order:
...
 - ICST'21 [paper](https://ieeexplore.ieee.org/document/9438578)
   and [talk](https://zenodo.org/record/4658966):  
   V. Malík and T. Vojnar, "Automatically Checking Semantic Equivalence between
   Versions of Large-Scale C Projects," 2021 14th IEEE Conference on Software
   Testing, Verification and Validation (ICST), 2021, pp. 329-339.
-- NETYS'22
-  [paper](https://link.springer.com/chapter/10.1007/978-3-031-17436-0_18) and
-  [talk](https://www.youtube.com/watch?v=FPOUfgorF8s):  
+- DevConf.CZ'19 talk  
+  [![DevConf.CZ'19 talk](http://img.youtube.com/vi/PUZSaLf9exg/0.jpg)](https://www.youtube.com/watch?v=PUZSaLf9exg)
```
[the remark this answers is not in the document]
PLukas2018: I reverted it.

docs/usage.md
```diff
@@ -0,0 +1,117 @@
...
+- `--no-opt-override`: Uses optimisation options provided in the project's
+  Makefile or specified with `--clang-append="-OX"`. With this option DiffKemp
+  can potentially handle more complex refactoring (report fewer false
+  positives). However, this may reduce precision in identifying the exact
+  location of semantic differences (eg. reporting if a difference is in
+  a macro).
+- `--clang-append="..."`: Allows specifying options that will be appended to
+  `clang` when compiling source files to LLVM IR.
+- Additional options to configure the project build, see `diffkemp build --help`
+  for the complete list.
```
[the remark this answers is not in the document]
PLukas2018: I added all options with descriptions to all commands, for some commands I rewrote the description from help with some rewording, for other I added some notes (eg. that `--enable-all-patterns` also runs the `control-flow-only` pattern which is not semantic-preserving. I have not yet used all options so if anybody has some useful notes which could be added, let me know. Also potentially we could leave adding the options to docs for another PR.

docs/usage.md
```diff
@@ -0,0 +1,117 @@
...
+In this case, the path to the file should be given in place of `PROJ_DIR`.
+
+#### Options
+
+- `--no-opt-override`: Uses optimisation options provided in the project's
+  Makefile or specified with `--clang-append="-OX"`. With this option DiffKemp
+  can potentially handle more complex refactoring (report fewer false
+  positives). However, this may reduce precision in identifying the exact
+  location of semantic differences (eg. reporting if a difference is in
+  a macro).
```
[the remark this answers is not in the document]
PLukas2018: Yes, you are right.

PLukas2018:
> * Also, why is the "symbol list" part blue? The way I understand it is that the build phase is black, the compare phase is blue, and the result phase is green/red. IMO the blue color is not needed at all and we could have most of the picture black.

I think I made the `symbol list` blue to represent that it is also 'used' in the semantic comparison (it sets what functions will be compared).

PLukas2018:
All comments should be now hopefully resolved.

I reworked the image using the Mermaid according to the comments. For now, I only included it in the main README and not in the `docs/usage.md` (to not repeat the same 'code'). Btw I am not sure I understood this correctly:
> It might be clearer if the "report for not equal symbols" was on the edge from "Semantic comparison" to "not equal" and the dashed edge to "Result viewer" led from the label only.

So if I misinterpreted it, let me know and I will try to fix it.

Also I am thinking if it is a good idea to commit the viewer screenshots to the repo, but I am not sure how to handle that differently. So if you have any suggestions, let me know.

viktormalik:
> All comments should be now hopefully resolved.
> 
> I reworked the image using the Mermaid according to the comments.

Nice! The only disadvantage is that the Mermaid image doesn't render on the GitHub app (at least not on my tablet) - I just see the code block.

> For now, I only included it in the main README and not in the `docs/usage.md` (to not repeat the same 'code').

Is there a way to define the image source in a separate file and then "include" it so that we don't have to maintain two versions?

> Btw I am not sure I understood this correctly:
> 
> > It might be clearer if the "report for not equal symbols" was on the edge from "Semantic comparison" to "not equal" and the dashed edge to "Result viewer" led from the label only.
> 
> So if I misinterpreted it, let me know and I will try to fix it.

Almost :-) I meant the dashed edge to lead from the "report for not equal symbols" label to the "Result viewer" box (with the "diffkemp view" command on top of it). That way it's more clear that the input to the viewer is the report which is one of the outputs of an unsuccessful comparison.

> Also I am thinking if it is a good idea to commit the viewer screenshots to the repo, but I am not sure how to handle that differently. So if you have any suggestions, let me know.

This is IMO fine.

docs/usage.md
```diff
@@ -0,0 +1,212 @@
...
+- `PROJ_DIR`: Path to a project's root directory or a path to a single C file.
+- `SNAPSHOT_DIR`: Output directory for storing the created snapshot.
+- `SYMBOL_LIST`: Path to file containing a list of symbols (each symbol on
+   a single line) which should be prepared for comparison.
+- `--no-opt-override`: Uses optimisation options provided in the project's
+  Makefile or specified with `--clang-append="-OX"`. With this option, DiffKemp
+  can potentially handle more complex refactoring (report fewer false
+  positives). However, this may reduce precision in identifying the exact
+  location (e.g. function or macro) of a semantic difference.
+- `--target TARGET`: Allows specifying make targets which should be used to
```
viktormalik, line 48: Let's try to use code-style formatting consistently. Especially `make` is important as it's a common verb in English and not formatting it makes the sentences harder to understand.
```suggestion
- `--target TARGET`: Allows specifying `make` targets which should be used to
```
Please also check other places (I may have missed something).

docs/usage.md
```diff
@@ -0,0 +1,212 @@
...
+  build the snapshot from the project.
+- `--reconfigure`: Reconfigures autotools-based project with `CC='DiffKemp
+  compiler wrapper'`.
```
viktormalik, line 51: Slightly more understandable (at least to me) that it's not a string but a path to the wrapper.
```suggestion
- `--reconfigure`: Reconfigures autotools-based project with `CC=<DiffKemp
  compiler wrapper>`.
```

docs/usage.md
```diff
@@ -0,0 +1,212 @@
...
+
+- `KERNEL_DIR`: Path to a kernel's root directory.
+- `SNAPSHOT_DIR`: Output directory for storing the created snapshot.
+- `SYMBOL_LIST`: Path to a file containing a list of symbols (each symbol on
+  a single line) to be prepared for comparison.
+  In case `--sysctl` is used, the list is interpreted as a list of sysctl
+  parameters.
+- `--sysctl`: Compares sysctl option.
+- `--no-souce-dir`: Does not store the path to the source kernel directory in
+  the snapshot.
```
viktormalik, line 117: A bit of more context (check wrapping please, it's just random):
```suggestion
- `--no-souce-dir`: Does not store the path to the source kernel directory in
  the snapshot. This is useful if the comparison is done on a different system
  than building the snapshot (i.e. the path to the original kernel tree does not
  exist anymore). May lead to generating more false positives in some situations.
```

docs/usage.md
```diff
@@ -0,0 +1,212 @@
...
+Syntactic diffs of the discovered differences are stored in separate files
+(one file for each compared function that is different) in a newly created
+directory. The name of the directory can be specified using the `-o` option,
+otherwise it is generated automatically.
+
+### Options
+
+- `SNAPSHOT_DIR_1`, `SNAPSHOT_DIR_2`: Paths to directories containing snapshots
+  of the project's version to be compared.
+- `--output-dir OUTPUT_DIR`, `-o OUTPUT_DIR`: Name of the output directory.
```
viktormalik, line 155: I think this is understandable and shorter:
```suggestion
- `-o`, `--output-dir OUTPUT_DIR`: Name of the output directory.
```
Same for the other options below which have both the short and the long variant.

docs/usage.md
```diff
@@ -0,0 +1,212 @@
...
+- `--show-diff`: Shows/creates syntactic diffs for symbols evaluated as
+  semantically different (used by default, `--no-show-diff` turns this off).
+- `--full-diff`: Shows syntactic diff for all functions (even semantically
+  equivalent ones).
+- `--enable-pattern`, `--disable-pattern`: Enables/disables specified built-in
+  pattern(s).
+- `--enable-all-patterns`,`--disable-all-patterns`: Enables/disables all
+  supported built-in patterns. Be careful with `--enable-all-patterns`, as it
+  also enables the `control-flow-only` pattern, which is not
+  semantic-preserving.
```
viktormalik, line 167: There are more patterns which are disabled by default. How about this:
```suggestion
- `--enable-all-patterns`,`--disable-all-patterns`: Enables/disables all
  supported built-in patterns. Be careful, `--enable-all-patterns`
  also enables patterns which are not on by default and may not be 
  semantics preserving.
```
Later, we'll add a reference to the documentation of patterns.

docs/usage.md
```diff
@@ -0,0 +1,212 @@
...
+- `--report-stat`: Reports basic statics of the analysis -- the number of
+  symbols evaluated as `equal` and `not equal` and other statistics:
```
viktormalik, line 169: Let's put all parts to the list:
```suggestion
- `--report-stat`: Reports basic statics of the analysis:
  - `equal`: ...
  - `not equal`: ...
```

docs/usage.md
```diff
@@ -0,0 +1,212 @@
...
+  - `empty diff`: Reports for how many not-equal symbols were found
+    differences in symbols with no syntax difference.
+  - `unknown`: Represents how many symbols DiffKemp could not evaluate
+    (mainly caused by the symbol occurrence only in one version of
+    the program),
+  - `errors`: Represents symbols for which the comparison failed.
+- `--extended-stat`: Reports extended statistics -- total number of compared
+  functions (including called ones) and other information, such as the number
+  of compared instructions and the total number of found differences.
```
viktormalik, line 178:
```suggestion
- `--extended-stat`: Reports extended statistics -- total number of compared
  functions (including called ones) and other information, such as the number
  of compared instructions and the total number of found differences. Beware
  that this may increase the analysis time.
```

docs/usage.md
```diff
@@ -0,0 +1,212 @@
...
+- `--show-errors`: Show functions that are either unknown or ended with an
+  error in statistics.
+- `--stdout`: Mainly for development, prints results to standard output instead
+  of saving them to the directory.
```
viktormalik, line 182: Not necessarily for development only:
```suggestion
- `--stdout`: Prints results to standard output instead
  of saving them to the directory.
```

README.md
```diff
@@ -10,278 +10,184 @@ The main use-case of DiffKemp is to compare selected functions and configuration
...
+    nodeBorder: black
+    mainBkg: white
+    clusterBkg: white
+    clusterBorder: white
+---
+flowchart LR
+    %% nodes
+    pv(project versions)
+    sl(symbol list)
+    subgraph sgSubgraph[diffkemp build\ndiffkemp build-kernel\ndiffkemp llvm-to-snapshot]
```
viktormalik, line 56: Is there a way to make these labels in "code-style font", i.e. `diffkemp build`, ...? It would help to understand that they are commands.

viktormalik (commented):
Just a couple of comments, mainly for the lists of options for CLI commands.

Btw. the image looks great!

README.md
PLukas2018: I changed the font to [monospace](https://github.com/PLukas2018/diffkemp/blob/4ae8ff05360d3fa9d353bdef93f589a316d9d082/README.md?plain=1#L41), it looks like it was not previously aplicated because of comment which I added, now it seems to be fine.

PLukas2018:
> Nice! The only disadvantage is that the Mermaid image doesn't render on the GitHub app (at least not on my tablet) - I just see the code block.

Yes, it unfortunately seems to [not be supported](https://github.com/orgs/community/discussions/11595#discussioncomment-2199491) in the GitHub app.

> Is there a way to define the image source in a separate file and then "include" it so that we don't have to maintain two versions?

I did not find a way how to do that for Mermaid. I am thinking of maybe leaving it just in the main README.





> Almost :-) I meant the dashed edge to lead from the "report for not equal symbols" label to the "Result viewer" box (with the "diffkemp view" command on top of it). That way it's more clear that the input to the viewer is the report which is one of the outputs of an unsuccessful comparison.

Something like this?
image

PLukas2018:
I found out, that there is a problem with the graph look in a dark mode
image
I will look if there is a way to fix it.

viktormalik:
> > Nice! The only disadvantage is that the Mermaid image doesn't render on the GitHub app (at least not on my tablet) - I just see the code block.
> 
> 
> 
> Yes, it unfortunately seems to [not be supported](https://github.com/orgs/community/discussions/11595#discussioncomment-2199491) in the GitHub app.
> 

Yeah, I saw that discussion, too. It's not that big of a problem, it's just weird that the mobile app shows the Mermaid code instead of some placeholder image. If we could fix just that, it would be great.

> 
> > Is there a way to define the image source in a separate file and then "include" it so that we don't have to maintain two versions?
> 
> 
> 
> I did not find a way how to do that for Mermaid. I am thinking of maybe leaving it just in the main README.

Agreed, let's leave it in the main README for now. 

> 
> 
> 
> 
> 
> 
> 
> 
> > Almost :-) I meant the dashed edge to lead from the "report for not equal symbols" label to the "Result viewer" box (with the "diffkemp view" command on top of it). That way it's more clear that the input to the viewer is the report which is one of the outputs of an unsuccessful comparison.
> 
> 
> 
> Something like this?
> 
> image
> 
> 

Yes, that's what I meant. I'm still not sure if it's clear enough but I think that's it's sufficient for now.

PLukas2018:
I fixed the options in the usage document according to the comments.


> it's just weird that the mobile app shows the Mermaid code instead of some placeholder image. If we could fix just that, it would be great.

I did not find a way to do this but at least I added a comment at the top of the code block:

    ```mermaid
    ---
    # This code renders an image showing the relation of the commands.
    # The image does not show on the GitHub app, use a browser to see the image.
    ...

I updated the image (it looks slightly different than I previously posted because I lost the code to the screenshot I posted).
Also, I fixed the image look in a dark mode by setting a theme and wrapping the whole graph in a subgraph so the image has a white background. Because it is wrapped in the subgraph it looks like it is rendered a little bit smaller than it was but I did not find a different way how to set a white background for the image.

docs/usage.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Options
+
+- `SNAPSHOT_DIR_1`, `SNAPSHOT_DIR_2`: Paths to directories containing snapshots
+  of the project's version to be compared.
+- `-o`, `--output-dir OUTPUT_DIR`: Name of the output directory.
+- `-p`, `--custom-patterns CUSTOM_PATTERNS`: Path to a custom pattern file or
+  a configuration.
+- `--show-diff`: Shows/creates syntactic diffs for symbols evaluated as
+  semantically different (used by default, `--no-show-diff` turns this off).
```
viktormalik, line 162: Nit: since `--show-diff` is used by default, we should rather state `--no-show-diff` here:
```suggestion
- `--no-show-diff`: Do not show/create syntactic diffs for symbols evaluated as
  semantically different.
```

viktormalik (approved):
Nice! Just one nit, otherwise looks good to me.

Great job!

@FrNecas do you want to have one more look or should we go on and merge?

docs/usage.md
PLukas2018: Thanks, fixed.

docs/development.md
```diff
@@ -0,0 +1,250 @@
...
+If you make changes to the SimpLL library, you will need to rebuild it by
+running `ninja -C build`. We are using [CFFI](https://cffi.readthedocs.io/en/stable/)
+for accessing the library from the Python. The CFFI is not rebuilt by default,
+so even if you rebuild the library, the changes will not be visible when running
+`diffkemp compare ...`. Currently, you have multiple options to overcome this:
+
+1. Use `diffkemp compare --disable-simpll-ffi ...` which will run SimpLL
+  through a binary instead of the CFFI.
+2. On some OS distributions, it is enough to run `pip install -e .` again.
+3. Another option is to to run `rm build/_simpll.abi3.so` and then run `cmake`
```
FrNecas, line 141: I'd suggest replacing `rm build/_simpll.abi3.so` with `tools/simpll_clean.sh`

FrNecas (approved):
Very nice! I have just one minor suggestion that I think could be worth including but otherwise LGTM. This will be very helpful especially for new contributors IMO.

docs/development.md
PLukas2018: Unfortunately, in this situation the script does not help -- it exits without doing anything because the version of `LLVM IR` and of `SimpLL` library would be the same:
https://github.com/diffkemp/diffkemp/blob/eb7511df05bd27347df3eac27eca7e6cc83713a1/tools/simpll_clean.sh#L32-L37

The script (`tools/simpll_clean.sh`) could be mentioned somewhere, but I would rather leave it after #344 is merged if the script is still useful (also the script will probably need to be updated in #344 or after that).

viktormalik:
I think that we are good to merge.

PR #347, opened 2024-06-21
Add definitions of macros to `diffkemp-out.yaml`

PLukas2018:
This PR contains the last but one part of #314 (remains enhancements of viewer for visualisation of code/diff for `macro-function`/`function-macro`  differences.
The PR adds extracted information about the location of macro definitions to the `diffkemp-out.yaml` file which is created when `diffkemp compare` is run. Thanks to the added macro definitions (and thanks to previously merged PRs) it will be now possible to see the code of the macros in the result viewer.

### DIFF
--- a/diffkemp/output.py
+++ b/diffkemp/output.py
@@ -1,8 +1,105 @@
+"""
+Contains logic for creating output of DiffKemp compare command.
+
+Specifically, the YAML file is created here which contains
+information about found differences and definitions of symbols
+which are located in the differences.
+"""
 from diffkemp.semdiff.caching import ComparisonGraph
 from diffkemp.utils import get_end_line, EndLineNotFound
 from diffkemp.semdiff.result import Result
 import os
 import yaml
+
+
+class MacroDefinitions:
+    """Class for extracting information about macro definitions."""
+    def __init__(self, snapshot_dir_old, snapshot_dir_new, result):
+        self.old_dir = snapshot_dir_old
+        self.new_dir = snapshot_dir_new
+        self.result = result
+        # Information from call stacks for extracting defs for differing macros
+        # - list of tuples (vertex name, differing/last macro name).
+        self.diff_macro_infos = set()
+        # Definitions in YamlOutput format.
+        self.definitions = {}
+
+    def extract_from_result(self, called_res, compared_fun):
+        """
+        Extracts information about macro definitions from result of called
+        entity.
+        :param called_res: Result of called entity.
+        :param compared_fun: Name of compared function.
+        """
+        # Note: Definitions of differing/last macro is not possible to extract
+        # directly from the call stacks, we will now extract for it only
+        # name of the macro and function/vertex name, the definition will be
+        # extracted later in `_add_defs_for_differing` method.
+        if called_res.first.callstack:
+            macro_defs, diff_macro_info = called_res.first.callstack \
+                .get_macro_defs(compared_fun)
+            self._add_definitions(macro_defs, "old")
+            if diff_macro_info is not None:
+                self.diff_macro_infos.add(diff_macro_info)
+        if called_res.second.callstack:
+            macro_defs, diff_macro_info = called_res.second.callstack \
+                .get_macro_defs(compared_fun)
+            self._add_definitions(macro_defs, "new")
+            if diff_macro_info is not None:
+                self.diff_macro_infos.add(diff_macro_info)
+
+    def get_yaml_defs(self):
+        """Returns definitions of macros in YamlOutput format."""
+        # Extracting the missing definitions for differing macros.
+        self._add_defs_for_differing()
+        return self.definitions
+
+    def _add_defs_for_differing(self):
+        """Updates definitions with definitions of differing macros."""
+        if self.result.graph is None:
+            return
+        # For each differing/last macro finding SyntaxDiff object
+        # which contains definition of the macro.
+        for vertex_name, diff_macro_name in self.diff_macro_infos:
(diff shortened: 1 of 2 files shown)


diffkemp/output.py
```diff
@@ -15,10 +112,15 @@ def __init__(self, snapshot_dir_old, snapshot_dir_new, result):
         self.output = {}
         # Sets with symbol names
         self.function_names = set()
-        self.macro_names = set()
         # Note: type_names contains tuples (type name, parent name)
+        #   parent name - name of function in which type is used
```
PLukas2018, line 116: Relevant comment from #314:
> Could be a macro, right?
> ```suggestion
>         #   parent name - name of function/macro in which type is used
> ```

I think, that currently, we are reporting type differences only from functions. Even if the type was originally used in a macro, the call stacks report type differences from the function in which the macro with the differing type was used. So the parent name (meaning the name of the previous call from the call stack) for type/struct should always be a function.

diffkemp/output.py
```diff
@@ -1,10 +1,107 @@
...
+        :param compared_fun: Name of compared function.
+        """
+        # Note: Definitions of differing/last macro is not possible to extract
+        # directly from the call stacks, we will now extract for it only
+        # name of the macro and function/vertex name, the definition will be
+        # extracted later in `_add_defs_for_differing` method.
+        if called_res.first.callstack:
+            macro_defs, diff_macro_info = called_res.first.callstack \
+                .get_macro_defs(compared_fun)
+            self._add_definitions(macro_defs, "old")
```
FrNecas, line 41: Hm, I am not a big fan of the `"old", "new", "kind"` string literal hard-coding. It would be nice if we could use a string enum for this instead but I am not sure which python versions diffkemp aims to support (we are still supporting something old, aren't we?). If that is the case, feel free to ignore this.

FrNecas (commented):
Looks good as far as I can tell, just a minor nitpick (obviously I may be missing some context of the previous PRs related to macros)

diffkemp/output.py
PLukas2018: Thanks for the review. I replaced the `old`/`new` with boolean parameter `is_old`. For the `kind` I used enum which inherits from `str, Enum` (this should be available from Python 3.4), I could use `StrEnum` but it is supported from Python 3.11 and as you said I am also not sure which Python versions we want to support.

diffkemp/output.py
```diff
@@ -1,10 +1,119 @@
...
+    """Kind of symbol."""
+    FUNCTION = "function"
+    TYPE = "type"
+    MACRO = "macro"
+
+    def __str__(self):
+        return self.value
+
+
+class MacroDefinitions:
```
viktormalik, line 26: Why is macro collection done using a class while function and types are handled directly by methods of `YamlOutput`. Are macro somewhat special that the collection must keep some state?

tests/unit_tests/result_test.py
```diff
@@ -129,13 +129,26 @@ def test_yaml_output(result, mocker):
...
+    # (does not contain path to snapshot).
     def mock_rel_path(path, start):
         return path
     mocker.patch("os.path.relpath", side_effect=mock_rel_path)
 
-    yaml = YamlOutput("/abs/path/to/old-snapshot", "/abs/path/to/new-snapshot",
-                      result).output
+    # Mocking path.join for the same reason as relpath.
+    # Note: This mock can be problem if diffkemp uses >2 paths in join.
+    def mock_path_join(path1, *path2):
```
viktormalik, line 144: Mocking functions from `os.path` sounds very dangerous. Are we not able to come up with a better solution?

viktormalik (commented):
TBH, I'm quite confused by the new code and it seems over-complicated to me at this point. I'm not saying that it's not the best way, it's just that I'm failing to see why that's the case.

diffkemp/output.py
PLukas2018: > Why is macro collection done using a class while function and types are handled directly by methods of `YamlOutput`. Are macro somewhat special that the collection must keep some state?

The macro collection is done in a separate class mostly because the extraction of the definition is a little bit more complex so I created a class for it so the logic is encapsulated, also it was done in reaction to this comment on the original PR https://github.com/diffkemp/diffkemp/pull/314#discussion_r1517429792

diffkemp/output.py
```diff
@@ -60,12 +175,14 @@ def _create_results(self):
                     if called_res.second.callstack else []
                 # updating symbol names
                 if called_res.first.callstack:
+                    # Note: Macro names are not currently used.
```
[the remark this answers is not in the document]
PLukas2018: > Also an explanation of why the macros are handled differently from functions and types would be very helpful.

- So for functions, the location of the definition (file and line where the definition is located) is saved directly in `Vertex` class/object (class representing cached results for a function) in `ComparisonGraph`.
- For types the location of the definition is cached directly in  `TypeDiff` class/object which is accessible from `Vertex` class/object (from the function which contains/uses the differing type).
- In the case of the macros, the location of definitions is not cached directly but it is possible to extract them from macros `callstack` in `SyntaxDiff` class which is accessible from `Vertex` class (function which uses macros with differences) or to extract them from the final created `callstack` (which I currently do). For the differing macro is the location of the definition not possible to extract from `callstack` but it is cached in `SyntaxDiff` (PR #333).

viktormalik:
Yeah, it seems that there are many ways to do this. From the first glance, option 3 seems like the best but I think that we should have an in-person discussion on this to better understand what the problem is and what the best solution would be.

PR #348, opened 2024-06-30
Refactoring: `CallInfo` and `FunctionInfo` to be using `CodeLocation`

PLukas2018:
In PR #333, the `CodeLocation` struct was created to encapsulate information about code location (name of symbol, line, and file on which the symbol appears). This PR refactors  `CallInfo` and `FunctionInfo` structs, which previously contained the `name`, `line`, and `file` fields, to now extend the `CodeLocation` struct.

### DIFF
--- a/diffkemp/simpll/Result.h
+++ b/diffkemp/simpll/Result.h
@@ -24,6 +24,22 @@
 
 using namespace llvm;
 
+/// Type for storing information about code location of an 'object'
+/// (eg. location of a function or a macro definition).
+struct CodeLocation {
+    // Name of the 'object'.
+    std::string name;
+    // Line in the file.
+    unsigned line;
+    // A C source file name.
+    std::string file;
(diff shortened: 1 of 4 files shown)


diffkemp/simpll/Result.h
```diff
@@ -24,6 +24,22 @@
 
 using namespace llvm;
 
+/// Type for storing information about code location of an 'object'
+/// (eg. location of macro definition).
```
viktormalik, line 28: Since functions are the most common example, let's have them in the comment.
```suggestion
/// (eg. location of a function or a macro definition).
```

viktormalik (approved):
Looks good! Just one small comment adjustment.

diffkemp/simpll/Result.h
PLukas2018: Thanks, fixed.

PR #349, opened 2024-07-03
Iterate over phisToCompare based on size

FrNecas:
Range-based for-loops do not iterate over elements that get added during iteration, e.g., by using .emplace_back(). This could have caused a false negative when comparing program with a specific arrangment of PHI nodes, e.g., when the result of one PHI is used directly inside another PHI. In this case, the first PHI is not present in the phisToCompare vector (its only user was skipped during analysis) but must be compared nonetheless. In this case, the first PHI is inserted during cmpPHIs, i.e., while iterating over phisToCompare. We must therefore consider even such PHI chains and support insertion during iteration.


viktormalik (approved):
Nice catch!

PR #350, opened 2024-07-04
CI: split workflow files

viktormalik:
So far, we put all CI actions into a single workflow file which is getting quite huge. This separates some jobs into their own workflow files.

### DIFF
--- a/.github/workflows/viewer.yml
+++ b/.github/workflows/viewer.yml
@@ -1,0 +1,35 @@
+name: Viewer CI
+
+on: [push, pull_request]
+
+jobs:
+  result-viewer:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [14.x, 16.x, 18.x, 20.x]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Setup Node ${{ matrix.node }}
+      uses: actions/setup-node@v4
+      with:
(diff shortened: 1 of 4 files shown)


.github/workflows/viewer.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Run Cypress tests
+      uses: cypress-io/github-action@v6
+      with:
+        install: false
+        component: true
+        browser: firefox
+        working-directory: view
+        config: video=false,screenshotOnRunFailure=false
+    - name: Check coding style with ESLint
+      run: npm run lint --prefix view
```
PLukas2018, line 33: I am wondering if we do not want to move this to the `code-style.yml` file?

PLukas2018 (approved):
Looks good to me. I have just one note.

.github/workflows/viewer.yml
viktormalik: Yeah, good idea. Does it make sense to run the linter for each Node version? Or just the latest one is fine?

viktormalik:
I did some changes to the code-style workflow (split into multiple jobs), another review would be useful.

.github/workflows/viewer.yml
PLukas2018: I think that just the latest one should be enough.

FrNecas (approved):

PR #357, opened 2024-07-10
Viewer: react-diff-view empty context line fix

PLukas2018:
The [`react-diff-view`](https://github.com/otakustay/react-diff-view/) package which we are using for rendering code in the viewer contained a bug that was fixed in `3.2.1` version. This PR updates the used version of `react-diff-view` package to the latest version (containing the fix).

If a diff contained an empty context line below the diff, the bug was causing the line was not rendered.  When the rest of the code was expanded the line was still missing and sometimes the last shown line was duplicated.
The bug could be encountered e.g. between `4.18.0-193.el8` and `4.18.0-240.el8` version when viewing code for `schedule_hrtimeout_range_clock` differing function in `schedule_hrtimeout_range` compared function.
The code looked before the update like this:
- Missing context line below diff: image
- Missing line: image
- Duplicated last lines: image

After the update it looks fine:
- Empty context line below diff is shown: image
- No missing line, no duplication of lines: image

Also after the version update, warnings about missing source map was showing when running `bin/diffkemp view --devel` which polluted a lot of the console output, so the second commit was added to hide the warnings.

### DIFF
--- a/view/package-lock.json
+++ b/view/package-lock.json
@@ -20,7 +20,7 @@
         "prop-types": "^15.8.1",
         "react": "^18.2.0",
         "react-bootstrap": "^2.7.2",
-        "react-diff-view": "^2.4.10",
+        "react-diff-view": "^3.2.1",
         "react-dom": "^18.2.0",
         "react-error-boundary": "^4.0.4",
         "react-scripts": "5.0.1",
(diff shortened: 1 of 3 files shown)


viktormalik (approved):
Nice catch!

PR #358, opened 2024-07-11
Tests in llvm ir continuation

PLukas2018:
In this PR, more `DifferentialFunctionComparator` unit tests were rewritten using LLVM IR.
To do this, methods allowing testing certain `DifferentialFunctionComparator` methods in LLVM IR were added. 
This allows to rewrite unit tests which looked like this:

```cpp
TEST_F(DifferentialFunctionComparatorTest,
       CmpOperationsWithOpDiffTypes) {
    BasicBlock *BBL = BasicBlock::Create(CtxL, "", FL);
    BasicBlock *BBR = BasicBlock::Create(CtxR, "", FR);

    auto ConstL = ConstantInt::get(Type::getInt32Ty(CtxL), 2);
    auto AddL = BinaryOperator::Create(
            BinaryOperator::Add, ConstL, ConstL, "", BBL);
    auto CondL = ICmpInst::Create(llvm::Instruction::ICmp,
                                  llvm::CmpInst::ICMP_EQ,
                                  AddL,
                                  AddL,
                                  "",
                                  BBL);

    auto ConstR = ConstantInt::get(Type::getInt64Ty(CtxR), 2);
    auto AddR = BinaryOperator::Create(
            BinaryOperator::Add, ConstR, ConstR, "", BBR);
    auto CondR = ICmpInst::Create(llvm::Instruction::ICmp,
                                  llvm::CmpInst::ICMP_NE,
                                  AddR,
                                  AddR,
                                  "",
                                  BBR);

    ASSERT_NE(DiffComp->testCmpOperationsWithOperands(CondL, CondR), 0);
}
```

to look like this:

```cpp
TEST_F(DFCLlvmIrTest, CmpOperationsWithOpDiffTypes) {
    auto left = R"(define void @f() {
        %1 = add i32 2, 2
        %cond = icmp eq i32 %1, %1
        ret void
    })";
    auto right = R"(define void @f() {
        %1 = add i64 2, 2
        %cond = icmp ne i64 %1, %1
        ret void
    })";
    CREATE_FROM_LLVM(left, right);
    ASSERT_NE(testCmpOperationsWithOperands("cond"), 0);
}
```

The PR contains 15 rewritten unit tests, 14/15 of the rewritten tests were using/testing certain `DifferentialFunctionComparator` methods (e.g. cmpGEPs), and the last one was testing custom patterns.
Tests that are NOT rewritten:
- tests using dbg informations (eg. CmpGepsRenamed),
- tests which are already short/simple (eg. CmpAttrs),
- tests that compare constants with something  (eg. CmpValuesCastFromUnion) because I did not find a good way of
  rewriting/testing this.

Note: For now I left the rewritten tests at the same place (in `DifferentialFunctionComparatorTest.cpp` file) so it is easier to review, after that I will move it to `DFCLlvmIrTest.cpp` file.


tests/unit_tests/simpll/DFCLlvmIrTest.cpp
```diff
@@ -65,6 +75,107 @@ std::unique_ptr<Module> DFCLlvmIrTest::stringToModule(const char *llvm,
...
+}
+
+/// Compares instructions specified by name using cmpOperations method.
+int DFCLlvmIrTest::testCmpOperations(const char *name,
+                                     bool &needToCmpOperands,
+                                     bool keepSN) {
+    checkPrepared();
+    auto instL = getInstByName(*FL, name);
+    auto instR = getInstByName(*FR, name);
+    return DiffComp->testCmpOperations(instL, instR, needToCmpOperands, keepSN);
```
viktormalik, line 109: Shouldn't `checkValue` be here?
```suggestion
    checkValue(...);
    return DiffComp->testCmpOperations(instL, instR, needToCmpOperands, keepSN);
```
IIUC, `getInstByName` can return `nullptr` if the value is not an instruction.

tests/unit_tests/simpll/DFCLlvmIrTest.cpp
```diff
@@ -65,6 +75,107 @@ std::unique_ptr<Module> DFCLlvmIrTest::stringToModule(const char *llvm,
...
+}
+
+/// Compares instructions specified by name using cmpOperationsWithOperands
+/// method.
+int DFCLlvmIrTest::testCmpOperationsWithOperands(const char *name,
+                                                 bool keepSN) {
+    checkPrepared();
+    auto instL = getInstByName(*FL, name);
+    auto instR = getInstByName(*FR, name);
+    return DiffComp->testCmpOperationsWithOperands(instL, instR, keepSN);
```
viktormalik, line 119: Same as above.

tests/unit_tests/simpll/DFCLlvmIrTest.cpp
```diff
@@ -65,6 +75,107 @@ std::unique_ptr<Module> DFCLlvmIrTest::stringToModule(const char *llvm,
...
+}
+
+/// Compares instructions specified by names using cmpValues method.
+int DFCLlvmIrTest::testCmpInstValues(const char *nameL,
+                                     const char *nameR,
+                                     bool keepSN) {
+    checkPrepared();
+    auto valueL = getInstByName(*FL, nameL);
+    auto valueR = getInstByName(*FR, nameR);
+    return DiffComp->testCmpValues(valueL, valueR, keepSN);
```
viktormalik, line 146: Same as above.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
```diff
@@ -799,56 +747,48 @@ TEST_F(DifferentialFunctionComparatorTest, CmpMemsets) {
...
-                          "",
-                          BBL);
-    CR = CallInst::Create(AuxFR->getFunctionType(),
-                          AuxFR,
-                          {ConstantInt::get(Type::getInt32Ty(CtxR), 5)},
-                          "",
-                          BBR);
-    ASSERT_EQ(DiffComp->testCmpCallsWithExtraArg(CL, CR), 0);
-    ASSERT_EQ(DiffComp->testCmpCallsWithExtraArg(CR, CL), 0);
+TEST_F(DFCLlvmIrTest, CmpCallsWithExtraArgRight) {
```
viktormalik, line 772: I don't think that we need this one. The functionality is not used much anyways.

viktormalik (commented):
I love this!

Just two small things, otherwise this is a great simplification.

tests/unit_tests/simpll/DFCLlvmIrTest.cpp
PLukas2018: Yes, it can probably return `nullptr` (I am not sure what other things than instructions are saved in the `ValueSymbolTable` ),
1. I could either add the `checkValue` method or
2. update [the condition](https://github.com/PLukas2018/diffkemp/blob/23d8ecf217e1a741fc988801039debe7f87818aa/tests/unit_tests/simpll/DFCLlvmIrTest.cpp#L82) in the `getInstByName` method to be like this:
   ```cpp
       if (value && isa<Instruction>(value)) { // <- contains check that the value is an instruction
        return dyn_cast<Instruction>(value);
    }
    // aborting -- instruction with the given name does not exist
   ```
I like more the second option, because it makes sure that the function returns instruction and also it solves a possible problems in other `testCmp...` methods if the `getInstByName` returned a `nullptr`.

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
PLukas2018: I am not totally sure how you meant this, remove the `CmpCallsWithExtraArgRight` and leave just `CmpCallsWithExtraArgLeft`? Or did you meant something else?

tests/unit_tests/simpll/DFCLlvmIrTest.cpp
viktormalik: Agreed, with a small adjustment of the code :-)
```cpp
    if (auto inst = dyn_cast<Instruction>(value)) {
     return inst;
 }
 // aborting -- instruction with the given name does not exist
```

tests/unit_tests/simpll/DifferentialFunctionComparatorTest.cpp
viktormalik: Yeah, just leave the `Left` variant (maybe just name it `CmpCallsWithExtraArg`).

tests/unit_tests/simpll/DFCLlvmIrTest.cpp
PLukas2018:
```cpp
...
    auto value = fun.getValueSymbolTable()->lookup(name);
    if (value) {
        if (auto inst = dyn_cast<Instruction>(value)) {
            return inst;
        }
    }
...
```

I also added ` if (value) {` because in case the symbol was not found, the `dyn_cast<Instruction>(value)` was asserting in the LLVM IR library because of `nullptr`, I think that it is more user friendly to leave the assertion on us to inform more detaily about the problem.

PLukas2018:
I addressed the comments and moved the rewritten tests to the `DFCLlvmIrTest.cpp` file.

viktormalik (approved):
Thanks!

PR #359, opened 2024-07-17
build-kernel: ignore sysctl usages in non-x86 arch

viktormalik:
When searching for usages of a global variable corresponding to a sysctl parameter, there may be multiple functions of the same name (using the sysctl variable) for different arches. In such a case, the snapshot would use the latest one found which may lead to false positives if one snapshot ends up using the implementation from a different architecture than the other.

For building symbol definitions, we always [prefer](https://github.com/diffkemp/diffkemp/blob/master/diffkemp/llvm_ir/kernel_llvm_source_builder.py#L298-L300) definitions for the x86 architecture. This is impossible to do when searching for all functions using a variable, so this disables adding functions from non-x86 architectures to the snapshot.

Also update the documentation to state that we only support x86 in `build-kernel`.

### DIFF
--- a/diffkemp/diffkemp.py
+++ b/diffkemp/diffkemp.py
@@ -346,6 +346,11 @@ def generate_from_sysctl_list(snapshot, sysctl_list):
                 for data_fun in data_mod.get_functions_using_param(data):
                     if data_fun == proc_fun:
                         continue
+                    # For now, we only support the x86 architecture in kernel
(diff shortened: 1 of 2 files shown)


viktormalik:
@lenticularis39 mind having a look?

FrNecas (approved):
Sounds reasonable

Issue #363, opened 2024-07-29
`getPointeeStructTypeInfo` does not handle usage of `typedef` instead of `struct` (LLVM >= 15)


PLukas2018:
For LLVM >= 15 `getPointeeStructTypeInfo`  function uses debug metadata for finding the pointee type,
it does not currently handle situation when the struct type uses alias (`typedef`).

Example of a program that it does not handle
```c
// old.c
#include <string.h>                                                             
typedef struct {                                                                
    int a;                                                                      
    int b;                                                                      
} s;                                                                            
void f(s *var, char x) {                                                        
    memset(var, x, sizeof(s));                                                  
}
// new.c
#include <string.h>                                                             
typedef struct {                                                                
    int a;                                                                      
    int b;                                                                      
    int c;                                                                      
} s;                                                                            
void f(s *var, char x) {                                                        
    memset(var, x, sizeof(s));                                                  
} 
```
compiled using
```
diffkemp build --clang-append="-O2" --clang-append="-g" --no-opt-override ...
```
This would be compared as equal if `typedef` was not used.

The problem is located on this line https://github.com/diffkemp/diffkemp/blob/c682da53c761c27ef6e6eadf34c886d842261d94/diffkemp/simpll/Utils.cpp#L641, because in case `typedef` was used the base type is not composite type but derived typedef type which base type contains info about the structure type
```llvm
!13 = !DIDerivedType(tag: DW_TAG_typedef, name: "s", baseType: !14)
!14 = distinct !DICompositeType(tag: DW_TAG_structure_type, ..., size: 96, elements: !15)
```

To solve this it would be necessary to check if it is not `typedef` and if it is then it would be necessary to recursively search the base types until we get to the `DICompositeType` (info about the structure type - mainly its size).

PR #364, opened 2024-07-29
Fixes in `getPointeeStructTypeInfo` functions

PLukas2018:
This PR tries to fix 2 problems with `getPointeeStructTypeInfo` function which we are using in the `cmpMemset` and `cmpAllocs` methods when the set/allocated size is not the same in both compared functions. With the help of this function, we check that the set/allocated type names are equal and the set/allocated size corresponds to the size of the type. For LLVM 15 and above (when opaque pointers became enabled by default), we use debug intrinsics and debug metadata to get the info about the type to which the pointer passed to `memset`/`allocs` points.

In the case of LLVM 15+ there were problems:
- When the pointer passed down to the `memset`/`allocs` was of `void *` type. It tried to extract the type of the structure from the debug info which it did not contain. This caused assert by trying to `dyn_cast` a `nullptr`.
- If there were multiple debug data describing the pointer passed down to the `memset`/`allocs`. Some of the debug data does not have to contain information about the pointee type because it does not have to belong to scope of currently analysed function but e.g. to stdlib C `memset` function itself which was 'inlined' into the currently analysed function and does not contain info about the type because it uses `void *` instead.

This PR tries to fix (see commit messages for more details).

The PR also contains tests for these scenarios. I did not figure out a way how to simplify the writing of the debug info in the unit tests so the tests are not very pleasant to read.
Note: I am thinking that similarly as we are using kernel for regression/end-to-end testing, we could create our own test functions written in C with some YAML file describing the result of comparison and options that should be used for creating the snapshots from the functions (e.g. `-O2` and `-g` in this case). But I am note sure if it is a good idea because we would lose a level of testing detail (end-to-end test instead of unit testing of `cmpMemset` method).

While trying to solve this problem I also discovered that we are not handling situation when the pointee type is `typedef`ed, for now, I opened an issue #363 for this. I will try to fix this and open PR for it in the future.

I need to yet check if I did not break something -- I will compare the security libraries using this PR and if I will not discover a problem then I will move this PR out of draft.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
             asan: OFF
             regression-tests: true
 
-          - llvm: 14
+          - llvm: 17
             env:
               CC: gcc-9
               CXX: g++-9
--- a/diffkemp/simpll/DifferentialFunctionComparator.cpp
+++ b/diffkemp/simpll/DifferentialFunctionComparator.cpp
@@ -522,8 +522,10 @@ int DifferentialFunctionComparator::cmpAllocs(const CallInst *CL,
             isa<BitCastInst>(CR->getNextNode()) ? CR->getNextNode() : CR;
 
     // Retrieve type names and sizes
-    TypeInfo TypeInfoL = getPointeeStructTypeInfo(ValL, &LayoutL);
-    TypeInfo TypeInfoR = getPointeeStructTypeInfo(ValR, &LayoutR);
+    TypeInfo TypeInfoL =
+            getPointeeStructTypeInfo(ValL, &LayoutL, FnL->getName());
+    TypeInfo TypeInfoR =
+            getPointeeStructTypeInfo(ValR, &LayoutR, FnR->getName());
     if (TypeInfoL.Name.empty() || TypeInfoR.Name.empty()) {
         PREP_LOG("pointee type not found", ValL, ValR);
         LOG_KEEP(1);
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
             asan: OFF
             regression-tests: true
 
-          - llvm: 14
+          - llvm: 17
```
FrNecas, line 38: Is this change necessary? I thought that there was a reason why we run the address-sanitization CI on LLVM 14, that's why I didn't change this in the LLVM 16->17 update PR (I assumed this was the case because otherwise it would have been `llvm: 16` at the time but maybe we just forgot to update it somewhere along the way)

FrNecas:
Overall the code looks good. I will need to take one more look at the tests though, pretty hard to follow (but it's not your fault... that's just debug info)

.github/workflows/ci.yml
PLukas2018: > Is this change necessary?

Unfortunately, it is. As I described in the second commit message, it was necessary to add the `DIBuilder::createFunction` method to the tests so it works correctly. The method (respectively the `DISubprogram` object) looks like was leaking until LLVM 17 - https://www.github.com/llvm/llvm-project/commit/9eda46b1817f60378143846477ef71f0131a121d, so the tests would not pass for LLVM 14.

> I thought that there was a reason why we run the address-sanitization CI on LLVM 14

I hope there is not a reason, it looks like that the run was added ([commit](https://github.com/diffkemp/diffkemp/commit/e02cd02faa20efde3042eb73fa982a54443497a0) ) when LLVM 14 was the latest LLVM that we used in the CI. So probably we just forgot to update it with the following LLVM releases.

PLukas2018: I looked at PRs connected to changes in the workflow:
- PR for LLVM 15 (#242) was opened on Aug 18, 2022 and merged on Nov 2, 2022
- PR for ASAN (#243) was opened on Oct 3, 2022 and was merged on Oct 19, 2022

So the ASAN PR was opened after LLVM 15 PR but before it was merged. When the ASAN was merged, the latest supported LLVM was still LLVM 14, the LLVM 15 PR was still being reviewed. So I think that the reason why the ASAN is run on LLVM 14 is that before LLVM 15 was merged it was rebased on the master containing the ASAN but the version on which the ASAN is run was forgotten to be updated. Just to be sure, @zacikpa do you know if there is some specific reason why ASAN is run on LLVM 14 or can it be updated?

viktormalik: I'm pretty sure LLVM 14 was picked because it was the latest one supported at the time.

viktormalik (approved):
LGMT. Agreed that the tests are hard to read, we should think about making them more legible.

FrNecas (approved):

PR #366, opened 2024-07-30
Add llvm.dbg.assign to supported debug intrinsics

PLukas2018:
Add support of `llvm.dbg.assign` instruction [1] which was introduced in LLVM 16 [2]. Now we are skipping this intrinsics instruction which should solve some false positives.

[1] https://llvm.org/docs/SourceLevelDebugging.html#llvm-dbg-assign
[2] https://www.github.com/llvm/llvm-project/commit/33c7ae55e729069be754f56c4d4606cdeddd377b


FrNecas (approved):
Looks good, thanks for the fix. I was thinking if it wouldn't be better to separate this into an `if` outside the `return` (just to avoid the preprocessor directives inside an expression), however it probably wouldn't help much. It's OK with me this way thanks to the expression being split to multiple lines.

viktormalik (approved):
Thanks!

Issue #313, opened 2023-12-17
DiffKemp requires input when building RHEL 8.4 and 8.5 kernels


FrNecas:
When building the RHEL 8.4 and 8.5 kernels (`4.18.0-305.el8` and `linux-4.18.0-348.el8`, respectively) with the current version, DiffKemp hangs and is waiting for input while building the first function (`__alloc_disk_node`). Upon pressing enter, the build continues. This seems like a bug, I will try to take a closer look when I have time.

Steps to reproduce:
1. download the kernels using `rhel-kernel-get`
2. Run either `bin/diffkemp build-kernel kernel/linux-4.18.0-305.el8 snapshots/linux-4.18.0-305.el8 kernel/linux-4.18.0-305.el8/kabi_whitelist_x86_64` or `bin/diffkemp build-kernel kernel/linux-4.18.0-348.el8 snapshots/linux-4.18.0-348.el8 kernel/linux-4.18.0-348.el8/kabi_whitelist_x86_64`

PLukas2018:
I have tried to build `4.18.0-305.el8` and it seems that you are correct.

viktormalik:
Hint: when you debug this, try to `cd` to the kernel directory and manually run a build command, e.g.:
```
$ make V=1 --just-print block/genhd.o
```
That should also halt and you should see the reason (perhaps a missing config?).

PLukas2018:
I looked at this and it looks like:
- When using the standard dev shell (`nix develop`) for both downloading (with `rhel-kernel-get` script) and building the mentioned kernels (`diffkemp build-kernel`) there is no problem (DiffKemp does not hang).
- When using the shell `nix develop .#test-kernel-buildenv` for downloading (`rhel-kernel-get`) the mentioned kernels, the `CONFIG_KASAN` is undefined. And then when using the normal dev shell (`nix develop`) for building the mentioned kernels (`diffkemp build-kernel`) the building hangs at
  ```
  *
  * Restart config...
  *
  *
  * KASAN: runtime memory debugger
  *
  KASAN: runtime memory debugger (KASAN) [N/y/?] (NEW)
  ```
  because of the undefined `CONFIG_KASAN configuration`.

  To solve this, we could theoretically add `scripts/config --disable CONFIG_KASAN` inside the `rhel-kernel-get` script or inside the `build-kernel` command.

viktormalik:
Thanks for the analysis @PLukas2018.

[...]

PR #372, opened 2024-08-16
Check kernel config before creating snapshot

PLukas2018:
As discribed in issue #313, there can be problem with creating snapshot of kernel if it has missing configuration option. DiffKemp in that case is hanging for input which sets the option. Therefore this commit add check if the kernel is properly configured before creating snapshot
(running `make V=1 --just-print ...` which is used to get command which should be used for compiling of the source file to LLVM IR and which caused the hanging for the config).

The check of kernel config is managed by running `make oldconfig` which prompts for setting of necessary unset options. If we find out the command was prompting for input (contained message 'Restart config' or run too long), we end DiffKemp with apropriate error message.

### DIFF
--- a/diffkemp/llvm_ir/kernel_llvm_source_builder.py
+++ b/diffkemp/llvm_ir/kernel_llvm_source_builder.py
@@ -7,7 +7,9 @@ from diffkemp.llvm_ir.llvm_source_finder import LlvmSourceFinder, \
     SourceNotFoundException
 from diffkemp.llvm_ir.optimiser import opt_llvm, BuildException
 import os
+import shutil
 from subprocess import check_call, check_output, CalledProcessError
+import subprocess
 
 
 class KernelLlvmSourceBuilder(LlvmSourceFinder):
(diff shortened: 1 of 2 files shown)


diffkemp/llvm_ir/kernel_llvm_source_builder.py
```diff
@@ -162,6 +164,37 @@ def _enable_asm_features(self):
...
                 pass
 
+    def is_configured(self):
+        """
+        Check if the kernel is properly configured (all necessary options are
+        set).
+        """
+        config_file = os.path.join(self.source_dir, ".config")
+        config_backup_file = os.path.join(self.source_dir,
+                                          ".config-diffkemp-bkup")
```
viktormalik, line 174: Let's name it `.config.bak`, that's sort of standard.

viktormalik (commented):
LGTM, just one nit

diffkemp/llvm_ir/kernel_llvm_source_builder.py
PLukas2018: It is renamed, thanks.

viktormalik (approved):

PR #375, opened 2024-09-06
Docs: Fix broken diagram in `README.md`

PLukas2018:
Github updated to the latest version of Mermaid (v11.1.0), which caused issues with our diagram in `README.md`.
image
It looks like Mermaid v11.1.0 no longer supports `\n` for line breaks (probably caused by the introduction of new string syntax [1]). This was fixed by replacing `\n` with `<br/>` tags.

There was also a problem with long labels in the diagram:
- In Firefox, the whole label disappeared [2].
  image
- In Chrome, the label was auto-wrapped, resulting in a poor rendering output.
  image

To fix these problems, the font size of cluster labels was reduced.
image
Note: The font size is not ideal (too small) but for now I do not have better solution.

[1] https://www.github.com/mermaid-js/mermaid/pull/4271
[2] https://www.github.com/mermaid-js/mermaid/issues/5785

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ config:
 /* redefinition of subgraphs labels */
 .cluster-label {
   font-family: monospace;
-  font-size: 16px;
+  font-size: 13px;
 }
 "
   themeVariables:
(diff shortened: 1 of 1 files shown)


viktormalik (approved):
Nice, I think it's all good with the smaller font.

Thanks for the fix!

PR #376, opened 2024-10-17
CI: Check DiffKemp builds by running their binary

PLukas2018:
This PR adds to the CI checking that it is possible to use DiffKemp after building it. We have multiple ways how to build DiffKemp:
- nix development build (`nix develop`, `cmake, ninja`, `setuptoolsShellHook`)
- local development build (`cmake`, `ninja`, `pip install -e .`)
- build and installation from sources (`cmake`, `ninja`, `pip install .`, `install ...`).

Each way how to build DiffKemp is done and after that check that the DiffKemp is usable is done by building and comparing `make`-based project (which is already part of the repository and used for unit tests).

The CI should be able to check basic functionality of #344 with some adjustments (removal of `setuptoolsShellHook`, `pip install -e.`, ...).

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1,6 +1,8 @@
 name: Builds
 
 on: [push, pull_request]
+env:
+  llvm: 17
 
 jobs:
   nix-build:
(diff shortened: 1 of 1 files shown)


.github/workflows/builds.yml
```diff
@@ -14,6 +16,68 @@ jobs:
...
+        if: ${{ matrix.build-type == 'release' }}
+        run: |
+          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=Release
+          sudo ninja -C build install
+          pip install .
+          sudo install -m 0755 bin/diffkemp /usr/bin/diffkemp
+      - name: Development build of DiffKemp
+        if: ${{ matrix.build-type == 'development' }}
+        run: |
+          cmake -S . -B build -GNinja
```
viktormalik, line 54: Maybe add explicit debug build type?
```suggestion
          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=Debug
```

viktormalik (approved):
Very nice!

Just one nit, otherwise LGTM.

.github/workflows/builds.yml
PLukas2018: Thanks, I added it, also I added it to the nix development build.

DanielKriz:
I don't see anything wrong, good job.

DanielKriz (approved):

Issue #386, opened 2025-07-31
CI: Cypress tests failing


PLukas2018:
Cypress tests are curently failing. It could be caused by running on older (not supported anymore) Node.js versions. The Node.js versions should be updated to newer ones,  ...

PR #388, opened 2025-08-03
View: Migration to Vite, Node.js upgrade

PLukas2018:
Because of the failing [Cypress tests in CI](https://github.com/diffkemp/diffkemp/issues/386), upgrades to the DiffKemp result viewer are made.

We were previously:
- Using [create-react-app](https://create-react-app.dev/). The library was used for building and developing DiffKemp result viewer, the library is now deprecated. For this reason, I decided to move to **[Vite](https://vite.dev/) build system** which replaced the create-react-app.
- Supporting and testing on **Node.js** v14-20. Versions [v14-v19 reached end-of-life](https://nodejs.org/en/about/previous-releases), for this reason I decided to upgrade to support **v20** (maintance status), **v22** (active status) and **v24** (current status).
- Using `jest` for testing, with migration to `Vite`, the tests are migrated to **[Vitest](https://vitest.dev/)**.
---
More details in the commit message.

Btw. most of the changes (38,998 changes: 10,506 additions & 28,492 deletions) is in the `package-lock.json` which is generated and I am not able to do much about it.

I also tested the viewer manually on the difference between `linux-4.18.0-80.el8` and `linux-4.18.0-147.el8`, and I went over the results and found two problems (not rendering the default button color and alert at all). These problems were caused for an unknown reason by the migration and are fixed. Other than that I did not find any problem.


viktormalik (approved):
Thanks for fixing this!

Issue #384, opened 2025-07-21
Refactor `diffkemp.py`


viktormalik:
The `diffkemp/diffkemp.py` file got way too long and is quite hard to navigate within. It would be nice to refactor it.

A rough list of what could be done:
- [x] Move `build_c_project` and `build_c_file` to a separate source file (under `building/`?).
- [x] Move `build_kernel` to a separate source file (under `building`?).
- [x] Split `build_c_project` into multiple functions.
- [x] Move functions for generating lists of functions within a snapshot to separate source file or into the `Snapshot` class.
- [x] Move `compare` (and related functions) into a separate source file and split it into multiple functions.
- [ ] Move `print_syntax_diff` into a separate source file (under `syndiff`?).
- [x] Move the `Viewer` class into a separate source file.

Bal0sk:
I will do it.

viktormalik:
@Bal0sk is this now finished?

PR #389, opened 2025-08-18
Compare function refactoring

Bal0sk:
This PR focuses on refactoring of the compare function, for readability and maintainability purposes.

- Moving it to its own source file `compare.py`
- Splitting it into multiple helper functions

No functional changes are expected as logic was not changed.

### Open Question
While refactoring the compare function, I noticed that some functions, specifically `function_diff` and after refactoring `_print_information`, pass around a lot of variables. I was wondering if it would be beneficial to:

-  create a class, `CompareContext`, to group them together (`mod_first`, `mod_second`, `fun`, `glob_var`, `cache`, ...)
-  maybe, use some of the helper functions, as methods for the class (`_modules_to_cache`, `_set_output_dir`, ...)

This would require further refactoring of not only the `compare.py` file, but also the `function_diff` and other functions which would be affected. I would like an input of whether this would be beneficial for the readability.

### DIFF
--- a/diffkemp/compare.py
+++ b/diffkemp/compare.py
@@ -1,0 +1,204 @@
+from diffkemp.diffkemp import print_syntax_diff
+from diffkemp.config import Config
+from diffkemp.utils import CMP_OUTPUT_FILE
+from diffkemp.llvm_ir.llvm_module import LlvmParam, LlvmModule
+from diffkemp.semdiff.caching import SimpLLCache
+from diffkemp.semdiff.function_diff import functions_diff
+from diffkemp.semdiff.result import Result
+from diffkemp.output import YamlOutput
+from tempfile import mkdtemp
+from timeit import default_timer
+import errno
+import os
+import re
+import sys
+
+
+def compare(args):
+    """
+    Compare the generated snapshots. Runs the semantic comparison and shows
+    information about the compared functions that are semantically different.
+    """
+    # Set the output directory
+    if not args.stdout:
+        if args.output_dir:
+            output_dir = args.output_dir
+            if os.path.isdir(output_dir):
+                sys.stderr.write("Error: output directory exists\n")
+                sys.exit(errno.EEXIST)
+        else:
+            output_dir = default_output_dir(args.snapshot_dir_old,
+                                            args.snapshot_dir_new)
+    else:
+        output_dir = None
+
+    config = Config.from_args(args)
+    result = Result(Result.Kind.NONE, args.snapshot_dir_old,
+                    args.snapshot_dir_old, start_time=default_timer())
+
+    for group_name, group in sorted(config.snapshot_first.fun_groups.items()):
+        group_printed = False
+
+        # Set the group directory
+        if output_dir is not None and group_name is not None:
+            group_dir = os.path.join(output_dir, group_name)
+        else:
+            group_dir = None
+
+        result_graph = None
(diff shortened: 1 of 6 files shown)


viktormalik:
> I am not sure if it would be useful to create `CompareContext` class, there are some shared variables between the `function_diff` and `_print_information`, but personally I think that maybe transforming the `compare` function into class would be better. It should theoretically eliminate passing `config`, `output_dir` and `args` to `_print_information`.

Agreed with @PLukas2018, transforming the `compare` function into a class sounds like the way to go.

But we can leave that to the follow-up PR if you want.

Bal0sk:
I agree to make it a follow-up PR. Maybe, I could combine it with #384 issue of moving `print_syntax_diff`, since it is only used in the compare function, maybe it could be a method of the class as well.

PLukas2018 (commented):
Thanks for the changes. I have a few more notes which you can take a look at, but maybe wait for @viktormalik's review because I am not too experienced.

viktormalik:
> Thanks for the changes. I have a few more notes which you can take a look at, but maybe wait for @viktormalik's review because I am not too experienced.

All suggestions by @PLukas2018 look reasonable, I'll have a look later this week.

diffkemp/utils.py
```diff
@@ -4,6 +4,8 @@
 import sys
 
 LLVM_FUNCTION_REGEX = re.compile(r"^define.*@(\w+)\(", flags=re.MULTILINE)
+# Name of YAML output file created by diffkemp compare command.
+CMP_YAML_FILENAME = "diffkemp-out.yaml"
```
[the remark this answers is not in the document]
PLukas2018: @Bal0sk Nit. the variable should probably be renamed to the final name in the second commit rather than in the last.

diffkemp/.vscode/settings.json
PLukas2018, line 1: Remove this file, we could probably add `.vscode` directory to `.gitignore` to avoid these things in the future.

diffkemp/compare.py
```diff
@@ -202,3 +176,71 @@ def _get_modules_to_cache(functions, group_name, other_snapshot,
...
+def _filter_result_by_regex(regex_filter, fun_result):
+    pattern = re.compile(regex_filter)
+    for called_res in fun_result.inner.values():
+        if pattern.search(called_res.diff):
+            return
+    fun_result.kind = Result.Kind.EQUAL
+
+
+def _print_fun_result(fun_result, config, output_dir, fun, group_dir,
+                      group_name, args, old_fun_desc, group_printed):
```
PLukas2018, line 190: Maybe rename the args to compare_args also in other functions.

PLukas2018 (commented):
I have a few more notes, otherwise it looks good to me, but I have not tested it yet.

diffkemp/.vscode/settings.json
viktormalik: You can also set your own global `core.excludesFile` to exclude files across all your local Git repos.

PLukas2018 (commented):
It looks good to me too. Good job. I just noticed one little thing connected with formatting of an error message (see comment below).

I also compared `linux-4.18.0-80.el8` and `linux-4.18.0-147.el8` to check if an error did not occur during this refactorization. I compared the results gained before PR and after, and there is only one difference that I discovered (an amount of printed dashes after printing `Statistics`) which we can leave like that.

PR #393, opened 2025-08-29
Create flake8 config file. Fixes #392

### LINKED ISSUE #392: Create config file for `flake8`
We are using `flake8` for checking the code style of the Python part of DiffKemp.
When using `flake8`, we are ignoring some rules, specifically:
- `F403,F405,W504` for the `diffkemp/` directory,
- None for the `tests/` directory.

To simplify the usage of the `flake8` tool, it would be good to create [`flake8` configuration file](https://flake8.pycqa.org/en/latest/user/configuration.html), that would ignore the rules for given directories.

waliamehak:
Creates .flake8 config to centralize code style settings.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,5 @@
 [flake8]
+extend-select = W503
 per-file-ignores =
     # ignore specific rules in diffkemp
     diffkemp/*: F403, F405, W504
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,6 @@
+[flake8]
+per-file-ignores =
+    # ignore specific rules in diffkemp
+    diffkemp/*: F403, F405, W504
+    # do not ignore anything in tests
+    tests/*:
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -33,8 +33,7 @@ jobs:
         working-directory: ${{ github.workspace }}
         shell: bash
         run: |
-          flake8 diffkemp --ignore=F403,F405,W504
-          flake8 tests
+          flake8 diffkemp tests
 
   code-style-js:
     runs-on: ubuntu-latest
--- a/tests/unit_tests/caching_test.py
+++ b/tests/unit_tests/caching_test.py
@@ -207,8 +207,8 @@ def test_populate_predecessor_lists(graph):
     graph.normalize()
     graph.populate_predecessor_lists()
     for side in ComparisonGraph.Side:
-        assert ({v.names for v in graph["main_function"].predecessors[side]}
-                == {dup("looping")})
+        assert ({v.names for v in graph["main_function"].predecessors[side]} ==
+                {dup("looping")})
         assert ({v.names for v in graph["do_check"].predecessors[side]} ==
                 {dup("main_function")})
         assert ({v.names for v in graph["missing"].predecessors[side]} ==


PLukas2018:
Hi, thanks for the PR. I tried to test the PR (`flake8` behavior) by applying the following patch to DiffKemp:
```
diff --git a/diffkemp/diffkemp.py b/diffkemp/diffkemp.py
index 780c5fd..e92f58c 100644
--- a/diffkemp/diffkemp.py
+++ b/diffkemp/diffkemp.py
@@ -31,6 +31,9 @@ YAML_FILE_NAME = "diffkemp-out.yaml"
 
 def build(args):
     # build snapshot from make-based c project
+    var = (args.source_dir
+           + args.output_dir)
+    print(var)
     if os.path.isdir(args.source_dir):
         build_c_project(args)
     # make snapshot from single c file
```

When running the following command before this PR
```
flake8 diffkemp --ignore F403,F405,W504
```
the output is following:
```
diffkemp/diffkemp.py:35:12: W503 line break before binary operator
```

When running the following command using the `flake8` config file created in this PR
```
flake8 diffkemp
```
it does not output anything.

---

I am not sure if the config is not correct/used, or if there is some `flake8` inner problem/bug when ignoring the `W504`  rule through the command option vs through the config file. I played with it a little bit, but did not come to anything.

waliamehak:
Wasn't seeing W503 and W504 errors as both are ignored by default. Ran Flake8 without ignoring W504 in the config and observed several line break warnings:
```
flake8 --select=W504 diffkemp
```

With W504 included in the .flake8 ignore list, the warnings disappeared, which means it’s working fine.

waliamehak:
Hi @PLukas2018, just following up on this PR. I wanted to check if you had a chance to look at it yet?

PLukas2018:
So I looked into this again and tested it.
The config works fine -- code style errors are reported, `F403` and `F405` are ignored.

You are right that currently both `W503` and `W504` (the rules are saying if line break should be before or after binary operator) are ignored by default in `flake8`, but it was not like this in the past (https://github.com/diffkemp/diffkemp/commit/a75a6e99f90593291f5d6e4f08e46665af5524df).

Somehow, when specifying `flake --ignore W504` (Python `v3.10.12`, flake8 `v7.3.0`) on the command line, `W503` (line break after binary operator) is enforced. It does not work like this when specified in the `.flake8` config file, we would need to use something like `extend-select=W503`.

So the question is, if we want to enforce some coding style connected with **line break before/after binary operator**.  [`PEP8`](https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) says:
> In Python code, it is permissible to break before or after a binary operator, as long as the convention is consistent locally. For new code, Knuth’s style is suggested.

So in our case, I would enforce `W503` (line breaks occurring after the binary operator):
```python3
# OK
var = var1 +
      var2
# Not OK
var = var1
      + var2
```
because most of our code already follows this. @viktormalik what is you opinion?

PLukas2018:
> So, IIUC, we want to add W504 to the ignore list and fix all the places where line break doesn't occur after the operator. Let's make that part of this PR, ideally.

Almost, in the case of the `.flake8` config file, it looks like we need to add `extend-select=W503` to enforce a line break after the operator.

@waliamehak could you please update the PR?
- fixing all the places where line break doesn't occur after the operator,
- updating flake config file to enforce `W503`

.flake8
```diff
@@ -4,3 +4,4 @@ per-file-ignores =
     diffkemp/*: F403, F405, W504
     # do not ignore anything in tests
     tests/*:
+extend-select = W503
```
PLukas2018: Could you add a new line character at the end of the file, please?

PLukas2018 (commented):
Thanks for the quick update :+1: I have just two notes, otherwise it looks good to me:\
- Could you add at least a short message to each commit message (not just the title) describing the reason for the change.
- Could you setup your editor not to remove the trailing newline character from files.

waliamehak:
Hey @PLukas2018, let me know if everything is up to speed now.

waliamehak:
@PLukas2018, thanks for letting me know, done :)

PLukas2018 (approved):
Thanks, good job. LGTM :+1:

viktormalik (approved):

viktormalik:
Merged, thanks!

PR #397, opened 2025-09-10
Bump Up the LLVM Version to 18

DanielKriz:
This PR contains the changes needed for compiling and running diffkemp with LLVM-18.
Some core functions were renamed in this release and because we want to support the older version we have to add some macros that wrap around these changes.

I expect to clean up the commit history after the testing is done.

We also need to bump up the version of `nixpkgs` which should be separate PR that should be done before this PR is merged.


diffkemp/simpll/Utils.cpp
```diff
@@ -39,6 +39,28 @@
 #include <set>
 #include <sstream>
 
+#if LLVM_VERSION_MAJOR >= 18
+
+bool refHasSuffix(StringRef ref, StringRef suffix) {
```
viktormalik, line 44: `hasSuffix` is IMHO a better name.

diffkemp/simpll/SmtBlockComparator.cpp
```diff
@@ -21,6 +21,17 @@
...
+
+#if LLVM_VERSION_MAJOR < 18
+using SDivOperator = llvm::SdivOperator;
+using UDivOperator = llvm::UdivOperator;
+#else
+using SDivOperator = llvm::PossiblyExactOperator;
+using UDivOperator = llvm::PossiblyExactOperator;
+#endif
+
+}
```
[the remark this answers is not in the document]
viktormalik: I'd go with option 2 and, it it doesn't work for some reason, option 1 afterwards.

diffkemp/simpll/Utils.cpp
viktormalik: That should be IMO fine since that one has just one param and C++ has function overloading.

But even better, I'd rename the other one to something like `hasNumberSuffix` to better express its purpose.

viktormalik:
> I have update the documentation and changed the length of commit header. The decreased number of detected semantic equalities is a problem; however, I don't know what could have caused it other that some difference in LLVM IR between LLVM 17 and 18. I will try to investigate it and I am not sure what should be our strategy, whether we should merge this and the resolve it later, or resolve it now.

The new differences all have empty diffs so it's likely false positives. I'd say that it's some subtle difference in the new version of LLVM causing this.

> We might find more of these cases with LLVM19-21 and then investigate the differences in a bulk, there could be many of them (or they could help in investigation what is the cause, because if the number of correct analyses would increase then we could compare the differences in IR between that versions too).
> 
> What is your take @viktormalik ?

Since the LLVM support update is a priority now in order to make DiffKemp usable on newer systems, let's merge these as soon as CI passes and investigate later. I'd even not investigate them now, it is possible that some later LLVM update will fix this.

PR #401, opened 2025-09-15
Solve CMake warnings

PLukas2018:
When running *CMake*, it reports warnings. This PR resolves the warnings to ensure *CMake* works fine, even with newer *CMake* versions.


DanielKriz (approved):
The project is buildable without any warning and every choice is commented in the commit messages and their separation makes sense. LGTM.

viktormalik (approved):

PR #407, opened 2025-09-16
Make SimpLL compilable with clang, test it in the CI

PLukas2018:
As noticed by @DanielKriz, we are testing in the CI if the project is compilable with `gcc` but also with `clang`, but when we migrated to using Nix, the CI is using only `gcc` even when `clang` is specified as the compiler in the CI job matrix environment variable.
Screenshot from 2025-09-16 16-30-36

The problem is that the environment variables set in the CI are not passed down to the `nix develop` shell (it overwrites them), therefore, we must set them again in the shell.

This PR fixes issues to make the project compilable with *clang* and fixes the CI to also use  *clang*.

Screenshot from 2025-09-16 16-34-44

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,8 @@ jobs:
       matrix:
         llvm: [12, 13, 14, 15, 16, 17, 18]
         env:
-          - CC: gcc-9
-            CXX: g++-9
+          - CC: gcc
+            CXX: g++
         asan: [OFF]
         regression-tests: [true]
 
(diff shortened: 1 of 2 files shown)


DanielKriz (approved):
Nice fix and good job with that `z3`. Commit messages make sense and the code is clear.

We could make use of [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) in the future to simplify the commands needed for build. After my work on LLVM 19 integration I will look into it.

LGTM.

PR #411, opened 2025-09-22
SimpLL Unit tests: Enable logging

PLukas2018:
This PR enables turning on the logger in SimpLL unit tests to simplify debugging of failing tests.

This would be useful, e.g., in debugging of the recent problem connected with the `SmtDistributive` test (https://github.com/diffkemp/diffkemp/commit/a7be53eef62fafa7edf0d08b618231acd26e90d0), where it was necessary to increase the time for the SMT solver. This PR should simplify such problems by inspecting the logs. In the mentioned problem, if the unit tests were run as this:

```sh
VERBOSITY=1 ninja -C nix-build-llvm17/ test
```

[the output](https://github.com/user-attachments/files/22473617/out.txt) would contain:

```txt
[ RUN      ] DFCSmtTest.SmtDistributive
Analysing debugging information...
Analysing debugging information...
Trying to find possible relocation:
  FAILURE: Possible relocation was NOT found
SMT solver ran out of time/home/lukas/diffkemp-org/diffkemp/tests/unit_tests/simpll/DFCLlvmIrTest.cpp:1098: Failure
Expected equality of these values:
  DiffComp->compare()
    Which is: -1
  0

[  FAILED  ] DFCSmtTest.SmtDistributive (523 ms)
[----------] 1 test from DFCSmtTest (523 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (523 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] DFCSmtTest.SmtDistributive

 1 FAILED TEST
```

mentioning `SMT solver ran out of time`.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -202,6 +202,22 @@ by:
 ninja -C build test
 ```
 
+In case the tests fail, enabling logger can simplify the debugging process.
+The logger is enabled by setting the `SIMPLL_VERBOSITY` environment variable to
+a number, where:
+
+- `0`: No logging
+- `1`: Minimal logging
+- `2`: Moderate logging
+- `3` and more: Detailed logging (the most verbose)
+
+For example, to enable moderate logging while running tests with `ninja`, use
+the following command:
+
+```sh
+SIMPLL_VERBOSITY=2 ninja -C build test
+```
+
 ### Tests for the result viewer
 
 The result viewer contains unit tests and integration tests located in
--- a/tests/unit_tests/simpll/SimpLLTest.cpp
+++ b/tests/unit_tests/simpll/SimpLLTest.cpp
@@ -1,6 +1,39 @@
+#include "Logger.h"
+
+#include <cerrno>
 #include <gtest/gtest.h>
+#include <ostream>
+
+/// Sets logging verbosity based on `SIMPLL_VERBOSITY` env. var.
+/// Returns true if successful, otherwise returns false.
+bool trySetVerbosity() {
+    const char *verbosity = std::getenv("SIMPLL_VERBOSITY");
+    if (!verbosity)
+        return true;
+
+    int level = 0;
+    try {
+        level = std::stoi(verbosity);
+    } catch (const std::invalid_argument &e) {
+        std::cerr << "Error: `SIMPLL_VERBOSITY` must be a number!" << std::endl;
+        return false;
+    }
+
+    if (level < 0) {
+        std::cerr << "Error: `SIMPLL_VERBOSITY` cannot be a negative number!"
+                  << std::endl;
+        return false;
+    }
+
+    logger.setVerbosity(level);
+    return true;
+}
 
 int main(int argc, char **argv) {
+    if (!trySetVerbosity()) {
+        return EINVAL;
+    }
+
     ::testing::InitGoogleTest(&argc, argv);
     return RUN_ALL_TESTS();
 }


DanielKriz (changes requested):
The added functionality seems to work and there is only a few nitpicks from my side.

There is also some typos in the first commit message. But again, all of these are small things.

docs/development.md
```diff
@@ -202,6 +202,22 @@ by:
...
+- `1`: Minimal logging
+- `2`: Moderate logging
+- `3` and more: Detailed logging (the most verbose)
+
+For example, to enable moderate logging while running tests with `ninja`, use
+the following command:
+
+```sh
+SIMPLL_VERBOSITY=2 ninja -C build test
+```
```
DanielKriz, line 219: Btw, we should add this as an option to cmake. We are going to work it in when I will add workflows, etc.

tests/unit_tests/simpll/SimpLLTest.cpp
```diff
@@ -1,6 +1,35 @@
+#include "Logger.h"
+
+#include <cerrno>
 #include <gtest/gtest.h>
+#include <ostream>
+
+/// Sets logging verbosity based on `SIMPLL_VERBOSITY` env. var.
+/// Returns non-zero if the value is invalid.
+int setVerbosity() {
```
DanielKriz, line 9: In some languages (especially C#) it is common that functions that return pass/fail status and do some work in their body to have `try` prefix in front of their names. I suggest renaming this to `trySetVerbosity` and to return boolean.

Then in main we could return `EXIT_SUCCESS` or `EINVAL` directly.

diffkemp/simpll/Logger.cpp
```diff
@@ -83,6 +83,39 @@ void Logger::setIndent(size_t target_level) {
...
+void Logger::setDebugTypes(std::vector<std::string> &debugTypes) {
+    if (!debugTypes.empty()) {
+        DebugFlag = true;
+        // Transform vector of strings into char ** (array of char *)
+        std::vector<const char *> types;
+        std::transform(debugTypes.begin(),
+                       debugTypes.end(),
+                       std::back_inserter(types),
+                       [](const std::string &s) { return s.c_str(); });
+        setCurrentDebugTypes(&types[0], debugTypes.size());
```
DanielKriz, line 115: If I am not mistaken, then the `debugTypes` is not changed in this function. Then it could be made `const` and we could use constant iterators here `cbegin` and `cend`.

DanielKriz (changes requested):
Good job working in the previous suggestion. However, I have found two more places where it could be, perhaps, made better. What is your opinion? Because otherwise it is good to merge.

docs/development.md
PLukas2018: Yeah, it is a run-time setting, and I think it is better to leave it as run-time than a build-one (You can run tests without logging, then if you find out that a test does not pass, you turn on logging without the need to rebuild the project.). I was not sure either how @DanielKriz exactly meant it.

tests/unit_tests/simpll/SimpLLTest.cpp
```diff
@@ -1,6 +1,36 @@
+#include "Logger.h"
+
+#include <cerrno>
 #include <gtest/gtest.h>
+#include <ostream>
+
+/// Sets logging verbosity based on `SIMPLL_VERBOSITY` env. var.
+/// Returns true if successful, otherwise returns false.
+int trySetVerbosity() {
```
DanielKriz, line 9: This should return a boolean.

tests/unit_tests/simpll/SimpLLTest.cpp
```diff
@@ -1,6 +1,36 @@
...
+    const char *verbosity = std::getenv("SIMPLL_VERBOSITY");
+    if (!verbosity)
+        return true;
+    int level;
```
DanielKriz, line 13: Just to make the code absolutely clear define its value. Perhaps you can use `int level{}`.

DanielKriz (changes requested):
Please, change the return type `trySetVerbosity` and add some definition, otherwise it is great.

tests/unit_tests/simpll/SimpLLTest.cpp
```diff
@@ -1,6 +1,36 @@
...
+        return false;
+    }
+    if (level < 0) {
+        std::cerr << "Error: `SIMPLL_VERBOSITY` cannot be a negative number!"
+                  << std::endl;
+        return false;
+    }
+    logger.setVerbosity(level);
+    return true;
+}
```
DanielKriz, line 27: Just to increase readability add spaces on lines: 13, 20 and 25.

DanielKriz (changes requested):
One last super minor thing and then it is LGTM (you don't even have to ask for review after it is resolved).

viktormalik:
@DanielKriz since you requested changes, we need your approval for GitHub to let me merge this (without bypassing rules).

DanielKriz (approved):

PR #413, opened 2025-09-24
Encapsulation of compare functions into a class

Bal0sk:
This PR aims to improve maintainability of the compare functions in the `compare.py` file:

- by encapsulating the contents into a class.
-  refactoring the contents further

As part of this PR the `print_syntax_diff` function was also moved to the class and refactored. (#384)

While testing, I also encountered an error with the `--extended-stats` option in `compare` function, where the `result.graph` variable was never set and so the option crashed when printing. Looking back, this issue has been there before the #384. For now, I implemented a fix in the first commit of this PR, by setting it before the `print_stats` function. (The testing was done on linux-kernels from `examples`)

No functional changes are expected as logic was not changed, except the issue fix.

Additional notes:
I find that the `compare.py` file is a little long now, after adding the `print_syntax_diff` and adding additional class for output writing.
- I was thinking of moving the `print_syntax_diff` (as it is now), into a separate file or the `output.py` as was suggested in the issue #384. However, decided against it, for the time being, since one of the reasons, for the encapsulation of the compare functions, was to avoid passing many arguments between helper functions and also `print_syntax_diff` and moving the function would revert that change (or `args` and `config` would have to be passed). 
- another idea was to move only the newly created `OutputWriter` class somewhere separately, but since it is used only by `print_syntax_diff`, I am not sure if it would make sense to move it. 

I would like your thoughts on this matter as well as any additional notes.

### DIFF
--- a/diffkemp/compare.py
+++ b/diffkemp/compare.py
@@ -19,311 +19,338 @@ class OutputDirExistsError(Exception):
 
 def compare(args):
     """
+    Compare two snapshots. Prepares configuration and runs the comparison.
+    """
+    comparator = SnapshotComparator(args)
+    return comparator.run()
+
+
+class SnapshotComparator:
+    """
     Compare the generated snapshots. Runs the semantic comparison and shows
     information about the compared functions that are semantically different.
     """
-    # Set the output directory
-    try:
-        output_dir = _set_output_dir(args)
-    except OutputDirExistsError as e:
-        sys.stderr.write("{}".format(e))
-        sys.exit(errno.EEXIST)
+    MINIMAL_CACHE_FREQ = 2
 
-    config = Config.from_args(args)
-    result = Result(Result.Kind.NONE, args.snapshot_dir_old,
-                    args.snapshot_dir_old, start_time=default_timer())
+    def __init__(self, args):
+        self.args = args
+        self.config = Config.from_args(args)
+        self.result = Result(Result.Kind.NONE, args.snapshot_dir_old,
+                             args.snapshot_dir_old, start_time=default_timer())
+        self.regex_pattern = re.compile(args.regex_filter) \
+            if args.regex_filter else None
+        # Set the output directory
+        try:
+            self.output_dir = self._set_output_dir()
+        except OutputDirExistsError as e:
+            sys.stderr.write("{}".format(e))
+            sys.exit(errno.EEXIST)
 
-    for group_name, group in sorted(config.snapshot_first.fun_groups.items()):
-        group_printed = False
+    def _set_output_dir(self):
+        if self.args.stdout:
+            return None
 
-        # Set the group directory
-        if output_dir is not None and group_name is not None:
-            group_dir = os.path.join(output_dir, group_name)
-        else:
-            group_dir = None
+        if not self.args.output_dir:
+            return self._default_output_dir(self.args.snapshot_dir_old,
+                                            self.args.snapshot_dir_new)
+        output_dir = self.args.output_dir
+        if os.path.isdir(output_dir):
+            raise OutputDirExistsError(
+                "Error: output directory {} exists\n".format(output_dir))
+        return output_dir
 
-        result_graph = None
-        cache = SimpLLCache(mkdtemp())
-        module_cache = {}
+    @staticmethod
+    def _default_output_dir(src_snapshot, dest_snapshot):
+        """Name of the directory to put log files into."""
+        base_dirname = "diff-{}-{}".format(
+            os.path.basename(os.path.normpath(src_snapshot)),
+            os.path.basename(os.path.normpath(dest_snapshot)))
+        if os.path.isdir(base_dirname):
+            suffix = 0
+            dirname = base_dirname
+            while os.path.isdir(dirname):
+                dirname = "{}-{}".format(base_dirname, suffix)
+                suffix += 1
+            return dirname
+        return base_dirname
 
-        if args.enable_module_cache:
-            modules_to_cache = _get_modules_to_cache(group.functions.items(),
-                                                     group_name,
-                                                     config.snapshot_second,
-                                                     2)
-        else:
-            modules_to_cache = set()
+    def run(self):
+        for group_name, group in sorted(self.config.snapshot_first.
+                                        fun_groups.items()):
+            group_printed = False
 
-        for fun, old_fun_desc in sorted(group.functions.items()):
+            # Set the group directory
+            if self.output_dir is not None and group_name is not None:
+                group_dir = os.path.join(self.output_dir, group_name)
+            else:
+                group_dir = None
 
+            result_graph = None
+            cache = SimpLLCache(mkdtemp())
+            module_cache = {}
+
+            if self.args.enable_module_cache:
+                modules_to_cache = \
+                    self._get_modules_to_cache(
+                        group.functions.items(),
+                        group_name,
+                        self.config.snapshot_second,
+                        self.MINIMAL_CACHE_FREQ
+                    )
+            else:
+                modules_to_cache = set()
+
+            for fun, old_fun_desc in sorted(group.functions.items()):
+
+                # Check if the function exists in the other snapshot
+                new_fun_desc = \
+                    self.config.snapshot_second.get_by_name(fun, group_name)
+                if not new_fun_desc:
+                    continue
+
+                # Check if the module exists in both snapshots
+                if old_fun_desc.mod is None or new_fun_desc.mod is None:
+                    fun_result = Result(Result.Kind.UNKNOWN, fun, fun)
+                    self.result.add_inner(fun_result)
+                    group_printed = self._print_fun_result(fun_result, fun,
+                                                           group_dir,
+                                                           group_name,
+                                                           old_fun_desc,
+                                                           group_printed)
+                    continue
+
+                # If function has a global variable, set it
+                glob_var = LlvmParam(old_fun_desc.glob_var) \
+                    if old_fun_desc.glob_var else None
+
+                # Run the semantic diff
+                fun_result = functions_diff(
+                    mod_first=old_fun_desc.mod, mod_second=new_fun_desc.mod,
+                    fun_first=fun, fun_second=fun,
+                    glob_var=glob_var, config=self.config,
+                    prev_result_graph=result_graph, function_cache=cache,
+                    module_cache=module_cache,
+                    modules_to_cache=modules_to_cache)
+                result_graph = fun_result.graph
+
+                if fun_result is not None:
+                    if self.args.regex_filter is not None:
+                        # Filter results by regex
+                        self._filter_result_by_regex(self.regex_pattern,
+                                                     fun_result)
+
+                    self.result.add_inner(fun_result)
+
+                    # Printing information about failures and
+                    # non-equal functions.
+                    if fun_result.kind in [Result.Kind.NOT_EQUAL,
+                                           Result.Kind.UNKNOWN,
+                                           Result.Kind.ERROR] \
+                       or self.config.full_diff:
+                        group_printed = self._print_fun_result(fun_result, fun,
+                                                               group_dir,
+                                                               group_name,
+                                                               old_fun_desc,
+                                                               group_printed)
+                # Clean LLVM modules (allow GC to collect the occupied memory)
+                old_fun_desc.mod.clean_module()
+                new_fun_desc.mod.clean_module()
+                LlvmModule.clean_all()
+
+        self.result.graph = result_graph
+        # Create yaml output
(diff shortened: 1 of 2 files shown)


PLukas2018:
For now, I have a note on the found bug. I will add notes for the rest of the code probably later today.

---

Good catch with the extended stats problem :+1:.

I think it would be better to fix the problem in another way. The `result.graph` is updated when calling `result.add_inner` method with the `fun_result`. I think that the problem is that we sometimes call the method with `fun_result`, which does not have a `graph` -- e.g.
https://github.com/diffkemp/diffkemp/blob/885b3b8efd77486afb7eb9c628ab7afc6e5f7d9d/diffkemp/compare.py#L67-L68
The total `graph` is, in this case, updated to `None`.

I think that it should be enough to add a check that `graph is not None` when updating the total graph in:
https://github.com/diffkemp/diffkemp/blob/885b3b8efd77486afb7eb9c628ab7afc6e5f7d9d/diffkemp/semdiff/result.py#L201-L213

 @Bal0sk, do you think that you could try to fix it in that way, test it if it works, and create a new PR just for the fix (leave in this PR just the `compare` function refactorization)?

diffkemp/compare.py
```diff
@@ -19,225 +18,413 @@ class OutputDirExistsError(Exception):
...
+            # Print the group name if needed
+            if group_name is not None and not group_printed:
+                print("{}:".format(group_name))
+                group_printed = True
+            print("{}: {}".format(fun, str(fun_result.kind)))
+            return group_printed
+
+        # Create the output directory if needed
+        if self.output_dir is not None:
+            self._ensure_dir_exists(self.output_dir)
```
PLukas2018, line 245: I would maybe inline the method in here:
```suggestion
            os.makedirs(path, exist_ok=True)
```

diffkemp/compare.py
```diff
@@ -19,225 +18,413 @@ class OutputDirExistsError(Exception):
...
+
+        # Create the group directory or print the group name
+        # if needed
+        if group_dir is not None:
+            self._ensure_dir_exists(group_dir)
```
PLukas2018, line 250: The same as above.

diffkemp/compare.py
```diff
@@ -19,225 +18,413 @@ class OutputDirExistsError(Exception):
...
+        new_dir_abs_path = os.path.join(
+            os.path.abspath(self.args.snapshot_dir_new), "")
+
+        if fun_result.kind == Result.Kind.NOT_EQUAL or (
+                self.config.full_diff and
+                any([x.diff for x in fun_result.inner.values()])):
+
+            # sets the output and indent in the writer class
+            self.writer._open_output_file_set_indent(
+                output_dir, fun, fun_tag, initial_indent)
```
PLukas2018, line 292: I would create a new `OutputWriter` instance and pass the arguments to the constructor.

diffkemp/compare.py
```diff
@@ -19,225 +18,413 @@ class OutputDirExistsError(Exception):
...
+
+        @staticmethod
+        def _text_indent(text, width):
+            """
+            Indent each line in the text by a number of spaces given by width
+            """
+            return ''.join(" "*width + line for line in text.splitlines(True))
+
+        def _open_output_file_set_indent(self, output_dir, fun,
+                                         fun_tag, initial_indent):
```
PLukas2018, line 346: I would move this to the constructor (`__init__` method).

PLukas2018 (changes requested):
Thanks for the refactorization. I think we are on the right track, but I think it will need some adjustments -- mainly, I think we could also create a `GroupComparator` class and move some methods into it so we do not pass too many arguments. There are also other changes I think would be good to make -- see below.

---

> I find that the `compare.py` file is a little long now, after adding the `print_syntax_diff` and adding additional class for output writing.

I do not think the file is too long, I would probably keep the `print_syntax_diff` here, maybe the `OutputWriter` class would make sense to move to `diffkemp/output.py`?

diffkemp/compare.py
PLukas2018: I did not realize that dir can be `None` (for stdout output), in that case it probably makes sense to leave it in the method.

Bal0sk:
Implemented notes into history, created the GroupComparator class, as suggested. For now, I moved the OutputWriter class to `output.py`.

diffkemp/compare.py
```diff
@@ -81,163 +166,152 @@ def compare(args):
...
+        @staticmethod
+        def _modules_exist(old_fun_desc, new_fun_desc):
+            return old_fun_desc.mod is not None and \
+                   new_fun_desc.mod is not None
+
+        def _handle_missing_module(self, fun, old_fun_desc):
+            fun_result = Result(Result.Kind.UNKNOWN, fun, fun)
+            self.result.add_inner(fun_result)
+            self._print_fun_result(fun_result, fun, old_fun_desc)
+            return
```
PLukas2018, line 188: Same as above, no need for `return`.

diffkemp/compare.py
```diff
@@ -81,163 +166,152 @@ def compare(args):
...
+                initial_indent=2 if (self.group_name is not None and
+                                     self.group_dir is None) else 0)
+
+            return
+
+        def print_syntax_diff(self, fun, fun_result, fun_tag, output_dir,
+                              initial_indent):
+            """
+            Log syntax diff of 2 functions. If log_files is set, the output is
+            printed into a separate file, otherwise it goes to stdout.
```
PLukas2018, line 250: Please update the comments (there is no `log_files` if I am not mistaken).

diffkemp/output.py
```diff
@@ -143,3 +144,92 @@ def _create_def_info(self, line, file, snapshot_dir, kind):
...
+                                               initial_indent))
+            indent = initial_indent + 4
+
+        self.output = output
+        self.indent = indent
+
+    def write_called_result(self, called_res,
+                            show_diff, snapshot_dir_old,
+                            snapshot_dir_new, old_dir_abs_path,
+                            new_dir_abs_path, output_dir):
```
PLukas2018, line 194: Also, it would be good if we could use only `old_dir_abs_path`/`new_dir_abs_path` or `snapshot_dir_old`/`snapshot_dir_new`, which would simplify the class even more. But maybe we can leave it for another PR because I think it would need changes also in the `Result.Callstack` class.

PLukas2018, line 194: If you set the `output_dir` as an attribute of the class in the constructor, you will not need to pass it here.
Additionally, it might make sense to pass here only the `called_res` and the rest when creating the class using the constructor, as the arguments are the same for all called functions' results.

diffkemp/compare.py
```diff
@@ -81,163 +166,152 @@ def compare(args):
...
+        def _handle_fun_result(self, fun_result, fun, old_fun_desc):
+
+            if self.args.regex_filter is not None:
+                # Filter results by regex
+                self._filter_result_by_regex(fun_result)
+
+            self.result.add_inner(fun_result)
+
+            # Printing information about failures and
+            # non-equal functions.
```
PLukas2018, line 199: I think it would be better if the wrapping of the code/comments
- was removed
- or wrapped later

if possible (check other locations).
 


```suggestion
            # Printing information about failures and non-equal functions.
```

diffkemp/compare.py
```diff
@@ -81,163 +166,152 @@ def compare(args):
...
+            self.print_syntax_diff(
+                fun=fun,
+                fun_result=fun_result,
+                fun_tag=old_fun_desc.tag,
+                output_dir=self.group_dir if self.group_dir else
+                self.output_dir,
+                initial_indent=2 if (self.group_name is not None and
+                                     self.group_dir is None) else 0)
+
+            return
```
PLukas2018, line 244: Same as above.

diffkemp/compare.py
```diff
@@ -81,163 +166,152 @@ def compare(args):
...
+                self.group_printed = True
+
+            self.print_syntax_diff(
+                fun=fun,
+                fun_result=fun_result,
+                fun_tag=old_fun_desc.tag,
+                output_dir=self.group_dir if self.group_dir else
+                self.output_dir,
+                initial_indent=2 if (self.group_name is not None and
+                                     self.group_dir is None) else 0)
```
PLukas2018, line 242: This could be maybe done at the top of the `print_syntax_dif` method so we reduce the number of parameters.

PLukas2018 (changes requested):
@Bal0sk, thanks for the changes 👍 . I went over the code again, and I have a few more notes. 

Also, I am not sure what the best way is of committing the changes (I went over the file changes rather than the individual commits). Maybe it will make sense to squash it all at the end into one commit, but right now, probably leave it as you have it.

diffkemp/compare.py
```diff
@@ -19,225 +18,413 @@ class OutputDirExistsError(Exception):
...
+            dirname = base_dirname
+            while os.path.isdir(dirname):
+                dirname = "{}-{}".format(base_dirname, suffix)
+                suffix += 1
+            return dirname
+        return base_dirname
+
+    def _compare_snapshots(self):
+        for group_name, group in sorted(self.config.snapshot_first.
+                                        fun_groups.items()):
```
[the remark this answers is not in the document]
viktormalik: Not a big fan of another class. Moreover, it looks to me that a number of members of `GroupComparator` can be moved to `SnapshotComparator` (which doesn't contain much besides `config` atm). See below for details.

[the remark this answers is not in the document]
PLukas2018: The reason why I suggested creating the `GroupComparator` class was to reduce the number of parameters passed down to the following methods. Another way, which comes to my mind, how we could reduce the amount of parameters is to create a class to wrap the parameters connected with the *group* (`group`, `group_name`, `group_printed`, ...) and leave the methods in the `SnapshotComparator` class and pass the instance of the class to the methods.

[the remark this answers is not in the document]
PLukas2018: One more option that comes to mind, which would not require any additional classes, is to make the variables related to *group* (`group_name`, `group_printed`, ...) as `SnapshotComparator` attributes. This way, all methods that use the variables have easy access to them, and for each *group* we would overwrite them.

PLukas2018 (changes requested):
So I tested the PR and found only one problem (see below), otherwise LGTM.

---
Follows detailed description of tested cases, just to track/inform what was tested:

#### Tested cases for function comparison (RHEL 8.0-8.1):
- **Printing results to stdout**:
  - `diffkemp compare --stdout  ...`
  - `diffkemp compare --stdout --report-stat --extended-stat ...`
- **Saving result to directory**: 
  - `diffkemp compare -o out --report-stat --extended-stat ...`
-  **Comparing only one function**:
  - `diffkemp compare -f __alloc_pages_nodemask ...`
- **Using regex filter** 
  - `diffkemp compare -f __alloc_pages_nodemask --stdout --regex-filter "20"`
  - `diffkemp compare -f __alloc_pages_nodemask --stdout --regex-filter "25"`

#### Tested cases for *sysctl* comparison (RHEL 8.0-8.1):
- **Printing results to stdout**:
  -  `diffkemp compare --stdout`
- **Saving result to dir + using stats**
  - `diffkemp compare -o out --report-stat --extended-stat`

diffkemp/compare.py
```diff
@@ -1,243 +1,301 @@
...
+        self.group = group
+        self.group_name = group_name
+        self._set_group_dir()
+        self.group_printed = False
+        self._set_modules_to_cache_if_enabled()
+        self.result_graph = None
+        self.cache = SimpLLCache(mkdtemp())
+        self.module_cache = {}
+
+    def _set_group_dir(self):
```
viktormalik: I think it would be much nicer if this returned the group dir instead of setting it directly (so it would be something like `_get_group_dir`) and above, we would do:
```
self.group_dir = self._get_group_dir()
```

diffkemp/compare.py
```diff
@@ -1,243 +1,301 @@
...
+        if self.output_dir is not None and self.group_name is not None:
+            self.group_dir = os.path.join(self.output_dir, self.group_name)
         else:
-            modules_to_cache = set()
-
-        for fun, old_fun_desc in sorted(group.functions.items()):
+            self.group_dir = None
 
+    def _set_modules_to_cache_if_enabled(self):
```
viktormalik: Same as above with `_set_group_dir`.

viktormalik (approved):
LGTM, thanks for the large amount of work!

viktormalik:
> Thank you very much for this PR.
> 
> According to the [bot](https://github.com/PLukas2018/diffkemp/pull/13#issuecomment-3427102071) the results are the same as before this PR, so hopefully it is ok.
> 
> @viktormalik should I test the PR again also manually (the cases that are not checked by the bot)?

No, I think we're good. Let's merge this.

PR #414, opened 2025-09-25
Bug fix: result.graph updating to None

Bal0sk:
This fixes an issue mentioned in #413. Tested on kernels from `examples`.

### DIFF
--- a/diffkemp/semdiff/result.py
+++ b/diffkemp/semdiff/result.py
@@ -210,7 +210,9 @@ class Result:
         # The graph of the latest inner result is the graph of the outer one.
         # Note: this is true because the graph is built incrementally, reusing
         # the already known results from the previous comparison.
-        self.graph = result.graph
+        # Update the total graph only if the entity changed it (is not None)
+        if result.graph is not None:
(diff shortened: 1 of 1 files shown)


PLukas2018:
Thanks. Could you add a commit message describing why it was necessary, and maybe also a comment in the code?

I will run this PR with our bot to see if all comparison succeeds. I am not sure how we did not discover this problem earlier...

viktormalik (approved):
Good catch! Thanks for the fix, LGTM.

I'll wait for @PLukas2018 approval since he has more context about this than I do.

PLukas2018 (approved):
So the bot (which uses `--extended-stats` option for the extraction of some information) did not crash, but it did not crash either before on the `master` branch, so I feel like the problem occurs only sometimes (not sure exactly why). I attempted to replicate the bug myself on the `master` branch when @Bal0sk reported it, and I encountered the same problem he reported. I tried to replicate it again on the `master` branch, and no problem occurred. Not sure why.

---

**Anyway, I think that this PR should fix the problem from appearing again.**

PR #418, opened 2025-11-10
Rewrite cc_wrapper to C++

### LINKED ISSUE #400: Rewrite `cc_wrapper` in C/C++
The [cc_wrapper](https://github.com/diffkemp/diffkemp/tree/master/diffkemp/building/cc_wrapper) component is used for building generic projects into snapshots (using `diffkemp build`). It acts as a wrapper over a C compiler which hijacks each compile command of a project build and injects a build into LLVM IR.

It is currently written in Python (for simplicity) and compiled into a native binary (for performance) using RPython. Unfortunately, RPython only supports Python 2 which is now deprecated on most distros and therefore hard to support. While there are Python 3 alternatives, none of them is performing enough so it would be better to rewrite the wrapper into C or C++ (probably C++ would be better since the wrapper does a lot of string manipulation).

Bal0sk:
Fixes #400 

This PR rewrites the `cc_wrapper` component from Python to C++.

The RPython only supports Python2, which is deprecated, making the wrapper difficult to support. This C++ rewrite ensures long-term maintainability and performance while removing the RPython and Python2 dependency.

### Changes

* Rewrites the `cc_wrapper` core logic
* Includes minor changes to error message output
* Removes environment variables, which were used to pass arguments, now passed directly as arguments

Code was tested with the make-based projects from the `Examples`.


diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    c_args.push_back(const_cast<char *>(file.c_str()));
+
+    for (const std::string &arg : args) {
+        c_args.push_back(const_cast<char *>(arg.c_str()));
+    }
+
+    c_args.push_back(NULL);
+
+    pid_t pid = fork();
```
PLukas2018: Is it necessary to fork??

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,234 @@
...
+    // Run GCC
+    std::vector<std::string> gcc_args;
+    gcc_args.reserve(argv_vec.size() + 1);
+    for (std::string arg : argv_vec) {
+        gcc_args.push_back(arg);
+    }
+
+    try {
+        check_call("gcc", gcc_args);
+    } catch (const std::exception &e) {
```
PLukas2018: Are not the exceptions too much heavy??

tests/unit_tests/build_test.py
```diff
@@ -21,6 +21,7 @@ def __init__(self, source_dir, output_dir, target=[],
         self.no_opt_override = no_opt_override
         # Required/used by build_c_project
         self.symbol_list = None
+        self.verbose = 0
```
PLukas2018: Is this necessary?

diffkemp/building/cc_wrapper/CMakeLists.txt
```diff
@@ -1,33 +1,22 @@
-set(CC_WRAPPER_SOURCE "${CMAKE_SOURCE_DIR}/diffkemp/building/cc_wrapper/cc_wrapper.py")
-add_custom_target(cc-wrapper-source ALL DEPENDS ${CC_WRAPPER_SOURCE})
-install(PROGRAMS ${CC_WRAPPER_SOURCE} RENAME diffkemp-cc-wrapper.py DESTINATION ${CMAKE_INSTALL_BINDIR})
+# Define the sources for the new C++ compiler wrapper
+set(CC_WRAPPER_SOURCES
+    "${CMAKE_SOURCE_DIR}/diffkemp/building/cc_wrapper/cc_wrapper.cpp"
+    "${CMAKE_SOURCE_DIR}/diffkemp/building/cc_wrapper/cc_wrapper_utils.cpp"
+    "${CMAKE_SOURCE_DIR}/diffkemp/building/cc_wrapper/cc_wrapper_utils.h"
```
PLukas2018: This could probably be simplified by using `CMAKE_CURRENT_SOURCE_DIR` var instead of `CMAKE_SOURCE_DIR`

diffkemp/building/build_c_project.py
```diff
@@ -65,35 +64,11 @@ def build_c_project(args):
...
-
-
-def _make_args(args):
+def _make_args(args, db_filename):
     # Generate wrapper for C/C++ compiler
-    cc_wrapper = get_cc_wrapper_path(args.no_native_cc_wrapper)
+    cc_wrapper, ccw_options = get_cc_wrapper_path_w_opts(args, db_filename)
 
-    make_cc_setting = 'CC="{}"'.format(cc_wrapper)
+    make_cc_setting = f'CC={cc_wrapper}{ccw_options}'
```
PLukas2018:
```suggestion
    make_cc_setting = f"CC={cc_wrapper}{ccw_options}"
```

Bal0sk:
Thanks for the replies and clarification, I will make changes accordingly. Looking back at the CI logs, `tools/check-clang-format.sh` runs on the wrapper, however, since the only argument set is -d, it did not check for the preferred style.

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,245 @@
...
+    // Run clang
+    try {
+        checkCall(clangBin, clangArgv);
+    } catch (const std::exception &e) {
+        std::cerr << "cc_wrapper: warning: clang failed" << e.what()
+                  << std::endl;
+        return 0;
+    }
+
+    // Update databse file
```
viktormalik:
```suggestion
    // Update the database file
```

diffkemp/building/cc_wrapper/cc_wrapper_utils.cpp
```diff
@@ -0,0 +1,98 @@
...
+    return "";
+}
+
+bool isInVector(const std::string &s, const std::vector<std::string> &vector) {
+    for (const std::string &val : vector) {
+        if (val == s) {
+            return true;
+        }
+    }
+    return false;
```
viktormalik: This could be implemented by
```suggestion
    vector.find(s) != vector.end()
```
and probably inlined directly into the code.

viktormalik (commented):
Thanks for the PR! Couple of comments, mainly wrt. C++ programming style.

PLukas2018 (commented):
Thanks for the PR.

Below are some notes on what I noticed. 

Also nit. the second commit message title could be more concrete (for what change?).

Note: Personally, I have not tested this PR yet.

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,245 @@
...
+        clangArgv.insert(clangArgv.end(), append.begin(), append.end());
+        // TODO: allow compiling into binary IR
+    } else {
+        // Keep only arguments with input files (and llvm-link itself)
+        std::vector<std::string> tempArgs;
+        for (const auto &arg : clangArgv) {
+            if (endsWith(arg, ".ll") || endsWith(arg, ".llw") || arg == "-o") {
+                tempArgs.push_back(arg);
+            }
+        }
```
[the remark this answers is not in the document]
DanielKriz: I agree with the ranges style; however, DiffKemp is currently using C++17 and the CMake for the CC wrapper does not specify the standard version. I would fully support switch to C++20, but it should probably be standalone PR.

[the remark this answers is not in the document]
viktormalik: Oh, I thought we were on C++ 20 already. Let's use the original code, then, and we can switch to C++20 later.

diffkemp/building/cc_wrapper/cc_wrapper_utils.h
```diff
@@ -0,0 +1,42 @@
...
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+// Utility functions for the cc_wrapper
+
+std::vector<std::string> split(const std::string &s, char delim);
+bool endsWith(const std::string &str, const std::string &suffix);
+std::vector<std::string> getClangDefaultOptions(bool defaultOptim = false);
+bool isExecutable(const std::string &path);
```
DanielKriz: This can be marked `noexcept`.

DanielKriz (changes requested):
First things first, thank you for your PR, the base of it in a really good shape.

Most of my comments are small nits as guys have found all the other issues. What I would like to see are some unit and regression tests (I am aware that usage of this wrapper in kernel tests is kind of a regression test, but we are not doing kernel test for each LLVM version), so that we are not afraid to make changes to this in the future.

viktormalik:
> And from `flake.nix` we should now be able to remove mentions of `WITHOUT_RPYTHON`.

Correct. We'll also need to update `diffkemp.spec` for packaging but that can be done when we release a new version.

diffkemp/building/build_c_project.py
PLukas2018: Nit. `no-native-cc-wrapper` should be removed.

PLukas2018 (commented):
Thanks for the changes.
I have just 2 nits. (I will leave the Cpp part rather to review someone else, as I do not have too much experience with Cpp)
I have not tested the changes yet, but I am planning to do so.

Also, I am curious if you tried some experiments with it and if so what were the results?
If you did not, I do not mind as I think this is still better than if we stick with the RPython solution.

diffkemp/building/cc_wrapper/cc_wrapper_utils.h
```diff
@@ -0,0 +1,32 @@
...
+struct WrapperArgs {
+    std::string dbFilename;
+    std::string clang;
+    std::vector<std::string> clangAppend;
+    std::unordered_set<std::string> clangDrop;
+    bool debug = false;
+    std::string llvmLink;
+    std::string llvmDis;
+    bool noOptOverride = false;
+    std::vector<std::string> clangArgs;
```
viktormalik: These are the arguments of the original (typically GCC) command, right? And they are not just use for Clang but also for the GCC call so the name is not the best. How about we call it `cc_args` or (respecting Makefile conventions) `cflags` or `orig_cflags`?

viktormalik (commented):
Couple of comments, otherwise looks very nice!

diffkemp/building/build_c_project.py
PLukas2018: This comment was meant to be to a line in
https://github.com/Bal0sk/diffkemp/blob/b92054bc414d682ac1eb2a1eb3d78928b32b0be4/diffkemp/building/build_c_project.py#L152

viktormalik:
> > Please, let me know what you think.
> 
> Personally, I would leave solving the logging for another PR so we do not prolong merging this, if it does not need to be solved right now. Right now I would leave the logging just for the SimpLL test binary (if it is possible). You can create for it issue and we can get back to it later.

Agreed, let's not add more content to this PR than necessary.

Bal0sk:
Thank you for the feedback.
I implemented the requested changes

Also tested the speed of different versions of the wrapper to see if it would improve the time. Tests show a marginal decrease in time compared to RPython and python:
- "pure" python: ~109.95s user 58.28s system 100% cpu 2:46.82 total (base)
- RPython: ~86.88s user 60.86s system 100% cpu 2:26.55 total (around 11.7% decrease from python)
- C++ rewrite: ~83.23s user 47.46s system 101% cpu 2:09.16 total (around 11.6% decrease compared to RPython and 22.1% compared to python)

These tests were done on the musl-1.2.3 library.

docs/development.md
```diff
@@ -222,6 +226,11 @@ the following command:
 SIMPLL_VERBOSITY=2 ninja -C build test
 ```
```
PLukas2018, line 229: I think this won't work now, because of missing `--output-on-failure` param in `CMakeLists.txt` (see below). Maybe it should contain a path to the `simpllTests` binary instead.

docs/development.md
```diff
@@ -197,14 +197,18 @@ also be used to download and configure the aforementioned kernels.
...
 
-### Tests for the SimpLL library
-
-Tests are located in `tests/unit_tests/simpll/` directory and they can be run
-by:
+### C++ tests
 
+All C++ tests can be run by the following command:
 ```sh
 ninja -C build test
```
PLukas2018, line 204: Maybe we should also mention (in individual subsections) that the tests can be run separately (*simpll* x *cc_wrapper*) using the tests binaries and where they are located.

tests/unit_tests/simpll/CMakeLists.txt
```diff
@@ -44,18 +28,16 @@ if (NOT ${Z3_FOUND})
...
 if (NOT ${system_libs} STREQUAL "")
-	target_link_libraries(runTests gtest ${system_libs})
+	target_link_libraries(simpllTests gtest ${system_libs})
 endif()
 
 include(GoogleTest)
-gtest_discover_tests(runTests)
-add_custom_target(test
-  COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure
-  DEPENDS runTests)
```
PLukas2018, line 58: Note: IIUC removing the `DEPENDS ...` param will mean that when running `ninja -C build test`, the command will not automatically rebuild the tests if a change was made in tests, and it is necessary to run firstly:
```
ninja -C build
```
to rebuild the tests, and then
```
ninja -C build test
```
to run the tests.

tests/unit_tests/simpll/CMakeLists.txt
```diff
@@ -44,18 +28,16 @@ if (NOT ${Z3_FOUND})
...
+
 if (NOT ${system_libs} STREQUAL "")
-	target_link_libraries(runTests gtest ${system_libs})
+	target_link_libraries(simpllTests gtest ${system_libs})
 endif()
 
 include(GoogleTest)
-gtest_discover_tests(runTests)
-add_custom_target(test
-  COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure
```
PLukas2018, line 57: IIUC By removing the `--output-on-failure` param, the `ninja -C build test` will contain only which tests are passing and which are not, but it will not contain why the failed tests are not passing.

E.g. this
```
66/85 Test #66: DFCSmtTest.SmtReorderedThreeInst .......................................***Failed    0.03 sec
      Start 67: DFCSmtTest.SmtFloatingPointSimple
...
The following tests FAILED:
	 66 - DFCSmtTest.SmtReorderedThreeInst (Failed)
Errors while running CTest
Output from these tests are in: /home/lukas/ssd/diffkemp-org/diffkemp/build-nix-19/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.

```

instead of this:
```
64/83 Test #64: DFCSmtTest.SmtReorderedThreeInst .......................................***Failed    0.03 sec
Note: Google Test filter = DFCSmtTest.SmtReorderedThreeInst
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from DFCSmtTest
[ RUN      ] DFCSmtTest.SmtReorderedThreeInst
/home/lukas/ssd/diffkemp-org/diffkemp/tests/unit_tests/simpll/DFCLlvmIrTest.cpp:1200: Failure
Expected equality of these values:
  DiffComp->compare()
    Which is: 0
  1

[  FAILED  ] DFCSmtTest.SmtReorderedThreeInst (9 ms)
[----------] 1 test from DFCSmtTest (9 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (9 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] DFCSmtTest.SmtReorderedThreeInst

 1 FAILED TEST
```

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,246 @@
...
+ * dbf   : <string> Path to the output snapshot database (db_filename).
+ * clang : <string> Path to the clang binary.
+ * cla   : vector<string> "Clang Append" - extra flags to pass to clang.
+ * cld   : unordered_set<string> "Clang Drop" - flags to remove from the clang
+ * invocation.
+ * debug : <bool> Enable debug logging if present.
+ * llink : <string> Path to the llvm-link binary.
+ * lldis : <string> Path to the llvm-dis.
+ * binary.
+ * noo   : <bool> "No Opt Override" - do not override optimization levels.
```
PLukas2018, line 14: Nit. I am not sure if the types (e.g., `vector<string>`) make sense in here, if the args should represent arguments which can be provided to the `cc_wrapper` binary...

docs/usage.md
```diff
@@ -40,15 +40,19 @@ In this case, the path to the file should be given in place of `PROJ_DIR`.
...
 - `SYMBOL_LIST`: Path to a file containing a list of symbols (each symbol on
    a single line) which should be prepared for comparison.
+- `--reconfigure`: Reconfigures autotools-based project with `CC=<DiffKemp
+  compiler wrapper>`.
+- `--target TARGET`: Allows specifying `Makefile` targets which should be used
+  to build the snapshot from the project.
+- `--build-program BUILD_PROGRAM`: `make` tool to be used for building
+  (default `make`).
+- `--build-file BUILD_FILE`: Filename of the project's `Makefile` to be used
+  for the build.
```
PLukas2018, line 50: Nit. If the text is the same, it could be left as it was... but the reorder probably makes sense to group the options that are passed down to the `cc_wrapper`...

PLukas2018 (approved):
Hi. Thanks for the changes and the info about the performance. Also, I am sorry for the late review.

The changes look mostly good to me. There are a few notes provided below, but I think it can be left as is. (Maybe having `--output-on-failure` would be good for quick debugging using the CI log, but it can be returned back in a different PR).

I have also tested the PR -- I have compared `musl-1.2.3` and `musl-1.2.4` using DiffKemp on the master branch and the PR branch. I have compared the snapshots and comparison outputs, and they seem similar to me, so I think the `cc_wrapper` logic remained the same. Good job 👍 .

I have also tried using the build command (`cc_wrapper`) when DiffKemp is installed, and the wrapper worked 👍 . 

Overall, LGTM.

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,246 @@
...
+#include <sys/wait.h>
+#include <unistd.h>
+#include <utility>
+#include <vector>
+
+namespace fs = std::filesystem;
+
+int checkCall(const std::string &cmd, const std::vector<std::string> &args) {
+    std::vector<char *> cArgs;
+    cArgs.reserve(args.size() + 2);
```
DanielKriz, line 39: Nit. The magic constant 2 here is a bit confusing, it can be seen in the code later, but on the first glance it adds confusion.

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+    if (pid == -1) {
+        std::cerr << "fork failed" << "\n";
+        return -1;
+    }
+
+    if (pid == 0) {
+        execvp(cmd.c_str(), cArgs.data());
+        std::cerr << "execvp failed for: " << cmd << "\n";
+        _exit(127);
```
DanielKriz, line 60: I know that this code might be standardized somewhere, but I would like to have some comment or constant here that explains why are we using this particular value)

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    int status;
+    pid_t waitResult = waitpid(pid, &status, 0);
+
+    if (waitResult == -1) {
+        return -1;
+    }
```
DanielKriz: The `waitResult` is not used anywhere else, so we don't have to declare it as a variable, we can simply write:
```cpp
if (waitpid(pid, &status, 0) == -1) {
    return -1;
}
```

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,246 @@
...
+    if (WIFEXITED(status)) {
+        int exitStatus = WEXITSTATUS(status);
+        if (exitStatus != 0) {
+            return exitStatus;
+        }
+    } else {
+        return 1;
+    }
+
+    return 0;
```
DanielKriz: This could be simplified to:
```cpp
if (!WIFEXITED(status)) {
    return EXIT_FAILURE;
}

return WEXITSTATUS(status);
```
where `EXIT_FAILURE` is equal to `1`.

I may be mistaken, but it seems that the return 0 is reachable only if `exitStatus` is equal to 0. So we have to call the `WEXITSTATUS` anyway if we would reach the line 76.

Please, check if I am correct.

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,246 @@
...
+    WrapperArgs wrapperArgs = WrapperArgs(argc, argv);
+
+    // Run GCC
+    int status = checkCall("gcc", wrapperArgs.cFlags);
+    if (status != 0) {
+        std::cerr << "cc_wrapper warning: original build command failed with "
+                     "code "
+                  << status << "\n";
+        return status;
+    }
```
DanielKriz: This is a great place to use the init-statement in the if-clause which would reduce the scope of status (which could be made `const` on top of that):
```cpp
if (const int status = checkCall(...); status != 0) { ... }
```

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,246 @@
...
+    // Run clang
+    status = checkCall(clangBin, clangArgv);
+    if (status != 0) {
+        if (wrapperArgs.debug) {
+            std::cerr << "cc_wrapper warning: clang failed with code " << status
+                      << "\n";
+        }
+        // clang error is non-fatal
+        return 0;
+    }
```
DanielKriz: The same thing with the init statement in an if.

diffkemp/building/cc_wrapper/cc_wrapper.cpp
```diff
@@ -0,0 +1,246 @@
...
+            if (fs::exists(path)) {
+                db << entry << "\n";
+            }
+        }
+    } else {
+        std::cerr << "cc_wrapper warning: cannot open DB file for append:"
+                  << wrapperArgs.dbFilename << "\n";
+    }
+
+    return 0;
```
DanielKriz: Nit. We could use `EXIT_SUCCESS` here.

diffkemp/building/cc_wrapper/CMakeLists.txt
```diff
@@ -1,33 +1,16 @@
-set(CC_WRAPPER_SOURCE "${CMAKE_SOURCE_DIR}/diffkemp/building/cc_wrapper/cc_wrapper.py")
-add_custom_target(cc-wrapper-source ALL DEPENDS ${CC_WRAPPER_SOURCE})
-install(PROGRAMS ${CC_WRAPPER_SOURCE} RENAME diffkemp-cc-wrapper.py DESTINATION ${CMAKE_INSTALL_BINDIR})
+# Define the sources for the new C++ compiler wrapper
+set(CC_WRAPPER_SOURCES
```
DanielKriz, line 2: In other parts of the project (SimpLL) we are using lowercase letters for user variables.

DanielKriz (changes requested):
First of all, thank you for incomporating the changes. The results are looking good and I am especially happy that you have added some tests.

I have found some small things that shouldn't affect the functionality of the solution. They are just small changes regarding the code quality. But it was really hard to find them, good job!

One nitpick that might hurt the git history is the fact, that we are using camelCase for C++ in SimpLL and the wrapper is using snake_case...

Perhaps my comments are not worth delaying this project. What does @viktormalik and @PLukas2018 think?

viktormalik (approved):

viktormalik:
Looks good! Let's get this merged, @DanielKriz could you please have a quick glance and unblock? Thanks!

DanielKriz (approved):
LGTM. Good job, it is looking nice!

viktormalik:
Merged. Thanks for the work @Bal0sk!

Issue #406, opened 2025-09-16
`diffkemp-out.yaml`: Add missing definitions for functions when comparing `sysctl` parameters


PLukas2018:
Currently, when comparing `sysctl` parameters, the `diffkemp-out.yaml` does not contain definitions of all functions. This problem is connected with #403 and is connected with the aggregation of graph.

TODO @PLukas2018 describe more thoroughly

PLukas2018:
This is also a problem for the combination of global variables and function comparison (#432).

The problem is that the graph in the overall result contains the graph of the last group (this is intentional, and changing this would probably pollute the results of other subgroups).

I think to fix this, instead of searching definitions in the overall `result.graph`, they would need to be searched in individual inner `result.graph`. This seems to unnecessarily complicate things, so probably moving the definitions information out of the comparison graph to a separate entity would be probably better. Alternatives to consider:
- Saving the definitions separately even in SimpLL and passing them down
- Not extracting definitions at all and doing this, e.g., in the viewer or at the end of comparison by using e.g. `cscope`  (the problem can be that there can be multiple locations of the same function... and the chosen one depends on preprocesor/macros/... -- DiffKemp knows which function it used).

PR #421, opened 2025-11-26
Add info about sysctl parameters

### LINKED ISSUE #403: `diffkemp-out.yaml`: Add info about differing `sysctl` parameters
When comparing the [*sysctl* parameters](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/managing_monitoring_and_updating_the_kernel/configuring-kernel-parameters-at-runtime_managing-monitoring-and-updating-the-kernel), the `diffkemp-out.yaml` file currently contains only information about in which functions are found semantic differences, but not which/how the functions are connected with the *sysctl* parameters. 

So e.g., for `sysctl-list` file containing the following *sysctl* groups:
```txt
kernel.*                                                                        
vm.*
```
when running following commands:
```bash
build/bin/diffkemp build-kernel kernel/linux-4.18.0-80.el8/ snap-80-sysctl sysctl-list --sysctl
build/bin/diffkemp build-kernel kernel/linux-4.18.0-147.el8/ snap-147-sysctl sysctl-list --sysctl
build/bin/diffkemp compare snap-80-sysctl/ snap-147-sysctl/
```
The `diffkemp-out.yaml` looks like [this](https://github.com/user-attachments/files/22359882/diffkemp-out.yaml).

It would be beneficial if the file also included information about to which *sysctl parameter* the function belongs to.

The updated format of `diffkemp-out.yaml` remains to be discussed, but it could theoretically look like:
1. ```yaml
    results:
    - function:
      diffs:
      - function:
        old-callstack:
        - name: ... 
          file: ... 
          line: ... 
        - # ...
        new-callstack:
        - # ...
      - # ...
      sysctl: # Name of sysctl parameter  <------------------------ Add
    - # ...
    ```
    In the case of the *sysctl parameter* comparison, there would be an added `sysctl` field to all functions containing the name of the compared *sysctl parameter* (e.g., `sysctl: kernel.acct`) based on which the function was compared.

2. An alternative solution would be:
    ```yaml
    results:
    - sysctl: # Name of sysctl parameter  <------------------------ Add
      results: # Differing functions for given parameter (indented inside the `sysctl` group) <--------- Change
      - function:
        diffs:
        - function:
          old-callstack:
          - name: ... 
            file: ... 
            line: ... 
          - # ...
          new-callstack:
          - # ...
        - # ...
      - # ...
    - sysctl: # Name of different sysctl parameter
      results: .... #  Differing functions for given parameter 
    ```
    In the case of the *sysctl parameter* comparison, the format would be different; the results of individual functions would be inserted within the results for the given *sysctl parameter*.

3. Different approach.

It would be good to also add a `kind` field (or something similar) to the output file, so we can easily distinguish if the results are of *function* comparison or *sysctl parameters* comparison when the user (or we in the viewer) parses them.

--- 

An approximate list of things that will need to be updated:
- Update [`Results` class](https://github.com/diffkemp/diffkemp/blob/8912507d38d3a9591ee55f00a6d7524b204d0255/diffkemp/semdiff/result.py#L6-L340): Currently we are using it to represent the result of
  - the overall project,
  - compared functions and
  - differing/inner functions.
  
  We are using the `inner` attribute of the class to contain the inner parts (`project -> compared functions -> differing functions`). For this update, we would also use the class to represent the result of the *sysctl parameter*, so the hierarchy would look like this `project -> sysctl -> compared functions -> differing functions`). The changes would probably also require adding a new attribute to the class that would distinguish between these 'kinds' of results and additional changes to the class to make it work correctly.
- Updating the [`compare` function](https://github.com/diffkemp/diffkemp/blob/8912507d38d3a9591ee55f00a6d7524b204d0255/diffkemp/compare.py#L21-L120) so it creates an instance of the `Result` class for a *sysctl parameter* and inserts the results of compared functions into it...
- Updating the [`YamlOutput` class](https://github.com/diffkemp/diffkemp/blob/8912507d38d3a9591ee55f00a6d7524b204d0255/diffkemp/output.py#L8-L145) so it outputs the updated `diffkemp-out.yaml` file format. 
- The `viewer` may also need to be also updated so it handles correctly the updated format of the `diffkemp-out.yaml` file.

davidkre525:
Fixes #403 
Function comparison now creates Result instance for every compared sysctl
parameter and changes the result structure, so it looks like:
project -> sysctl -> compared functions -> differing functions
instead of:
project -> compared functions -> differing functions
This new format is then parsed into the diffkemp-out.yaml file in output.py.
The result viewer has been adjusted to handle this new format.

One thing I noticed when comparing diffkemp-out.yaml before and after the changes in this PR was that in the new version there were few functions, which weren't present in the original. I assumed this might happen when compare function overwrites found differences with an empty diff, which is then not parsed into the output file. Is that possible, or is it caused by a mistake in the changes?

### DIFF
--- a/tests/unit_tests/result_test.py
+++ b/tests/unit_tests/result_test.py
@@ -108,10 +108,7 @@ def test_callstack_get_symbol_names(callstack):
 
 
 @pytest.fixture
-def result(graph):
-    result = Result(Result.Kind.NONE, "old-snapshot-path", "new-snapshot-path",
-                    hierarchy_level=Result.HierarchyLevel.OVERALL)
-
+def single_function_result(graph):
     comp1_result = Result(Result.Kind.NONE, "main_function", "main_function")
     objects_to_compare, *_ = \
         graph.graph_to_fun_pair_list("main_function", "main_function", False)
@@ -120,13 +117,60 @@ def result(graph):
         fun_result.first = fun_pair[0]
         fun_result.second = fun_pair[1]
         comp1_result.add_inner(fun_result)
+    yield comp1_result
 
-    result.add_inner(comp1_result)
+
+@pytest.fixture
+def multi_functions_result(graph, single_function_result):
+    result = Result(Result.Kind.NONE, "old-snapshot-path", "new-snapshot-path",
+                    hierarchy_level=Result.HierarchyLevel.OVERALL)
+    result.add_inner(single_function_result)
     result.graph = graph
     yield result
 
 
-def test_yaml_output(result, mocker):
+@pytest.fixture
+def sysctl_group_result(graph, single_function_result):
+    result = Result(Result.Kind.NONE, "old-snapshot-path", "new-snapshot-path",
+                    hierarchy_level=Result.HierarchyLevel.OVERALL)
+    group_result = Result(Result.Kind.NONE, "group_name", "group_name",
+                          hierarchy_level=Result.HierarchyLevel.GROUP)
+    group_result.add_inner(single_function_result)
+    result.add_inner(group_result)
+    result.graph = graph
+    yield result
+
+
+@pytest.fixture
+def multi_groups_result(graph, single_function_result):
+    result = Result(Result.Kind.NONE, "old-snapshot-path", "new-snapshot-path",
+                    hierarchy_level=Result.HierarchyLevel.OVERALL)
+    outer_group_result = Result(Result.Kind.NONE, "outer_name", "outer_name",
+                                hierarchy_level=Result.HierarchyLevel.GROUP)
+    inner_group_result = Result(Result.Kind.NONE, "inner_name", "inner_name",
+                                hierarchy_level=Result.HierarchyLevel.GROUP)
+    inner_group_result.add_inner(single_function_result)
+    outer_group_result.add_inner(inner_group_result)
+    result.add_inner(outer_group_result)
+    result.graph = graph
+    yield result
+
+
+@pytest.fixture
+def result(request):
+    return request.getfixturevalue(request.param)
+
+
+@pytest.mark.parametrize(
+        "result,tree_kind",
+        [
+            ("multi_functions_result",) * 2,
+            ("sysctl_group_result",) * 2,
+            ("multi_groups_result",) * 2
+        ],
+        indirect=["result"]
+)
+def test_yaml_output(result, tree_kind, mocker):
     """Tests YAML representation of compare result made by YamlOutput class."""
     mocker.patch("diffkemp.output.get_end_line", return_value=2958)
 
@@ -142,7 +186,22 @@ def test_yaml_output(result, mocker):
     assert yaml["new-snapshot"] == "/abs/path/to/new-snapshot"
     assert len(yaml["results"]) == 1
 
-    main_function_result = yaml["results"][0]
+    if tree_kind == "multi_functions_result":
+        main_function_result = yaml["results"][0]
+    elif tree_kind == "sysctl_group_result":
+        group_result = yaml["results"][0]
+        assert len(group_result["results"]) == 1
+        assert group_result["sysctl"] == "group_name"
+        main_function_result = group_result["results"][0]
+    else:
+        outer_group_result = yaml["results"][0]
+        assert len(outer_group_result["results"]) == 1
+        assert outer_group_result["sysctl"] == "outer_name"
+        inner_group_result = outer_group_result["results"][0]
+        assert len(inner_group_result["results"]) == 1
+        assert inner_group_result["sysctl"] == "inner_name"
+        main_function_result = inner_group_result["results"][0]
+
     assert main_function_result["function"] == "main_function"
     assert len(main_function_result["diffs"]) == 3
     for diff in main_function_result["diffs"]:


diffkemp/compare.py
```diff
@@ -139,10 +139,18 @@ def _compare_snapshots(self):
         return 0
 
     def _compare_groups(self, group_info):
+        group_name = group_info.group_name
+        if group_name is not None:
+            group_result = Result(Result.Kind.NONE, group_name, group_name)
+            self.result.add_inner(group_result)
+            self.result.has_inner_groups = True
+        else:
+            group_result = self.result
```
viktormalik: I'd just do
```suggestion
        if group_name is not None:
            self.result.add_inner(Result(Result.Kind.NONE, group_name, group_name))
```

view/src/Result.js
```diff
@@ -44,42 +44,54 @@ export default class Result {
...
+      });
     });
-    this.#createAllCompAndDiffFuns(yaml.results);
+    this.#createAllCompAndDiffFuns();
   }
 
   /**
    * Creates allCompFuns and allDiffFuns fields.
    * Expects that names of compared and differing functions are sorted in the `results`.
    * @param results - Differences found by DiffKemp compare phase.
```
PLukas2018: Update the comment please.

PLukas2018 (changes requested):
Thanks for the PR.

Firstly, some higher-level notes:
1. Based on the code, IIUC, you chose the second format suggested in issue #403. I was thinking about it more thoroughly, and I guess the format is OK.

   The first one would probably be nicer for existing scripts, as it would change the format minimally, and it should not break existing scripts. The one you choose does not change format in case of comparison of the projects' functions, but for sysctl parameters, the change is more radical.

   As most of the scripts (side projects) we currently have are created for comparison of functions and not sysctl parameters, I think the selected format is acceptable. It will break, e.g., our bot, but it should be easy to fix. Hopefully, there are no external scripts relying on the current format of output for sysctl params.
2. Tip: It is nice to link the issue to the PR. If it is linked, when the PR is merged, the issue will be automatically closed. The linking can be done manually or by mentioning the issue in a specific way in the PR description (see [docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)).
3. I think the commit messages could be better, specifically the second one (why the change, how it changes the output).
4. Please check and update the code for `Result.report_symbol_stat`, `Result.report_object_stat`, so it works correctly also in case of comparison of *sysctl* parameters.

---

> One thing I noticed when comparing diffkemp-out.yaml before and after the changes in this PR was that in the new version there were few functions, which weren't present in the original. I assumed this might happen when compare function overwrites found differences with an empty diff, which is then not parsed into the output file. Is that possible, or is it caused by a mistake in the changes?

If I am not mistaken, the call stacks can be a  little bit different for each run of comparison, because it depends on which branch it chooses for comparison as first or something like that... So the call stacks can change, for some runs they can be longer, for other shorter, but the compared function and the last function in the call stack should be always the same (I think). 

Personally, while testing this PR, I did not find a case where there would be more differing functions for the compared one (If I did not overlook something). E.g. this should be fine, because the same function (in this case, macro) is different, it just was found by a different path of function calls.
```yaml
# old
results:
- function: __alloc_pages_nodemask
  diffs:
  - function: NR_PAGEFLAGS
    old-callstack:
    - name: get_page_from_freelist
      file: mm/page_alloc.c
      line: 4390
    - name: rmqueue
      file: mm/page_alloc.c
      line: 3363
    - name: check_new_pages
      file: mm/page_alloc.c
      line: 3025
    - name: check_new_page
      file: mm/page_alloc.c
      line: 1910
    - name: PAGE_FLAGS_CHECK_AT_PREP (macro)
      file: mm/page_alloc.c
      line: 1869
    - name: NR_PAGEFLAGS (macro)
      file: ./include/linux/page-flags.h
      line: 759
    new-callstack:
    - name: get_page_from_freelist
      file: mm/page_alloc.c
      line: 4407
    - name: rmqueue
      file: mm/page_alloc.c
      line: 3372
    - name: check_new_pages
      file: mm/page_alloc.c
      line: 3034
    - name: check_new_page
      file: mm/page_alloc.c
      line: 1918
    - name: PAGE_FLAGS_CHECK_AT_PREP (macro)
      file: mm/page_alloc.c
      line: 1877
    - name: NR_PAGEFLAGS (macro)
      file: ./include/linux/page-flags.h
      line: 785
# new
results:
- function: __alloc_pages_nodemask
  diffs:
  - function: NR_PAGEFLAGS
    old-callstack:
    - name: get_page_from_freelist
      file: mm/page_alloc.c
      line: 4390
    - name: _deferred_grow_zone
      file: mm/page_alloc.c
      line: 3331
    - name: deferred_grow_zone
      file: mm/page_alloc.c
      line: 1724
    - name: deferred_free_pages
      file: mm/page_alloc.c
      line: 1704
    - name: deferred_free_range
      file: mm/page_alloc.c
      line: 1524
    - name: __free_pages
      file: mm/page_alloc.c
      line: 1300
    - name: __free_pages_ok
      file: mm/page_alloc.c
      line: 4457
    - name: free_pages_prepare
      file: mm/page_alloc.c
      line: 1273
    - name: PAGE_FLAGS_CHECK_AT_PREP (macro)
      file: mm/page_alloc.c
      line: 1049
    - name: NR_PAGEFLAGS (macro)
      file: ./include/linux/page-flags.h
      line: 759
    new-callstack:
    - name: get_page_from_freelist
      file: mm/page_alloc.c
      line: 4407
    - name: _deferred_grow_zone
      file: mm/page_alloc.c
      line: 3340
    - name: deferred_grow_zone
      file: mm/page_alloc.c
      line: 1732
    - name: deferred_free_pages
      file: mm/page_alloc.c
      line: 1712
    - name: deferred_free_range
      file: mm/page_alloc.c
      line: 1532
    - name: __free_pages
      file: mm/page_alloc.c
      line: 1308
    - name: __free_pages_ok
      file: mm/page_alloc.c
      line: 4474
    - name: free_pages_prepare
      file: mm/page_alloc.c
      line: 1281
    - name: PAGE_FLAGS_CHECK_AT_PREP (macro)
      file: mm/page_alloc.c
      line: 1050
    - name: NR_PAGEFLAGS (macro)
      file: ./include/linux/page-flags.h
      line: 785
```

 So, if you meant something like that, you can probably ignore it. In any case, you can post a snippet of the output that you think is weird and we can look at it.

---



More detailed notes are below:

diffkemp/compare.py
```diff
@@ -139,10 +139,18 @@ def _compare_snapshots(self):
         return 0
 
     def _compare_groups(self, group_info):
+        group_name = group_info.group_name
+        if group_name is not None:
+            group_result = Result(Result.Kind.NONE, group_name, group_name)
+            self.result.add_inner(group_result)
```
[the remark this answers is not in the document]
PLukas2018: Btw. I think that calling `self.result.add_inner`  below `self._compare_function` should also solve #406.

diffkemp/semdiff/result.py
```diff
@@ -194,6 +194,7 @@ def __init__(self, kind, first_name, second_name, start_time=None,
         self.inner = dict()
         self.start_time = start_time
         self.stop_time = stop_time
+        self.has_inner_groups = False
```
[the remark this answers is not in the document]
PLukas2018: I am not sure, as I do not have enough experience with project growth, and I am currently not really sure what other 'levels' there could be in the future.

I think not naming the level `sysctl` but rather more generally  `group` could help. Also, maybe for the current PR, you are right that being able to distinguish between `top-level`, `compared function`, `differing function` could be enough, but I am not sure what would be a good solution to do this...

diffkemp/semdiff/result.py
```diff
@@ -220,18 +221,14 @@ def report_symbol_stat(self, show_errors=False):
...
-                    if r.kind in [Result.Kind.ERROR, Result.Kind.TIMEOUT]])
-        empty_diff = len([r for r in iter(self.inner.values()) if all(map(
-            lambda x: x.diff == "", r.inner.values())) and
-            r.kind == Result.Kind.NOT_EQUAL])
+        stats = 0, 0, 0, 0, 0, 0
+        if self.has_inner_groups:
+            for group_result in self.inner.values():
+                stats = group_result.group_symbol_stat(stats)
+        else:
+            stats = self.group_symbol_stat(stats)
```
PLukas2018: I am not sure, but maybe a recursion would make sense, but some variable would be necessary (`hierarchy_level` or `result_kind` to find that the result describes a result for *differing* / *compared* function) or based on if the inner is empty, then stop the recursion and return appropriate values.

PLukas2018 (commented):
Thanks for the changes. I am providing a few comments, but I am not exactly sure what would be the best approach to implement the `Result` class hierarchy and the class methods...

davidkre525:
Thanks for the notes, to address them I made the following changes:

- in` Result` class, `has_inner_groups` field was removed,` hierarchy_level` enum field was added
- every function working with the result created in compare function treats it as a tree, that means doesnt explect fixed height and doesnt expect the height to be the same in every branch of the tree

I ended up making the hierarchy_level enum with three values: `overall_result`, `group` and` function`. From my prespective, adding also` differing_function` doesnt make that much sense, because when function working with the tree sees `hierarchy_level == function`, the child of this result has to be `differing_function`. And adding this value would mean adding parameter every time result class is initialized. What do you think about that?
I still need to adjust the result viewer to work with the tree-like format and add some tests into tests/unit_tests/result_test.py, as you mentioned above.

diffkemp/semdiff/result.py
```diff
@@ -183,8 +183,23 @@ def __init__(self, name, filename=None, line=None, callstack=None,
             self.diff_kind = diff_kind
             self.covered = covered
 
+    class Hierarchy_level(IntEnum):
+        """Describes position of result instance in result comparison tree"""
+        OVERALL = 0
+        GROUP = 1
+        FUNCTION = 2
+
+    class Symbol_stat:
```
PLukas2018: `SymbolStat` would be better.

tests/unit_tests/result_test.py
```diff
@@ -119,13 +117,60 @@ def result(graph):
...
+    result = Result(Result.Kind.NONE, "old-snapshot-path", "new-snapshot-path",
+                    hierarchy_level=Result.HierarchyLevel.OVERALL)
+    result.add_inner(main_func_result)
+    result.graph = graph
+    yield result
 
-    result.add_inner(comp1_result)
+
+@pytest.fixture
+def result2(graph, main_func_result):
```
PLukas2018: I would rename this to smthing like `sysctl_result`.

tests/unit_tests/result_test.py
```diff
@@ -119,13 +117,60 @@ def result(graph):
...
+                          hierarchy_level=Result.HierarchyLevel.GROUP)
+    group_result.add_inner(main_func_result)
+    result.add_inner(group_result)
     result.graph = graph
     yield result
 
 
-def test_yaml_output(result, mocker):
+@pytest.fixture
+def result3(graph, main_func_result):
```
PLukas2018: I am not sure if adding this testing result makes sense because it does not currently represent real usage. I will leave it up to you if you leave or remove it.

diffkemp/output.py
```diff
@@ -31,48 +31,70 @@ def _create_output(self):
...
+
+        inner_results = []
+        for child_name, child_result in current_result.inner.items():
+            self._create_results(child_result, inner_results, child_name)
+
+        if current_result.hierarchy_level == Result.HierarchyLevel.OVERALL:
+            self.output["results"] = inner_results
+        else:
+            outer_results.append({
+                "sysctl": name,  # sysctl is the only usecase of groups
```
PLukas2018: The `sysctl` 'group name' should not be probably fixed but extracted somehow from the `hierarchy_level` / `result`.

diffkemp/compare.py
```diff
@@ -139,9 +141,24 @@ def _compare_snapshots(self):
...
     def _compare_groups(self, group_info):
+        group_name = group_info.group_name
+        # Checks if there are groups (e.g. for sysctl parameters),
+        # creates group_result to be able to connect results in the group
+        if group_name is not None:
+            group_info.group_result = Result(
+                Result.Kind.NONE,
+                group_name,
+                group_name,
+                hierarchy_level=Result.HierarchyLevel.GROUP)
```
PLukas2018: So, we should probably also provide somehow the group name -- in this case `sysctl`.

tests/unit_tests/result_test.py
```diff
@@ -108,9 +108,7 @@ def test_callstack_get_symbol_names(callstack):
 
 
 @pytest.fixture
-def result(graph):
-    result = Result(Result.Kind.NONE, "old-snapshot-path", "new-snapshot-path")
-
+def main_func_result(graph):
```
viktormalik: How about
```suggestion
def single_function_result(graph):
```

tests/unit_tests/result_test.py
viktormalik: I think it's ok, at least we can check that it works with multiple hierarchy levels.

view/src/Result.js
```diff
@@ -36,6 +36,16 @@ export default class Result {
...
 
+  /**
+   * Comparison results with the following structure:
+   * -results:
+   *   -sysctl: name (this field exists only in case of sysctl comparison)
+   *    results:
+   *     -function: name
+   *       -diffs
+  */
+  groupResults;
```
viktormalik: I think this could be just named `results`. Any reason why we need to stress the "group" part here?

viktormalik:
> 2.   I tested this PR, I discovered that it provides different stats for *sysctl* parameters than previously. This does not have to be a necessary thing that should be fixed, but we should discuss if we are ok that it will, from now on, provide a little bit different stats than previously.

This is an excellent finding, thank you for the thorough investigation @PLukas2018!

> E.g. -- the field `Not equal:`:
> - In case of *sysctl* parameters `kernel.hung_task_panic` and `kernel.hung_task_warnings` is compared with the function `check_hung_task` and for both of them the function is evaluated as non-equal - previously it was counted only once, now it is counted twice. I think it should be counted twice as it is now, because IIUC it was compared twice. The cached result could not be reused because of slicing (in `kernel.hung_task_warnings` using `sysctl_hung_task_warnings` var and in `kernel.hung_task_panic` using `sysctl_hung_task_panic` var).

I think that this captures the essence of the problem. We seem to have a logical flaw somewhere - we should always use a combination of function name and global variable when slicing was applied (because technically, different functions were compared). We seem to correctly use it in some places (since caching was not applied) but not in the others. 

> I think the current stats should be more precise (IIUC, previously, the precision of the stats was lost by accumulating all results in one overall result without subgroups for `sysctl` params). I think the stats changes introduced by this PR should be hopefully ok, @viktormalik WDYT?

Agreed, the current results are more precise. This PR basically mitigates the problem described above but there may be more places affected by this. But for that, a better investigation is necessary. For now, we can proceed with this PR. 

> Btw. the differences in stats are also visible in [this comment](https://github.com/PLukas2018/diffkemp/pull/15#issuecomment-3867782276) from our bot.

Nice!

viktormalik (approved):
Looks great to me!

Just one small nit in the tests, otherwise I think this is ready to go.

viktormalik:
#431 is merged, this needs a rebase on master

viktormalik:
Nice work!

Issue #415, opened 2025-09-25
Cpp code style checking - update `clang-format` config


PLukas2018:
When using `tools/check-clang-format.sh -d` in `nix develop .#diffkemp-llvm18` shell, the linter reports  incorrectly formatted C++ files, these are not reported when using LLVM v17 or lower. This is probably caused by changes in `clang-format` in v18. We should probably create/update the linter configuration so it behaves the same on all versions. Btw. the reason why this is not reported by CI is that we forgot to update the LLVM version (respectively the nix development shell 'version') on which the code style checking is run in #397.

---

The reported incorrect formatting. 

```txt
Checking the format of file diffkemp/simpll/ResultsCache.h
38c38,39
<     ResultsCache(std::string cacheDirectory) : cacheDirectory(cacheDirectory){};
---
>     ResultsCache(std::string cacheDirectory)
>             : cacheDirectory(cacheDirectory) {};

Checking the format of file diffkemp/simpll/ValuePatternComparator.h
42c42
<               ParentPattern(ParentPattern){};
---
>               ParentPattern(ParentPattern) {};

Checking the format of file diffkemp/simpll/library/SysctlTable.h
43c43
<             : Mod(Mod), CtlTable(CtlTable){};
---
>             : Mod(Mod), CtlTable(CtlTable) {};

Checking the format of file diffkemp/simpll/Result.h
103c103
<     NonFunctionDifference(DiffKind Kind) : Kind(Kind){};
---
>     NonFunctionDifference(DiffKind Kind) : Kind(Kind) {};
131c131
<             : NonFunctionDifference(SynDiff), syntaxKind(Kind::UNKNOWN){};
---
>             : NonFunctionDifference(SynDiff), syntaxKind(Kind::UNKNOWN) {};

Checking the format of file diffkemp/simpll/Logger.h
234c234
<     Logger(){};
---
>     Logger() {};

Checking the format of file diffkemp/simpll/InstPatternComparator.h
52c52
<               ParentPattern(ParentPattern){};
---
>               ParentPattern(ParentPattern) {};

Checking the format of file diffkemp/simpll/CustomPatternSet.cpp
145,146c145,146
<         LOG("Failed to open difference "
<             << "pattern configuration " << ConfigPath << ".\n");
---
>         LOG("Failed to open difference " << "pattern configuration "
>                                          << ConfigPath << ".\n");
156,157c156,157
<         LOG("Failed to parse difference "
[...]

PR #423, opened 2025-11-28
Fix formating issue between versions

Bal0sk:
Fixes: #415

Clang-format 18 always adds spaces before constructor bodies. v17 omitted this space when a constructor had an initializer list. This is easily fixed by enforcing the space in v17 in the config file.

**Second issues** is more troublesome. After modifying the config file, with the fix above, only one file was flagged when run with v18. `CustomPatternSet.cpp` is the only conflicting file, due to the `<<` operators in 4 `LOG` macros.
- v17 enforces a line break before the first `<<` operator, since it cant fit both `<<` operators into one line, it prefers to have each string in the stream, in a new line.
- v18 keeps the line and breaks before the second `<<` operator, opting for a more concise packing.

Changing the penalties/rules regarding this conflict resulted in more files breaking, while the issue persisted. I tried creating separate config files as well with the same result. The simplest solutions I found:
- combine the strings in the file. This would prevent v17 from splitting it to multiple lines
- making the linter skip these snippets of code by commenting it out using `//clang-format off/on`, as is done in this commit.

With the nature of this issue being code style, I think one of these fixes is sufficient, even though it doesn't prevent future issues. But, since the fix would require changing the code itself, rather than the config file, or other method, I would like more opinions on this matter.

_Couple notes:_
The other solution I thought of, would be to force one specific version of `config-format` using nix flakes, however, in my opinion, this would create an unnecessary overhead and would only fix the problem if it was run in nix environment.

In clang16 a new attribute to the config was added `InsertNewlineAtEOF`, which could be useful, but since we support versions under 16, we can't use the attribute yet. We would need a different config file for v16+.


Bal0sk:
AFAIK, the `--Wno-error=unknown` should only change the error of unknown command found and should not interfere with the formatting errors, since I don't think any "unknown" formatting errors should appear. I also messed with a file to check if it worked with v15 and it reported the problem correctly, without issues.
Also, while searching, if that could happen, I found an issue in a VS-code extension [repo](https://github.com/microsoft/vscode-cpptools/issues/10072) which used this method to do, what we want as well.
From the official clang website, the explanation is just:

> Wno-error=foo
Turn warning “foo” into a warning even if [-Werror](https://clang.llvm.org/docs/UsersManual.html#cmdoption-Werror) is specified.

I will wait for others' opinion, before making changes.

PLukas2018 (commented):
Thanks for the changes. I have just two notes; otherwise, the changes LGTM.

PR #432, opened 2026-03-17
Add comparison of global variables

### LINKED ISSUE #79: Implement KABI global variable comparison.
KABI lists contain two types of symbols - function and global variables. Currently values that aren't function are skipped.
To implement global variable comparison for each global variable a list containing all functions that use it must be built using functions from llvmcpy and stored in functions.yaml as a group containing the global variable name (similarly to how sysctl comparison is handled).
Then the `compare` function must be modified to be able to handle the new kind of group.

davidkre525:
Fixes #79 
Adds comparison of global variables if they are present in the symbol list during function comparison.
Adds group_kind to Result to distinguish between different kinds of groups (now when there is more then one).
I still need to adjust the viewer for the new format.
And I wanetd to ask, is calling LlvmParam(glob_var) the correct way of creating llvm parameter from global variable name, or is there some method for that?

### DIFF
--- a/diffkemp/building/build_kernel.py
+++ b/diffkemp/building/build_kernel.py
@@ -4,6 +4,7 @@ from diffkemp.llvm_ir.kernel_source_tree import KernelSourceTree
 from diffkemp.snapshot import Snapshot
 from diffkemp.building.build_utils import (
     generate_from_function_list,
+    add_funcs_for_data,
     read_symbol_list,
     EMSG_EMPTY_SYMBOL_LIST)
 import errno
@@ -97,7 +98,8 @@ def generate_from_sysctl_list(snapshot, sysctl_list):
             # Proc handler function for sysctl
             _add_proc_handler(snapshot, sysctl, proc_fun)
             # Adds functions which use the sysctl data variable
-            _add_funcs_for_data(snapshot, sysctl, sysctl_mod, proc_fun)
+            data = sysctl_mod.get_data(sysctl)
+            add_funcs_for_data(snapshot, sysctl, data, proc_fun)
 
 
 def _add_proc_handler(snapshot, sysctl, proc_fun):
@@ -118,34 +120,3 @@ def _add_proc_handler(snapshot, sysctl, proc_fun):
                             snapshot.source_tree.source_dir)))
     except SourceNotFoundException:
         print("  could not build proc handler")
-
-
-def _add_funcs_for_data(snapshot, sysctl, sysctl_mod, proc_fun):
-    data = sysctl_mod.get_data(sysctl)
-
-    if not data:
-        return
-
-    for module in \
-            snapshot.source_tree.get_modules_using_symbol(data.name):
-        # For now, we only support the x86 architecture in kernel
-        if "/arch/" in module.llvm and \
-                "/arch/x86/" not in module.llvm:
-            continue
-
-        curr_mod_path = os.path.relpath(module.llvm,
-                                        snapshot.source_tree.source_dir)
-        for func in module.get_functions_using_param(data):
-            if func == proc_fun:
-                continue
-
-            snapshot.add_fun(
-                name=func,
-                llvm_mod=module,
-                glob_var=data.name,
-                tag="using data variable \"{}\"".format(data.name),
-                group=sysctl)
-            print("  {}: {} (using data variable \"{}\")".format(
-                func,
-                curr_mod_path,
-                data.name))
--- a/diffkemp/building/build_kernel.py
+++ b/diffkemp/building/build_kernel.py
@@ -4,6 +4,7 @@ from diffkemp.llvm_ir.kernel_source_tree import KernelSourceTree
 from diffkemp.snapshot import Snapshot
 from diffkemp.building.build_utils import (
     generate_from_function_list,
+    add_funcs_for_globvar,
     read_symbol_list,
     EMSG_EMPTY_SYMBOL_LIST)
 import errno
@@ -97,7 +98,8 @@ def generate_from_sysctl_list(snapshot, sysctl_list):
             # Proc handler function for sysctl
             _add_proc_handler(snapshot, sysctl, proc_fun)
             # Adds functions which use the sysctl data variable
-            _add_funcs_for_data(snapshot, sysctl, sysctl_mod, proc_fun)
+            data = sysctl_mod.get_data(sysctl)
+            add_funcs_for_globvar(snapshot, sysctl, data, proc_fun)
 
 
 def _add_proc_handler(snapshot, sysctl, proc_fun):
@@ -118,34 +120,3 @@ def _add_proc_handler(snapshot, sysctl, proc_fun):
                             snapshot.source_tree.source_dir)))
     except SourceNotFoundException:
         print("  could not build proc handler")
-
-
-def _add_funcs_for_data(snapshot, sysctl, sysctl_mod, proc_fun):
-    data = sysctl_mod.get_data(sysctl)
-
-    if not data:
-        return
-
-    for module in \
-            snapshot.source_tree.get_modules_using_symbol(data.name):
-        # For now, we only support the x86 architecture in kernel
-        if "/arch/" in module.llvm and \
-                "/arch/x86/" not in module.llvm:
-            continue
-
-        curr_mod_path = os.path.relpath(module.llvm,
-                                        snapshot.source_tree.source_dir)
-        for func in module.get_functions_using_param(data):
-            if func == proc_fun:
-                continue
-
-            snapshot.add_fun(
-                name=func,
-                llvm_mod=module,
-                glob_var=data.name,
-                tag="using data variable \"{}\"".format(data.name),
-                group=sysctl)
-            print("  {}: {} (using data variable \"{}\")".format(
-                func,
-                curr_mod_path,
-                data.name))
--- a/diffkemp/building/build_utils.py
+++ b/diffkemp/building/build_utils.py
@@ -1,4 +1,5 @@
 from diffkemp.llvm_ir.source_tree import SourceNotFoundException
+from diffkemp.llvm_ir.llvm_module import LlvmParam
 import sys
 import os
 
@@ -41,9 +42,44 @@ def generate_from_function_list(snapshot, fun_list):
                 snapshot.add_fun(symbol, llvm_mod)
                 print(os.path.relpath(llvm_mod.llvm,
                                       snapshot.source_tree.source_dir))
+            elif llvm_mod.has_global(symbol):
+                if add_funcs_for_data(snapshot, symbol, LlvmParam(symbol)):
+                    snapshot.list_kind = "glob_var"
             else:
                 snapshot.add_fun(symbol, None)
-                print("not a function")
+                print("not a function nor a global variable")
         except SourceNotFoundException:
             print("source not found")
             snapshot.add_fun(symbol, None)
+
+
+def add_funcs_for_data(snapshot, name, data, proc_fun=None):
+    functions_added = False
+    if not data:
+        return functions_added
+
+    for module in \
+            snapshot.source_tree.get_modules_using_symbol(data.name):
+        # For now, we only support the x86 architecture in kernel
+        if "/arch/" in module.llvm and \
+                "/arch/x86/" not in module.llvm:
+            continue
+
+        curr_mod_path = os.path.relpath(module.llvm,
+                                        snapshot.source_tree.source_dir)
+        for func in module.get_functions_using_param(data):
+            if func == proc_fun:
+                continue
+
+            functions_added = True
+            snapshot.add_fun(
+                name=func,
+                llvm_mod=module,
+                glob_var=data.name,
+                tag="using data variable \"{}\"".format(data.name),
+                group=name)
+            print("  {}: {} (using data variable \"{}\")".format(
+                func,
+                curr_mod_path,
+                data.name))
+    return functions_added
--- a/diffkemp/building/build_utils.py
+++ b/diffkemp/building/build_utils.py
@@ -1,4 +1,5 @@
 from diffkemp.llvm_ir.source_tree import SourceNotFoundException
+from diffkemp.llvm_ir.llvm_module import LlvmParam
 import sys
 import os
 
@@ -41,9 +42,45 @@ def generate_from_function_list(snapshot, fun_list):
                 snapshot.add_fun(symbol, llvm_mod)
                 print(os.path.relpath(llvm_mod.llvm,
                                       snapshot.source_tree.source_dir))
+            elif llvm_mod.has_global(symbol):
+                print()
+                if add_funcs_for_globvar(snapshot, symbol, LlvmParam(symbol)):
+                    snapshot.list_kind = "glob_var"
             else:
                 snapshot.add_fun(symbol, None)
-                print("not a function")
+                print("not a function nor a global variable")
         except SourceNotFoundException:
             print("source not found")
             snapshot.add_fun(symbol, None)
+
+
+def add_funcs_for_globvar(snapshot, name, data, skip_fun=None):
+    functions_added = False
+    if not data:
+        return functions_added
+
+    for module in \
+            snapshot.source_tree.get_modules_using_symbol(data.name):
+        # For now, we only support the x86 architecture in kernel
+        if "/arch/" in module.llvm and \
+                "/arch/x86/" not in module.llvm:
+            continue
+
+        curr_mod_path = os.path.relpath(module.llvm,
+                                        snapshot.source_tree.source_dir)
+        for func in module.get_functions_using_param(data):
+            if skip_fun is not None and func == skip_fun:
+                continue
+
+            functions_added = True
+            snapshot.add_fun(
+                name=func,
+                llvm_mod=module,
+                glob_var=data.name,
+                tag="using global variable \"{}\"".format(data.name),
+                group=name)
+            print("  {}: {} (using global variable \"{}\")".format(
+                func,
+                curr_mod_path,
+                data.name))
+    return functions_added
--- a/diffkemp/compare.py
+++ b/diffkemp/compare.py
@@ -33,7 +33,7 @@ def compare(args):
 
 class GroupInfo:
     def __init__(self, group, group_name, enable_module_cache, config,
-                 output_dir):
+                 output_dir, group_kind):
         self.group = group
         self.group_name = group_name
         self.group_dir = self._get_group_dir(output_dir)
@@ -44,6 +44,7 @@ class GroupInfo:
         self.cache = SimpLLCache(mkdtemp())
         self.module_cache = {}
         self.group_result = None
+        self.group_kind = group_kind
 
     def _get_group_dir(self, output_dir):
         if output_dir is not None and self.group_name is not None:
@@ -92,7 +93,8 @@ class SnapshotComparator:
         self.result = Result(Result.Kind.NONE, cmd_args.snapshot_dir_old,
                              cmd_args.snapshot_dir_new,
                              start_time=default_timer(),
-                             hierarchy_level=Result.HierarchyLevel.OVERALL)
+                             hierarchy=Result.Hierarchy(
+                                        Result.Hierarchy.Level.OVERALL))
         self.regex_pattern = re.compile(cmd_args.regex_filter) \
             if cmd_args.regex_filter else None
         self.output_dir = None
@@ -131,10 +133,12 @@ class SnapshotComparator:
 
     def _compare_snapshots(self):
         for group_name, group in sorted(self.config.snapshot_first
-                                        .fun_groups.items()):
+                                        .fun_groups.items(),
+                                        key=lambda x: (x[0] is not None, x)):
             group_info = GroupInfo(group, group_name,
                                    self.args.enable_module_cache,
-                                   self.config, self.output_dir)
+                                   self.config, self.output_dir,
+                                   self.config.snapshot_first.list_kind)
             self._compare_groups(group_info)
 
         self._finalize_output()
@@ -149,7 +153,9 @@ class SnapshotComparator:
                 Result.Kind.NONE,
                 group_name,
                 group_name,
-                hierarchy_level=Result.HierarchyLevel.GROUP)
+                hierarchy=Result.Hierarchy(
+                            Result.Hierarchy.Level.GROUP,
+                            group_info.group_kind))
         else:
             group_info.group_result = self.result
 
--- a/diffkemp/compare.py
+++ b/diffkemp/compare.py
@@ -33,7 +33,7 @@ def compare(args):
 
 class GroupInfo:
     def __init__(self, group, group_name, enable_module_cache, config,
-                 output_dir):
+                 output_dir, group_kind):
         self.group = group
         self.group_name = group_name
         self.group_dir = self._get_group_dir(output_dir)
@@ -44,6 +44,7 @@ class GroupInfo:
         self.cache = SimpLLCache(mkdtemp())
         self.module_cache = {}
         self.group_result = None
+        self.group_kind = group_kind
 
     def _get_group_dir(self, output_dir):
         if output_dir is not None and self.group_name is not None:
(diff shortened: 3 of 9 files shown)


diffkemp/building/build_utils.py
```diff
@@ -41,9 +42,44 @@ def generate_from_function_list(snapshot, fun_list):
...
             else:
                 snapshot.add_fun(symbol, None)
-                print("not a function")
+                print("not a function nor a global variable")
         except SourceNotFoundException:
             print("source not found")
             snapshot.add_fun(symbol, None)
+
+
+def add_funcs_for_data(snapshot, name, data, proc_fun=None):
```
viktormalik: Since this function is no longer specific to sysctls, let's rename the last parameter:
```suggestion
def add_funcs_for_data(snapshot, name, data, skip_fun=None):
```

diffkemp/building/build_utils.py
```diff
@@ -41,9 +42,44 @@ def generate_from_function_list(snapshot, fun_list):
...
+            snapshot.source_tree.get_modules_using_symbol(data.name):
+        # For now, we only support the x86 architecture in kernel
+        if "/arch/" in module.llvm and \
+                "/arch/x86/" not in module.llvm:
+            continue
+
+        curr_mod_path = os.path.relpath(module.llvm,
+                                        snapshot.source_tree.source_dir)
+        for func in module.get_functions_using_param(data):
+            if func == proc_fun:
```
viktormalik: Just to make sure we don't match on `None`:
```suggestion
            if proc_fun is not None and func == proc_fun:
```

diffkemp/building/build_utils.py
viktormalik: Also, we tend to use "global variable" instead of "data" so let's rename it, too:

```suggestion
def add_funcs_for_globvar(snapshot, name, data, proc_fun=None):
```

diffkemp/building/build_utils.py
```diff
@@ -41,9 +42,44 @@ def generate_from_function_list(snapshot, fun_list):
...
+                continue
+
+            functions_added = True
+            snapshot.add_fun(
+                name=func,
+                llvm_mod=module,
+                glob_var=data.name,
+                tag="using data variable \"{}\"".format(data.name),
```
viktormalik: Also here, let's replace "data variable" by "global variable".

```suggestion
                tag="using global variable \"{}\"".format(data.name),
```
Same two lines below.

diffkemp/compare.py
```diff
@@ -131,10 +133,12 @@ def _default_output_dir(self):
 
     def _compare_snapshots(self):
         for group_name, group in sorted(self.config.snapshot_first
-                                        .fun_groups.items()):
+                                        .fun_groups.items(),
+                                        key=lambda x: (x[0] is not None, x)):
```
viktormalik: What is the purpose of this?

viktormalik (commented):
Just a couple of suggestions, otherwise looks good.

diffkemp/compare.py
davidkre525: When functions are compared, they are all stored in one group with group_name = None [(add_fun)](https://github.com/davidkre525/diffkemp/blob/80112c2734783d08c195ed3489c3f404655311bf/diffkemp/snapshot.py#L134). If global variable is present in the symbol list, new group is created with group_name being the name of the variable. And python cannot compare None and str.
I have renamed the lambda parameter and added a comment to make it more readable. But maybe it would be better to use empty string instead of None?

davidkre525:
Thanks for the notes, everything mentioned should be fixed now.

PLukas2018:
Hi. Sorry for the late review. The code itself seems ok-ish to me. I have a few questions to understand this better and also to 'be used as documentation':

IIUC in case the list provided to `build-kernel` will contain not only functions but also global variables, then the `snapshot.yaml` has `glob_var` set as `list_kind` instead of `function`?
https://github.com/davidkre525/diffkemp/blob/9761afe466a4b08c3ea468f26e9b1c72bb0f2b70/diffkemp/building/build_utils.py#L47

IIUC The format of `diffkemp-out.yaml` looks something like this now?

```yaml
results:
  - function: fun3
    diffs:
      - function: fun2
        old-callstack: []
        new-callstack: []
      - function: fun3
        old-callstack: []
        new-callstack: []
  - glob_var: glob1
    results:
      - function: fun1
        diffs:
          - function: fun2
            old-callstack: []
            new-callstack: []
          - function: fun1
            old-callstack: []
            new-callstack: []
```

I am thinking if we should not add argument to `build-kernel` command to allow building snapshots for just functions (so it behaves the same as previously), but probably it is not necessary as the kabi (or other) list can be pre-filtered manually.

Note: I have not tested the changes yet, [the bot](https://github.com/PLukas2018/diffkemp/pull/17#issuecomment-4728278991) did not handle the evaluation (I think maybe because of the change of the output format, but I also forgot to specify `--rebuild` arg). I will test this manually and fix the bot.

PLukas2018:
Also maybe one more question, if you tried to build and compare some kernel versions and how does comparing also the global variables increase the time necessary for building/comparing the kernels?

viktormalik (approved):
LGTM, thanks!

davidkre525:
> IIUC in case the list provided to `build-kernel` will contain not only functions but also global variables, then the `snapshot.yaml` has `glob_var` set as `list_kind` instead of `function`? https://github.com/davidkre525/diffkemp/blob/9761afe466a4b08c3ea468f26e9b1c72bb0f2b70/diffkemp/building/build_utils.py#L47


Yes. And not only the `list_kind` field, but the whole format of `snapshot.yaml` is different when the symbol list contains a global variable.
If only functions are present in the symbol list, the `list` field contains directly the functions:
```yaml
- created_time: ...
  diffkemp_version: ...
  list:
  - glob_var: null
    llvm: lib/decompress_unxz.ll
    name: kmalloc
    tag: null
    ...
```
But when there is also a global variable, the `list` field contains groups of functions. One group is for all the functions present in the symbol list and each of the remaining groups corresponds to one global variable from the symbol list.
```yaml
- created_time: ...
  diffkemp_version: ...
  list:
  - glob_var: null
    functions:
    - glob_var: null
      llvm: kernel/sched/fair.ll
      name: rebalance_domains
      tag: null
    ...
  - glob_var: sysctl_sched_rr_timeslice
    functions:
    - glob_var: sysctl_sched_rr_timeslice
      llvm: kernel/sched/rt.ll
      name: sched_rr_handler
      tag: using global variable "sysctl_sched_rr_timeslice"
   ...
```
> 
> IIUC The format of `diffkemp-out.yaml` looks something like this now?
> 
> ```yaml
> results:
>   - function: fun3
>     diffs:
>       - function: fun2
>         old-callstack: []
>         new-callstack: []
>       - function: fun3
>         old-callstack: []
>         new-callstack: []
>   - glob_var: glob1
>     results:
>       - function: fun1
>         diffs:
>           - function: fun2
>             old-callstack: []
>             new-callstack: []
>           - function: fun1
>             old-callstack: []
>             new-callstack: []
> ```

Yes, this is how it looks now.

PLukas2018:
Ok, so I have tried to test this PR and have some results:

1. The [bot is still getting stuck](https://github.com/PLukas2018/diffkemp/pull/17#issuecomment-4728278991), I am not sure why right now but I will try to fix it (it can be e.g. because of a timeout or some parsing problem, ...) .
2. When trying to compare the whole kabi list locally (I have tried 8.0, 8.1), the `build-kernel` phase seems to get stuck on `jiffies` symbol. I have not looked at it yet why, but my assumption is that it could be e.g. because of some missing config var to be configured (we have met this [in the past](https://github.com/diffkemp/diffkemp/issues/313), but it could be something different).

   Reproducer:
   ```sh
   echo jiffies > list
   diffkemp build-kernel linux-4.18.0-80.el8  4.18.0-80-jiffies list
   ```
3. I have tried to compare subset of kabi list for 8.0/8.1 and compare it with results for the subset on master:

    The used kabi subset:
    ```
    __alloc_pages_nodemask
	__alloc_skb
	__alloc_workqueue_key
	__bitmap_and
	__bitmap_andnot
	__bitmap_equal
	__bitmap_intersects
	__bitmap_parse
	__bitmap_weight
	__check_object_size
	__class_register
	__const_udelay
	__cpu_online_mask
	__cpu_possible_mask
	__cpu_present_mask
	__cpuhp_remove_state
	__cpuhp_setup_state
	__fentry__
	__free_pages
	__get_free_pages
	__get_user_2
	__init_rwsem
	__init_waitqueue_head
	__kmalloc
	__list_add_valid
	__list_del_entry_valid
	__local_bh_enable_ip
	__lock_page
	__mmu_notifier_register
	__module_get
	__msecs_to_jiffies
	__mutex_init
	__napi_schedule
	__ndelay
	__node_distance
	__page_file_index
	__page_mapcount
	__pagevec_release
	__per_cpu_offset
	__printk_ratelimit
	__pskb_pull_tail
	__put_cred
	__put_page
	__put_task_struct
	__put_user_2
	__put_user_4
	__put_user_8
	__radix_tree_insert
	__radix_tree_next_slot
	__register_nmi_handler
	__release_region
	__request_region
	__skb_gso_segment
	__stack_chk_fail
	__task_pid_nr_ns
	__tasklet_schedule
	__usecs_to_jiffies
	__uv_cpu_info
	__uv_hub_info_list
	__vmalloc
	__wake_up
	__warn_printk
	__x86_indirect_thunk_r10
	__x86_indirect_thunk_r11
	__x86_indirect_thunk_r12
	__x86_indirect_thunk_r13
	__x86_indirect_thunk_r14
	__x86_indirect_thunk_r15
	__x86_indirect_thunk_r8
	__x86_indirect_thunk_r9
	__x86_indirect_thunk_rax
	__x86_indirect_thunk_rbp
	__x86_indirect_thunk_rbx
	__x86_indirect_thunk_rcx
	__x86_indirect_thunk_rdi
	__x86_indirect_thunk_rdx
	__x86_indirect_thunk_rsi
	_atomic_dec_and_lock
	_cond_resched
	_copy_from_user
	_copy_to_user
	_ctype
	_raw_read_lock
	_raw_read_lock_irq
	_raw_read_lock_irqsave
	_raw_read_unlock_irqrestore
	_raw_spin_lock
	_raw_spin_lock_bh
	_raw_spin_lock_irq
	_raw_spin_lock_irqsave
	_raw_spin_trylock
	_raw_spin_unlock_bh
	_raw_spin_unlock_irqrestore
	_raw_write_lock
	_raw_write_lock_bh
	_raw_write_lock_irq
	_raw_write_lock_irqsave
	_raw_write_unlock_bh
	_raw_write_unlock_irqrestore
	add_timer
	add_wait_queue
	add_wait_queue_exclusive
	alloc_cpumask_var
	alloc_netdev_mqs
	alloc_pages_current
	apic
   ```
   It managed to compare 5 more symbols (`__cpu_online_mask`, `__cpu_possible_mask`, `__cpu_present_mask`, `__per_cpu_offset`, `__uv_cpu_info`, `__uv_hub_info_list`, `_ctype`, `apic`).

   Stats:
   ```
   Statistics master
    -----------
    Total symbols: 106
    Equal:         66 (62%)
    Not equal:     12 (11%)
    (empty diff):  0 (0%)
    Unknown:       28 (26%)
    Errors:        0 (0%)
    
    Elapsed time:            97.28 s
    Functions compared:      1410
    Lines compared:          12421
    Instructions compared:   38154
    1:1 equal instructions:  37704 (99%)
    
    Total differences:       35
    In functions:            25 (71%)
    In types:                2 (6%)
    In macros:               8 (23%)
    In inline assembly code: 0 (0%)
    Empty diffs:             1 (3%)   
   ```
   ```
   Statistics PR
    -----------
    Total symbols: 196
    Equal:         160 (82%)
    Not equal:     16 (8%)
    (empty diff):  1 (1%)
    Unknown:       20 (10%)
    Errors:        0 (0%)
    
    Elapsed time:            160.19 s
    Functions compared:      280
    Lines compared:          2002
    Instructions compared:   6534
    1:1 equal instructions:  6520 (100%)
    
    Total differences:       35
    In functions:            27 (77%)
    In types:                1 (3%)
    In macros:               7 (20%)
    In inline assembly code: 0 (0%)
    Empty diffs:             2 (6%)
   ```

   There seems to be a **bug in generating `diffkemp-out.yaml`** -- it adds results for the differing global variables (`__uv_hub_info_list`, `apic`), but it only contains one compared function which was found as differing (`__alloc_pages_nodemask`), but it does not contain other compared functions determined as differing (`__alloc_workqueue_key`). 

   Also when comparing individual results, there seems to be different found difffering functions in some functions, e.g. `__alloc_workqueue_key` previously contained `percpu_from_op`, now `alloc_worker`. I am not sure if there is some problem with polluting result cache or if comparing also the global variables somehow changed (enhanced?) the behaviour...
   
   When checking result viewer it seems to be ok.

@davidkre525 Overall, my personal opinion is that ideally, the bug with missing compared functions in `diffkemp-out.yaml` should be fixed as part of this PR. Also, ideally, we should check, using some kabi list, when keeping only function symbols leaves the behaviour/results of diffkemp unchanged. The other issues (not able to create a snapshot for the whole kabi list -- including global variables, ...),  I think, could be fixed later (in a different PR). @viktormalik WDYT?

Also sorry for not reviewing this sooner.

viktormalik:
Thanks for the thorough testing @PLukas2018! I agree with everything, we should fix the `diffkemp-out.yaml` generation and can leave the rest for a follow-up work (preferably open issues once this is merged to track).

davidkre525:
Thanks for the notes. But I dont fully understand, what exactly is the bug . When I tried to reproduce it and compared results from master and from PR on the symbol list above and:

- The function-symbols results were similar with some differences in differing functions as you mentioed

- There were 6 results for global variable-symbols (the ones you mentioned above, but two of them were missing for some reason). And there was one differing function for `__cpu_online_mask` and two for `apic`.

How do I know, which functions are missing? And I am not sure, why are my results different from yours. 8.0/8.1 means
linux-4.18.0-80 and linux-4.18.0-147, right?

PLukas2018:
> How do I know, which functions are missing? And I am not sure, why are my results different from yours. 8.0/8.1 means
linux-4.18.0-80 and linux-4.18.0-147, right?

@davidkre525 Yes, `linux-4.18.0-80` and `linux-4.18.0-147`. IIRC the differing functions were correctly saved to directory (the `....diff` files) but the `diffkemp-out.yaml` was missing multiple functions. It is possible that I did some mistake when testing. I will try to retest in afternoon and if I get to the same problem, I will send some logs/files.

Edit: Thinking about it, I may have edit the output because I wanted to have some example how the new output looks so I can update the bot, so maybe there were originally all functions, I will need to recheck it.

davidkre525:
Missing function definitions in `diffkemp-out.yaml` should be fixed now. The change should also fix  #406. The stats now seem to be correct as well:

```
Statistics - master
-----------
Total symbols: 106
Equal:         66 (62%)
Not equal:     12 (11%)
(empty diff):  0 (0%)
Unknown:       28 (26%)
Errors:        0 (0%)

Elapsed time:            113.68 s
Functions compared:      1416
Lines compared:          12527
Instructions compared:   38484
1:1 equal instructions:  37944 (99%)

Total differences:       35
In functions:            25 (71%)
In types:                2 (6%)
In macros:               8 (23%)
In inline assembly code: 0 (0%)
Empty diffs:             1 (3%)
```
```
Statistics - PR
-----------
Total symbols: 196
Equal:         160 (82%)
Not equal:     16 (8%)
(empty diff):  1 (1%)
Unknown:       20 (10%)
Errors:        0 (0%)

Elapsed time:            140.73 s
Functions compared:      1768
Lines compared:          15871
Instructions compared:   49408
1:1 equal instructions:  48846 (99%)

Total differences:       37
In functions:            27 (73%)
In types:                2 (5%)
In macros:               8 (22%)
In inline assembly code: 0 (0%)
Empty diffs:             2 (5%)
```
The missing definitions are what you meant, or did you find some missing functions in the `results` as well?
Also I noticed that in [add_funcs_for_globvar](https://github.com/davidkre525/diffkemp/blob/245ab69796ccc99e8550c233331668b0ef863381/diffkemp/building/build_utils.py#L57) parameter `name` was redundant and [generate_from_function_list](https://github.com/davidkre525/diffkemp/blob/245ab69796ccc99e8550c233331668b0ef863381/diffkemp/building/build_utils.py#L23) omitted to print one newline character for each compared global variable, so I changed also these two things.

diffkemp/building/build_kernel.py
```diff
@@ -97,7 +98,8 @@ def generate_from_sysctl_list(snapshot, sysctl_list):
             # Proc handler function for sysctl
             _add_proc_handler(snapshot, sysctl, proc_fun)
             # Adds functions which use the sysctl data variable
-            _add_funcs_for_data(snapshot, sysctl, sysctl_mod, proc_fun)
```
PLukas2018, line 100: I am not 100% sure but I think the sysctl param is important -- without it we would lose connection to which sysctl param the functions belong to. IIRC sysctl != data (data = global variable set by the sysctl != sysctl). Feel free to correct me.

diffkemp/compare.py
```diff
@@ -33,17 +33,18 @@ def compare(args):
...
     def __init__(self, group, group_name, enable_module_cache, config,
-                 output_dir):
+                 output_dir, group_kind, result_graph):
         self.group = group
         self.group_name = group_name
         self.group_dir = self._get_group_dir(output_dir)
         self.group_printed = False
         self.modules_to_cache = \
             self._get_modules_to_cache_if_enabled(enable_module_cache, config)
-        self.result_graph = None
```
PLukas2018, line 43: I think we should not do this as IIRC it could pollute results.
IIRC the result_graph is used for caching results, when comparing global variables a 'slicer' is used which I think could cause that the same function is equal w.r.t. to certain global var but non-equal w.r.t. different one (I am not 100% certain about this, so take me with a grain of salt). By reusing the `result_graph` I think it could pollute the result cache, and it could happen that we lose some results.

Feel free to correct me if you think it is not that case.

diffkemp/building/build_utils.py
```diff
@@ -41,9 +42,45 @@ def generate_from_function_list(snapshot, fun_list):
                 snapshot.add_fun(symbol, llvm_mod)
                 print(os.path.relpath(llvm_mod.llvm,
                                       snapshot.source_tree.source_dir))
+            elif llvm_mod.has_global(symbol):
+                print()
```
PLukas2018, line 46: Good catch!

PLukas2018 (commented):
@davidkre525 Hi!

> But I dont fully understand, what exactly is the bug . When I tried to reproduce it and compared results from master and from PR on the symbol list above and:
> The missing definitions are what you meant, or did you find some missing functions in the results as well?

FYI I meant that in the `diffkemp-out.yaml` were missing some compared functions in the `results`. As I mentioned, I probably made a mistake when reviewing the results  (I think I updated the `diffkemp-out.yaml` because I wanted to extract a shorter part of the output to update the bot, and then when reviewing the results, I forgot about that 😅 ). When reviewing again, I do not see the bug. Sorry for the false alarm!

> Missing function definitions in diffkemp-out.yaml should be fixed now. The change should also fix https://github.com/diffkemp/diffkemp/issues/406. The stats now seem to be correct as well:

Yeah, I personally overlooked that the stats showed a lower number of compared functions and missing functions in definitions, so good catch, but I think the latest changes are problematic, and it should be handled differently (but that would be better left for different PR). I think the sharing of the `result_graph` between global variables could pollute the results.

> Also I noticed that in [add_funcs_for_globvar](https://github.com/davidkre525/diffkemp/blob/245ab69796ccc99e8550c233331668b0ef863381/diffkemp/building/build_utils.py#L57) parameter name was redundant and [generate_from_function_list](https://github.com/davidkre525/diffkemp/blob/245ab69796ccc99e8550c233331668b0ef863381/diffkemp/building/build_utils.py#L23) omitted to print one newline character for each compared global variable, so I changed also these two things.

I am not sure if it is redundant. In case of global variables, yes, but in case of sysctl parameters, IIRC the args represent different things -- name of the sysctl parameter vs global variable which the sysctl parameter sets.

---

Overall, I do not like [the last change](https://github.com/diffkemp/diffkemp/compare/3bec8a22316c4fedb540a4924faa2983a4c48e8a..245ab69796ccc99e8550c233331668b0ef863381) much (the added print is ok), I think it could cause problems (pollution of results, missing info about sysctl name), but feel free to disprove me.

Otherwise, if the last change were reverted, there are the following issues which I think would be ok to leave for a different PR, and this could probably be merged (?):
- https://github.com/diffkemp/diffkemp/issues/436
-  https://github.com/diffkemp/diffkemp/issues/437
- Missing definition (https://github.com/diffkemp/diffkemp/issues/406) and stats not reflecting reality

---
Personally, I have still compared only subset of kabi and think it would be good to try compare the whole list (only functions) to check that it does not introduce other issues -- I have tried to update the bot to compare only functions (it previously time-outed on the global vars building)  so hopefully it will bring some results -- https://github.com/PLukas2018/diffkemp/pull/17#issuecomment-4827442405.

davidkre525:
You are right, I reverted the changes (besides the newline).
One more thing we discussed was addition of a cli option `symbol-kind` (or something like that), which would limit the comparison to only functions/global variables from the symbol list. But I am not sure, if we decided to add it or not in the end.

PLukas2018:
> One more thing we discussed was addition of a cli option symbol-kind (or something like that), which would limit the comparison to only functions/global variables from the symbol list. But I am not sure, if we decided to add it or not in the end.

@davidkre525 I think that we decided to leave it for different PR, I have create #437 so we do not forget on it.

PLukas2018:
\evaluate --rebuild

PLukas2018:
\evaluate --rebuild

PLukas2018:
So building and comparison of only functions seems ok. I am wondering about one more thing -- if combination of global variables and seperate functions will not cause that the result_graph will be cleaned even between individual top level functions which would cause that we will not use cached results in cases when it would be useful and it would prolong the comparison. I have not yet checked code with regards to this, but it is something which occured me and I am wondering how it behaves with respect to this.

davidkre525:
I think this is not the case, because the `result_graph` is set to `None` only when new instance of [GroupInfo](https://github.com/davidkre525/diffkemp/blob/a5aa1518a9d6f2d35ebfba1614500015af4c3334/diffkemp/compare.py#L35) is created in [_compare_snapshots](https://github.com/davidkre525/diffkemp/blob/a5aa1518a9d6f2d35ebfba1614500015af4c3334/diffkemp/compare.py#L134) . And since all top level functions are in one group, they should share the graph. ([here](https://github.com/davidkre525/diffkemp/blob/a5aa1518a9d6f2d35ebfba1614500015af4c3334/diffkemp/compare.py#L193) the graph from `group_info` is parsed to `_compare_function`)

PLukas2018:
> I think this is not the case, because the result_graph is set to None only when new instance of [GroupInfo](https://github.com/davidkre525/diffkemp/blob/a5aa1518a9d6f2d35ebfba1614500015af4c3334/diffkemp/compare.py#L35) is created in [_compare_snapshots](https://github.com/davidkre525/diffkemp/blob/a5aa1518a9d6f2d35ebfba1614500015af4c3334/diffkemp/compare.py#L134) . And since all top level functions are in one group, they should share the graph. ([here](https://github.com/davidkre525/diffkemp/blob/a5aa1518a9d6f2d35ebfba1614500015af4c3334/diffkemp/compare.py#L193) the graph from group_info is parsed to _compare_function)

I think you are right, IIUC/IIRC the functions are probably in their own 'subgroup' (with `group_name` being `None`) when being compared.

Looking at the code, I just found this: 
https://github.com/diffkemp/diffkemp/blob/master/diffkemp/compare.py#L147-L154
But IIRC this should not have impact on the comparison graph, it just should mean that the functions results are being added directly to overall result instead of subgroup result, which is probably fine :thinking: but I am not certain.
I have created https://github.com/diffkemp/diffkemp/issues/441 to track this, but I think it is ok to be left as is.

PLukas2018 (approved):
@viktormalik I think this could be merged if you are ok with some issues not resolved but being tracked.

Summary:
- FYI the bug (regarding missing results in `diffkemp-out.yaml`) that I mentioned was a mistake on my side when doing the review, so there seem to be all results.
- I updated the bot to compare just functions, and the results are the same between master and the pr ^^.
- I compared the kabi sublist again and the results seem ok-ish.
- I have created issues for things which I discovered:
  - https://github.com/diffkemp/diffkemp/issues/437 As discussed, I think it would be useful to add option to `build-kernel` to build snapshot only for functions/global var/both. This would be useful for the bot (as currently there is workaround/patch to do this) and I think it would be useful also for other use cases.
  - https://github.com/diffkemp/diffkemp/issues/436 I was unable to compare the whole kabi list (including all global var) as the building time is too slow, ...
  - https://github.com/diffkemp/diffkemp/issues/406 There is a problem with the extraction of definitions for functions and global var combination, but also for sysctl parameters.
  - https://github.com/diffkemp/diffkemp/issues/438  There is also a problem with reporting of some stats for a similar reason.
  - https://github.com/diffkemp/diffkemp/issues/441 We are putting results of directly compared functions to the overall result without sub-group -- this is not probably a problem, it is more for tracking it if it appeared as problem in the future.

PR #434, opened 2026-04-30
Bitcode continuance

Bal0sk:
This PR is a continuation of #382

While finishing the PR, I ran into an issue with the `kernel_module_test` failing when using clang-19. It turns out newer versions of Clang reject certain inline assembly code even when we're just trying to generate LLVM IR. I fixed this in `llvm_ir/kernel_source_builder.py` by adding a fallback `check_call`. If the initial build fails, it retries and strips out the assembly statements using `-D__asm__(...)=` and `-Dasm(...)=`.

This inline assembly is the exact same thing causing the segmentation fault when comparing musl libraries, specifically `dl_start.c`. The inline assembly results in a null pointer dereference. I tried to find a workaround during the build phase, but there doesn't seem to be a complete solution without actually modifying the SimpLL library itself.

As @PLukas2018 pointed out in #382, the best fix is probably to add a check in the SimpLL library to make sure the function isn't nullptr after we get it by name.

Let me know what you think about this approach, or if you have any other feedback.

### DIFF
--- a/diffkemp/llvm_ir/kernel_llvm_source_builder.py
+++ b/diffkemp/llvm_ir/kernel_llvm_source_builder.py
@@ -70,7 +70,7 @@ class KernelLlvmSourceBuilder(LlvmSourceFinder):
                 llvm_filename = self._build_source_to_llvm(source_path)
                 if os.path.isfile(llvm_filename):
                     mod = LlvmModule(llvm_filename)
-                    if mod.has_function(symbol) or mod.has_global(symbol):
+                    if mod.has_definition(symbol):
                         break
             except BuildException:
                 pass
@@ -395,9 +395,9 @@ class KernelLlvmSourceBuilder(LlvmSourceFinder):
             if param.startswith('-D"DEBUG_HASH2='):
                 param = '-D"DEBUG_HASH2=1"'
 
-            # Output name is given by replacing .c by .ll in source name
+            # Output name is given by replacing .c by .bc in source name
             if param.endswith(".c"):
-                output_file = "{}.ll".format(param[:-2])
+                output_file = "{}.bc".format(param[:-2])
 
             command.append(KernelLlvmSourceBuilder._strip_bash_quotes(param))
         if output_file is None:
@@ -414,10 +414,10 @@ class KernelLlvmSourceBuilder(LlvmSourceFinder):
         :param ld_command: Command to convert
         :return Corresponding llvm-link command.
         """
-        command = ["llvm-link", "-S"]
+        command = ["llvm-link"]
         for param in ld_command.split():
             if param.endswith(".o"):
-                command.append("{}.ll".format(param[:-2]))
+                command.append("{}.bc".format(param[:-2]))
             elif param == "-o":
                 command.append(param)
         return command
@@ -486,8 +486,12 @@ class KernelLlvmSourceBuilder(LlvmSourceFinder):
 
     @staticmethod
     def _get_build_source(command):
-        """Get name of the object file built by the command."""
-        return command[command.index("-c") + 1]
+        """
+        Get name of the source file built by the command.
+        The commands start with 'clang -c -emit-llvm' to avoid linking
+        and the source file follows '-c' later in the command.
+        """
+        return command[command.index("-c", 3) + 1]
 
     def _kbuild_object_command(self, object_file):
         """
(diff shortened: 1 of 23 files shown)


Bal0sk:
Based on the discussion during the meeting I:
- added a `nullptr` check to the `SimpLL` library to avoid segmentation fault.
- cleaned up the commit history.

**Notes:**
After the addition of the `nullptr` check to the `SimpLL` library, another issue surfaced. When building the `Comparison Graph`, a function was in the function list but not in the vertices of the graph. These two issues are linked to the same problem.

Why did compare `seg fault` on `musl-1.2.3` library:
The compare pipeline assumes that the first (Main) function of a module, is a regular function. This is not the case for `_dlstart` and `_start` in the `musl` library.

My assumption is that when building the snapshot with implicit `function list`, the `llvm-nm` recognizes them as symbols and adds them to the function list. However, during comparison, the functions are derived from the `.yaml` file, where in this case, `_start` is associated with the file `rcrt1.c` where it is only defined in a macro: `#define START "_start"` and therefore, is not a "regular" function. Later, when the program tries to find the function definition, it doesn't find any and passes a `nullptr` to the `CalledFunctionsAnalysis` and based on the naive assumption mentioned above, it causes a segmentation fault. The same problem applies to the `Comparison Graph`.

My solution currently adds 2 safe guards for these issues, one in `run` function of the `CalledFunctionAnalysis` and an additional one for the `Graph` in `reachable_from` in `caching.py`.

I would like your thoughts on this matter as well as any additional notes.

PLukas2018:
\evaluate --rebuild

PLukas2018:
> My assumption is that when building the snapshot with implicit function list, the llvm-nm recognizes them as symbols and adds them to the function list. However, during comparison, the functions are derived from the .yaml file, where in this case, _start is associated with the file rcrt1.c where it is only defined in a macro: #define START "_start" and therefore, is not a "regular" function. Later, when the program tries to find the function definition, it doesn't find any and passes a nullptr to the CalledFunctionsAnalysis and based on the naive assumption mentioned above, it causes a segmentation fault. The same problem applies to the Comparison Graph.

I think you are right, the `llvm-nm` is not very accurate for our use-case -- based on Gemini:

> The core issue is that llvm-nm treats any text symbol as T or t whether it has a standard high-level LLVM IR function body or if it's just an embedded piece of raw assembly/linker symbol. To llvm-nm, they look completely identical.

IIRC/IIUC, the main problem with respect to musl is that when the function list is not specified, we try to explore the files and extract all functions which are defined in the C projects and compare all of them (I think this is not probably very realistic use case of DiffKemp, but I do not know). We could probably use  `llvm-extract --func` to check that the LLVM byte-code contains the functions which we extracted using `llvm-nm`, but I am not sure how much it would slow down the snapshot creation process and if it is worth it... Example:

```sh
# Extraction of functions
$ llvm-nm musl-1.2.3/obj/crt/rcrt1.bc
                 w _DYNAMIC
---------------- T __dls2
                 U __libc_start_main
                 w _fini
                 w _init
---------------- T _start               # We would extract _start
---------------- T _start_c
                 U _start_c
                 U main

# Double checking that the functions are really defined in the files
$ llvm-extract --func _start musl-1.2.3/obj/crt/rcrt1.bc -o /dev/null 
llvm-extract: program doesn't contain function named '_start'!
# we would remove _start from the function list

$  llvm-extract --func _start_c musl-1.2.3/obj/crt/rcrt1.bc -o /dev/null
# Success, we would not remove it, but it is probably not ideal...
```

diffkemp/llvm_ir/kernel_llvm_source_builder.py
```diff
@@ -585,9 +589,20 @@ def _build_source_to_llvm(self, source_file):
...
-                        raise BuildException(
-                            "Could not build {}".format(llvm_file))
+                        # Some sources contain architecture-specific inline
+                        # assembly that newer clang versions reject even when
+                        # only LLVM IR is needed (e.g. for include metadata).
+                        # Retry once with asm statements stripped.
+                        fallback_command = command + [
+                            "-D__asm__(...)=",
+                            "-Dasm(...)=",
+                        ]
```
PLukas2018, line 599: I am not sure if this is a good solution; it could probably hide/remove part, causing semantic differences, and we could report some functions as equal even if there would be semantic differences (increase IIRC *false negatives*).

I am not sure what would be a good way to solve this (keep snapshot building failure?)...

PR #435, opened 2026-05-19
Module parameters comparison

davidkre525:
Adds the CLI option --module-params. When used, the symbol list is
interpreted as a list of pairs in the form
`path_to_module:parameter`. For each pair, all functions from the
module using the global variable corresponding to the module
parameter are compared.

This PR is built on top of the commits from the
[Add comparison of global variables](https://github.com/diffkemp/diffkemp/pull/432)
PR, because both modify the same functions in
[diffkemp/building/build_utils.py](https://github.com/davidkre525/diffkemp/blob/module-parameters-comparison/diffkemp/building/build_utils.py).

Since the mentioned PR has not been merged yet, its commits are also
visible here.

### DIFF
--- a/diffkemp/building/build_kernel.py
+++ b/diffkemp/building/build_kernel.py
@@ -5,8 +5,10 @@ from diffkemp.snapshot import Snapshot
 from diffkemp.building.build_utils import (
     generate_from_function_list,
     add_funcs_for_globvar,
+    add_funcs_for_globvar_from_module,
     read_symbol_list,
     EMSG_EMPTY_SYMBOL_LIST)
+from pathlib import Path
 import errno
 import os
 import sys
@@ -32,6 +34,8 @@ def build_kernel(args):
     # Generate snapshot contents
     if args.sysctl:
         generate_from_sysctl_list(snapshot, symbol_list)
+    elif args.module_params:
+        generate_from_module_param_list(snapshot, symbol_list)
     else:
         generate_from_function_list(snapshot, symbol_list)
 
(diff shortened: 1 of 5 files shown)


diffkemp/building/build_kernel.py
```diff
@@ -102,6 +110,26 @@ def generate_from_sysctl_list(snapshot, sysctl_list):
...
+def generate_from_module_param_list(snapshot, module_param_list):
+    """
+    Generate a snapshot from a list of pairs path:param.
+    :param snapshot: Existing Snapshot object to fill
+    :param module_param_list: List of pairs path_to_module:parameter_name.
+    """
+    for pair in module_param_list:
+        try:
+            path, param = pair.split(":")
+            dir = "/".join(path.split("/")[:-1])
```
viktormalik: Can we use `Path` from `pathlib` to represent the module path? It should make working with the path (such as getting the module directory and name) easier and more understandable.

diffkemp/snapshot.py
```diff
@@ -226,15 +226,12 @@ def _from_yaml(self, yaml_file):
...
-                functions = g["functions"]
-            else:
-                group = None
-                functions = g
+            group = None
+            functions = g
+            for list_kind in ["sysctl", "glob_var",  "module:param"]:
+                if list_kind in g:
+                    group = g[list_kind]
+                    functions = g["functions"]
```
viktormalik, line 234: `break` here

diffkemp/cli.py
```diff
@@ -84,6 +84,10 @@ def make_argument_parser():
         "--sysctl",
         action="store_true",
         help="interpret symbol list as a list of sysctl parameters")
+    build_kernel_ap.add_argument(
+        "--module-params",
+        action="store_true",
+        help="interpret symbol list as a list of pairs path:parameter")
```
viktormalik:
```suggestion
        help="interpret symbol list as a list of kernel module parameters in the form module_path:parameter")
```

viktormalik (commented):
It would be also nice to document this. For the very least, add the new option to `docs/usage.md`

It would be nice to also better document sysctls and module params comparison but we can do that as a follow-up to this PR.

davidkre525:
Thanks for the notes, I have added the documentation and fixed other mentioned issues.

PLukas2018:
\evaluate --rebuild

PLukas2018:
@davidkre525 Could you rebase this, please.

viktormalik (approved):

